PR #48, opened 2020-07-13
fix excluding rules

bhirsz:


robocop/config/config.py
```diff
@@ -56,7 +56,7 @@ def is_rule_enabled(self, msg):
             if msg.msg_id not in self.include and msg.name not in self.include:
                 return False
         if self.exclude:
-            if msg.msg_id in self.include or msg.name in self.include:
+            if msg.msg_id in self.exclude or msg.name in self.exclude:
```
bhirsz, line 59: This show how proper testing is important :D @mnojek

robocop/config/config.py
```diff
@@ -23,8 +23,8 @@ def __init__(self):
         self.configure = list()
         self.format = "{source}:{line}:{col} [{severity}] {msg_id} {desc}"
         self.paths = []
-        self.include_patterns = None
-        self.exclude_patterns = None
+        self.include_patterns = list()
```
mnojek, line 26: Why not to just create empty list by calling `self.include_patterns = []`? I think it's more concise and even a bit faster.

robocop/config/config.py
```diff
@@ -55,15 +55,12 @@ def is_rule_enabled(self, msg):
         if self.include and not self.include_patterns:
             if msg.msg_id not in self.include and msg.name not in self.include:
                 return False
-        if self.exclude:
-            if msg.msg_id in self.include or msg.name in self.include:
+        if msg.msg_id in self.exclude or msg.name in self.exclude:
```
mnojek, line 58: You can also type 
```python
if any(x in [msg.msg_id, msg.name] for x in self.exclude):
```
but I am not sure if it is more readable.

bhirsz: actually in this case self.exclude is set so it's faster to check if msg is in self.exclude that other way around. 

Your code would be then this:
```
any(m in self.exclude for m in (msg.msg_id, msg.name)
```
But I think just two ors is bit more readable than that

PR #49, opened 2020-07-13
detect resource and init types and parse them

### LINKED ISSUE #6: Register possible messages while registering rules in common store

bhirsz:
Closes #6

### DIFF
--- a/robocop/checkers/tags.py
+++ b/robocop/checkers/tags.py
@@ -37,5 +37,5 @@ class TagChecker(BaseChecker):
     def check_tag(self, tag, node):
         if ' ' in tag.value:
             self.report("tag-with-space", node=node, lineno=tag.lineno, col=tag.col_offset + 1)
-        if 'or' in tag.value.lower() or 'and' in tag.value.lower():
+        if 'OR' in tag.value or 'AND' in tag.value:
             self.report("tag-with-reserved", node=node, lineno=tag.lineno, col=tag.col_offset + 1)
--- a/robocop/run.py
+++ b/robocop/run.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from robocop import checkers
 from robocop.config import Config
 from robocop import reports
-from robocop.utils import DisablersFinder
+from robocop.utils import DisablersFinder, FileType, FileTypeChecker
 
 
 SUPPORTED_FORMATS = ('.robot')
@@ -13,6 +13,7 @@ SUPPORTED_FORMATS = ('.robot')
 
 class Robocop:
     def __init__(self):
+        self.files = dict()
         self.checkers = []
         self.messages = dict()
         self.reports = []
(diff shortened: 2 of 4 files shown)


robocop/run.py
```diff
@@ -24,17 +25,30 @@ def __init__(self):
...
+            file_type_checker.visit(model)
+
+    def run_checks(self):
+        for file in self.files:
+            print(f"Parsing {file}")
             self.register_disablers(file)
             if self.disabler.file_disabled:
                 continue
-            model = get_model(str(file))
+            model = self.files[file].get_parser()(str(file))
```
bhirsz, line 51: more functional programming ;)

robocop/utils/file_types.py
```diff
@@ -0,0 +1,38 @@
...
+        }[self.value]
+
+
+class FileTypeChecker(ast.NodeVisitor):
+    def __init__(self, files):
+        self.files = files
+        self.source = None
+
+    def visit_ResourceImport(self, node):
+        path_normalized = node.name.replace('${/}', os.path.sep)
```
bhirsz, line 29: we need to add more normalization methods, I will create issue for it (replace CURDIR etc)

robocop/checkers/tags.py
```diff
@@ -37,5 +37,5 @@ def check_tags(self, node):
     def check_tag(self, tag, node):
         if ' ' in tag.value:
             self.report("tag-with-space", node=node, lineno=tag.lineno, col=tag.col_offset + 1)
-        if 'or' in tag.value.lower() or 'and' in tag.value.lower():
+        if 'OR' in tag.value or 'AND' in tag.value:
```
mnojek, line 40: Same as in previous PR 😄 : 
```python
if any(x in ['OR', 'AND'] for x in tag.value):
```

robocop/run.py
```diff
@@ -5,14 +5,15 @@
...
-from robocop.utils import DisablersFinder
+from robocop.utils import DisablersFinder, FileType, FileTypeChecker
 
 
 SUPPORTED_FORMATS = ('.robot')
 
 
 class Robocop:
     def __init__(self):
+        self.files = dict()
```
mnojek, line 16: Why not `self.files = {}`?

bhirsz: You're right and I'm not consistent through the same class. I will change this to {}

robocop/checkers/tags.py
bhirsz: In this case tag.value is string (check_tag is invoked on each tag found). 
But using your approach we can do:
```
if any(reserved in tag.value for reserved in ('OR', 'AND'):
```
Should I change it?

mnojek: You're right - it's str, I didn't notice.
If we are here to play than another option is to use:
`any(map(tab.value.__contains__, ['OR', 'AND'))`
But I think that the original one is simple and more understandable.

bhirsz: It's starting to look like code golf :D

PR #53, opened 2020-07-13
configurable allowed file types

### LINKED ISSUE #11: Parse only allowed extensions (*.robot for now)

bhirsz:
Closes #11 
Also fixes two bugs:
Robocop was scanning files without extension (like .gitignore, .pabotsuitenames where filename is '.gitignore' and extension '')
Robocop was scanning only 'robot' file types (where in RF 3.2 default types are 'robot' and 'resource')

### DIFF
--- a/robocop/config/config.py
+++ b/robocop/config/config.py
@@ -15,6 +15,14 @@ class ParseCheckerConfig(argparse.Action):
         container.append(values)
 
 
+class ParseFileTypes(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        container = getattr(namespace, self.dest)
+        for filetype in values.split(','):
+            filetype = filetype if filetype.startswith('.') else '.' + filetype
+            container.add(filetype)
+
+
 class OpenOutputFile(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         file = open(values, 'w')  # TODO: raise custom exception if fail (ie dont have permissions)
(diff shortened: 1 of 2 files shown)


robocop/config/config.py
```diff
@@ -31,6 +39,7 @@ def __init__(self):
         self.paths = []
         self.include_patterns = list()
         self.exclude_patterns = list()
+        self.filetypes = {'.robot', '.resource'}
```
mnojek, line 42: .robot and .resource formats are supported by default but RF still supports .tsv and .rst or .rest if they are indicated explicitly. It would be good to support them in the future.

robocop/config/config.py
```diff
@@ -54,8 +63,10 @@ def parse_opts(self):
                             help="Configure checker with parameter value")
         parser.add_argument('-o', '--output', action=OpenOutputFile, default=self.output,
                             help='Path to output file')
+        parser.add_argument('--filetypes', action=ParseFileTypes, default=self.filetypes,
+                            help='Comma seperated list of file extensions to be scanned by Robocop')
```
mnojek, line 67: Typo: `separated`

bhirsz: Thanks!

robocop/config/config.py
bhirsz: They're supported - you just need to indicate it with:

--filetypes:robot,resource,tsv,rst,rest 

(by default only robot and resource is parsed but you can configure whatever extension you want, even txt or png ;) )

mnojek: Writing robot tests in PNG would be something unusual 😄

PR #57, opened 2020-07-14
Feature/rawfile checker

### LINKED ISSUE #54: Add raw file checker type and visitor type (default, all current rules are visitor type)

### LINKED ISSUE #56: line too long rule

bhirsz:
* Closes #54 We can now define our checker RawFileChecker or VisitorChecker. RawFileChecker have visit_line method that need to be overriden -> it's called for every line in file. If you need more customized file parsing you can also override visit_file which opens file. 

* Applied comments to my previous PRs from @mnojek . 

* Closes #56 - line too long rule. 

* Removed OpenOutputFile in favor of builtin arg parse FileType


robocop/checkers/__init__.py
```diff
@@ -61,6 +60,26 @@ def report(self, msg, *args, node=None, lineno=None, col=None):
...
+    type = 'visitor_checker'
+
+    def visit_File(self, node):
+        self.generic_visit(node)
+
+
+class RawFileChecker(BaseChecker):
+    type = 'rawfile_checker'
+
+    def visit_file(self):
```
mnojek, line 75: Inconsistent case in model naming. sometimes you use capitals, and sometimes lowercase :)

bhirsz: I'm forced to used capitals if I'm implementing checkers based on official robot parsing api. It only matches visit_ClassName methods, where ClassName is title capitalized. I could change method names in RawFileChecker so it would be not confused with VisitorChecker

PR #58, opened 2020-07-14
Add parser info and organize help messages

mnojek:
- Add file containing tool version: `robocop/version.py`
- Extract help messages to Config class attribute `HELP_MSGS`
- Add `optional` and `required` argument groups
- Update version variable in `setup.py`


robocop/config/config.py
```diff
@@ -45,22 +62,33 @@ def translate_patterns(self):
...
+                              help=self.HELP_MSGS['help_reports'])
+        optional.add_argument('-f', '--format', type=str, default=self.format, help=self.HELP_MSGS['help_format'])
+        optional.add_argument('-c', '--configure', action=ParseCheckerConfig, default=self.configure,
+                              help=self.HELP_MSGS['help_configure'])
+        optional.add_argument('-o', '--output', type=argparse.FileType('w'), default=self.output,
+                              help=self.HELP_MSGS['help_output'])
+        optional.add_argument('--filetypes', action=ParseFileTypes, default=self.filetypes,
+                              help=self.HELP_MSGS['help_filetypes'])
+        optional.add_argument('-h', '--help', action='help', help=self.HELP_MSGS['help_info'])
+        optional.add_argument('-v', '--version', action='version', version=__version__,
```
bhirsz, line 89: This whole code block is nicely reformatted and more readable, good job

robocop/config/config.py
```diff
@@ -45,22 +62,33 @@ def translate_patterns(self):
...
-        parser.add_argument('-f', '--format', type=str, help='Format of output message', default=self.format)
-        parser.add_argument('-c', '--configure', action=ParseCheckerConfig, default=self.configure,
-                            help="Configure checker with parameter value")
-        parser.add_argument('-o', '--output', type=argparse.FileType('w'), default=self.output,
-                            help='Path to output file')
-        parser.add_argument('--filetypes', action=ParseFileTypes, default=self.filetypes,
-                            help='Comma separated list of file extensions to be scanned by Robocop')
-        parser.add_argument('paths', metavar='paths', type=str, nargs='+',
-                            help='List of paths (files and directories) to be scanned by Robocop')
+        parser = argparse.ArgumentParser(prog='robocop',
```
bhirsz, line 65: This whole code block is nicely reformatted and more readable, good job

robocop/config/config.py
bhirsz, line 89: what happens if I want to type this:
```
robocop -v
```
? Most of the python tools support only asking for version without need for supplying required argument. But we would need to change behaviour a bit. For example preparse options and leave program (after printing version).

mnojek: It works out of the box in argparse by using dedicated actions like `help` and `version`. It then ignores other parameters, prints the info and exit the application.

PR #62, opened 2020-07-19
Add basic unit tests for argument validation

mnojek:


utest/test_argument_validation.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertSetEqual(args.include, set())
+        self.assertSetEqual(args.exclude, set())
+        self.assertSetEqual(args.reports, set())
+        self.assertListEqual(args.configure, [])
+        self.assertEqual(args.format, "{source}:{line}:{col} [{severity}] {msg_id} {desc}")
+        self.assertListEqual(args.paths, [''])
+        self.assertIsNone(args.output)
+
+    def test_overwrite_default_filetypes(self):
+        args = self.config.parse_opts(['--filetypes', 'txt', ''])
```
bhirsz, line 42: what happens if you provide empty string? it will just overwrite it with empty set?

utest/test_argument_validation.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertSetEqual(args.filetypes, {'.txt'})
+
+    def test_duplicate_default_filetypes(self):
```
bhirsz, line 45: we could use name of argument in the beggining so it's easy to see what feature is tested. like we have "test_include.." in this case "test_filetypes_duplicate_default", "test_filetypes_overwrite_default" and so on. what do you think?

Or maybe there is some way to tag test cases

utest/test_argument_validation.py
mnojek: will take next argument as a value for this option

utest/test_argument_validation.py
mnojek: I think it's a good idea but I will look into unittest a little bit more. 
We can also use pytest instead of unittest but need to find out if that is not an overkill for us.

PR #66, opened 2020-07-20
Add pylama configuration and refactor code

mnojek:
- Add `pylama.ini` file for local and GitHub Actions purposes
- To run pylama with configuration file execute `pylama -o pylama.ini`
- Refactor code accordingly to issues reported by pylama

### DIFF
--- a/robocop/checkers/__init__.py
+++ b/robocop/checkers/__init__.py
@@ -26,9 +26,9 @@ to refer to rule (in include/exclude statements, configurations etc).
 You can configure rule severity and optionally other parameters.
 """
 import ast
+import inspect
 from pathlib import Path
 from importlib import import_module
-import inspect
 from robocop.messages import Message
 
 
@@ -46,7 +46,8 @@ class BaseChecker:
         for key, value in msgs.items():
             msg = Message(key, value)
             if msg.name in self.messages:
-                raise ValueError("Duplicate message name in checker")  # TODO: add better handling for duplicate messages
+                raise ValueError("Duplicate message name in checker")
+                # TODO: add better handling for duplicate messages
             self.messages[msg.name] = msg
 
     def report(self, msg, *args, node=None, lineno=None, col=None):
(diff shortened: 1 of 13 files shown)


robocop/checkers/__init__.py
```diff
@@ -61,25 +62,25 @@ def configure(self, param, value):
...
+class VisitorChecker(BaseChecker, ast.NodeVisitor):  # noqa
     type = 'visitor_checker'
 
-    def visit_File(self, node):
+    def visit_File(self, node):  # noqa
         self.generic_visit(node)
 
 
-class RawFileChecker(BaseChecker):
+class RawFileChecker(BaseChecker):  # noqa
```
bhirsz, line 72: Why here noqa?

robocop/run.py
```diff
@@ -70,8 +70,9 @@ def report(self, msg):
             return
         for report in self.reports:
             report.add_message(msg)
-        self.log_message(source=msg.source, line=msg.line, col=msg.col, severity=msg.severity.value,
-                         msg_id=msg.msg_id, desc=msg.desc)
+        self.log_message(source=msg.source, line=msg.line, col=msg.col,
```
bhirsz, line 73: Maybe in such cases lets go with one line one kwarg? I wonder what looks better

robocop/checkers/__init__.py
mnojek: I don't remember, something with the class name. Comment out and check it or I can do it in the evening.

robocop/run.py
mnojek: We can try but then we will have code stretched vertically so that less code would fit in one screen height.

robocop/checkers/__init__.py
mnojek: Ok, I checked and it was my mistake. The error was about class docstring. Removing the comment now...

robocop/run.py
mnojek: Ok, I split arguments to separate lines.

PR #71, opened 2020-07-21
Handle invalid cli values

mnojek:
- Handle invalid values for `--configure`
- Add `__str__` method to `Message` class
- Handle open output file
- Small refactor
- Fix typos

### DIFF
--- a/robocop/run.py
+++ b/robocop/run.py
@@ -38,6 +38,8 @@ class Robocop:
         self.recognize_file_types()
         self.run_checks()
         self.make_reports()
+        if not self.out.closed:
+            self.out.close()
 
     def recognize_file_types(self):
         """
--- a/robocop/run.py
+++ b/robocop/run.py
@@ -138,9 +138,12 @@ class Robocop:
 
     def configure_checkers(self):
         for config in self.config.configure:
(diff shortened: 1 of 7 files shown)


robocop/run.py
```diff
@@ -130,16 +132,20 @@ def any_rule_enabled(self, checker):
 
     def configure_checkers(self):
         for config in self.config.configure:
-            # TODO: handle wrong format, not existing checker
+            if config.count(':') != 2:
+                print(f'Provided invalid config: \'{config}\' (pattern: <rule>:<param>:<value>)')
+                continue
             rule, param, value = config.split(':')
             if rule not in self.messages:
+                print(f'Provided rule \'{rule}\' does not exists')
```
bhirsz, line 140: Or is disabled? Im not sure if disabled rules are still kept in messages or not

mnojek: `self.messages` contains all messages. You need to call `self.config.is_rule_enabled(msg)` to see which is enabled. I can extend the functionality or add new cli option to list all enabled/disabled messages.

bhirsz: It would be handy for verbose output (printing rules with disabled/enabled flag before scan).

PR #73, opened 2020-07-21
Add listing all rules

mnojek:
Implements #19 
- Added CLI option to list all available rules
- Changed `nargs` for `paths` from `+` to `*` which means that it is no longer a required parameter and need to be checked (possible missing check in some places, to be tested)
- A flag for listing all enabled/disabled rules still to be added


robocop/run.py
```diff
@@ -83,6 +86,11 @@ def log_message(self, **kwargs):
 
     def load_checkers(self):
         checkers.init(self)
+        if self.config.list:
+            for checker in self.checkers:
+                for msg in checker.messages:
+                    print(checker.messages[msg])
```
bhirsz, line 92: How its printed, it not require some basic sorting? Or it doesnt matter

mnojek: It's printed by using `__str__` magic method which is a part of PR #71 : 
``` python
def __str__(self):
        return f'Message - {self.msg_id} [{self.severity.value}]: {self.name}: {self.desc}'
```
So it needs to be merged before this PR 😃

PR #74, opened 2020-07-24
Feature/pylama warns

bhirsz:
Fix pylama reported issues (and ignore not important ones)
Refactor importing modules to use one common method instead

I had to ignore whole docs in pylama since some files need to use specific format

### DIFF
--- a/robocop/run.py
+++ b/robocop/run.py
@@ -41,9 +41,9 @@ class Robocop:
 
     def recognize_file_types(self):
         """
-        Preparse files to recognize types. If the filename is "__init__.robot" file type is INIT.
-        If the file is imported somewhere then file type is RESOURCE. Otherwise file type is GENERAL.
-        Those types are important since are used to define parsing class for robot API.
+        Pre-parse files to recognize their types. If the filename is `__init__.robot`, the type is `INIT`.
+        If the file is imported somewhere then file type is `RESOURCE`. Otherwise file type is `GENERAL`.
+        These types are important since they are used to define parsing class for robot API.
         """
         files = self.config.paths
         for file in self.get_files(files, True):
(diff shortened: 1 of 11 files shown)


robocop/utils/file_types.py
```diff
@@ -23,6 +26,10 @@ def get_parser(self):
 
 
 class FileTypeChecker(ast.NodeVisitor):
+    """
+    Check if file contain import statements. If the import is in list of files to be scanned, update its type
```
mnojek, line 30: Typo: contains

robocop/run.py
```diff
@@ -23,17 +26,25 @@ def __init__(self):
...
 
     def run(self):
+        """ Entry point for running scans """
         self.recognize_file_types()
         self.run_checks()
         self.make_reports()
 
     def recognize_file_types(self):
+        """
+        Preparse files to recognize types. If the filename is "__init__.robot" file type is INIT.
```
mnojek, line 44: Typo: Prepare

robocop/run.py
```diff
@@ -23,17 +26,25 @@ def __init__(self):
...
+        If the file is imported somewhere then file type is RESOURCE. Otherwise file type is GENERAL.
+        Those types are important since are used to define parsing class for robot API.
```
mnojek, line 46: I would rephrase it to:
```suggestion
        Prepare files to recognize their types. If the filename is `__init__.robot`, the type is `INIT`.
        If the file is imported somewhere then file type is `RESOURCE`. Otherwise file type is `GENERAL`.
        These types are important since they are used to define parsing class for robot API.
```

bhirsz: it's actually "preparse" as we're parsing files (in limited way) to detect file type. So what could be best name?

mnojek: Oh, ok 😄 I updated the suggestion.

PR #79, opened 2020-07-27
Handle invalid checker/message & add unit test

### LINKED ISSUE #7: Handle invalid messages (duplicates, not existing severity)
Depends on #6

bhirsz:
Unit test are created using pytest - we need to add it as dependency to our CI (and also we could create separate profile extras_requires in setup.py)

Closes #7

### DIFF
--- a/robocop/checkers/__init__.py
+++ b/robocop/checkers/__init__.py
@@ -3,8 +3,6 @@ Robocop lint rules are interally grouped into similar groups called checkers.
 Each checker can scan for multiple related issues (like LengthChecker checks both for min and max length of keyword).
 You can refer to specific messages reported by checkers by its name or id (for example `0501` or `too-long-keyword`).
 
-Each message have configurable severity and optionally other parameters.
-
 Checkers are categorized into following groups:
  * 01: base
  * 02: documentation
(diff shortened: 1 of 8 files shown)


robocop/checkers/__init__.py
```diff
@@ -3,8 +3,6 @@
 Each checker can scan for multiple related issues (like LengthChecker checks both for min and max length of keyword).
 You can refer to specific messages reported by checkers by its name or id (for example `0501` or `too-long-keyword`).
 
-Each message have configurable severity and optionally other parameters.
```
bhirsz, line 6: It is duplicated in one of the child docs

robocop/run.py
```diff
@@ -8,18 +8,20 @@
 from robocop.config import Config
 from robocop import reports
 from robocop.utils import DisablersFinder, FileType, FileTypeChecker
+from robocop.exceptions import DuplicatedMessageError
 
 
 class Robocop:
-    def __init__(self):
+    def __init__(self, from_cli=False):
```
bhirsz, line 15: thanks for this argument we can create Robocop instance inside unit test without cli (if we don't need it)

PR #81, opened 2020-07-28
Add script for parsing pylama, update workflow

### LINKED ISSUE #78: Parse pylama's output and fail step in Unit tests only on errors
Now pylama returns status other than 0 if any issue is found. Warnings and refactors should not fail unit tests but because pylama does not have an option to control return code, we need to send it to null and then parse the output independently.

mnojek:
Adds script that runs pylama and parses its output by looking for errors. If no errors were found, it returns 0, otherwise it returns the number of reported errors. This will prevent `Unit tests` workflow to fail on other issues reported by pylama like warnings, refactors etc. 

Closes #78

### DIFF
--- a/robocop/utils/pylama_parse.py
+++ b/robocop/utils/pylama_parse.py
@@ -1,0 +1,13 @@
+"""Run Pylama and parse output to find errors and exit process with proper return code"""
+import sys
+import subprocess
+
+output = subprocess.run('python -m pylama -o pylama.ini', stdout=subprocess.PIPE, stderr=subprocess.STDOUT, cwd='.',
+                        shell=True)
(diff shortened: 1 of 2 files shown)


robocop/utils/pylama_parse.py
```diff
@@ -0,0 +1,8 @@
+"""Run Pylama and parse output to find errors and exit process with proper return code"""
+import re
+import sys
+import subprocess
+
+output = subprocess.run('python -m pylama -o pylama.ini', stdout=subprocess.PIPE)
```
bhirsz, line 6: what if errors are reported to std.error? Fatal errors like missing files, wrong config are reposted to stderr instead

mnojek: I changed the stderr stream to be directed to stdout now.

PR #103, opened 2020-08-23
add tags rules

### LINKED ISSUE #102: If all tests inside suite have the same tags, warn that Force Tags could be used instead

bhirsz:
Closes #102

### DIFF
--- a/robocop/checkers/tags.py
+++ b/robocop/checkers/tags.py
@@ -7,11 +7,12 @@ from robocop.messages import MessageSeverity
 
 
 def register(linter):
-    linter.register_checker(TagChecker(linter))
+    linter.register_checker(TagNameChecker(linter))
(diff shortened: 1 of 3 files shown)


robocop/checkers/tags.py
```diff
@@ -42,4 +48,73 @@ def check_tag(self, tag, node):
...
+        self.default_tags_node = None
+        self.test_cases_count = 0
+        super().__init__(*args)
+
+    def visit_File(self, node):  # noqa
+        self.tags = []
+        self.force_tags = []
+        self.test_cases_count = 0
+        self.force_tags_node = None
+        super().visit_File(node)
```
bhirsz, line 90: thanks to overriding visit_File in this way, we make sure whole tests is pared and all nodes are hit before we start our analyze. This approach can be reused in different checkers if we need to compare values (like check for duplicates)

PR #136, opened 2020-08-31
Fix default tags not cleared between files

bhirsz:


robocop/checkers/tags.py
```diff
@@ -95,6 +96,8 @@ def visit_File(self, node):  # noqa
         if self.default_tags:
             self.report("unnecessary-default-tags",
                         node=node if self.default_tags_node is None else self.default_tags_node)
+        if self.test_cases_count < 2:
```
bhirsz, line 99: do not report "could-be-forced-tags" if there is only one test in suite

mnojek (approved):

PR #137, opened 2020-09-01
Add rule for checking redundant equal sign, add test data

### LINKED ISSUE #115: Rule: Omit redundant equal sign when assigning returned values
Using equal sign is not required in RF when assigning returned values to variables. 
It is also a potential place where bugs can appear when two spaces are made instead of one between variable and equal sign. 
Since robocop is about writing a beautiful code, this should be forced.

mnojek:
Closes #115

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -125,7 +125,7 @@ class InevenIndentChecker(VisitorChecker):
 class EqualSignChecker(VisitorChecker):
     """ Checker for redundant equal signs when assigning return values. """
     rules = {
-        "0902": (
+        "0904": (
             "redundant-equal-sign",
             "Redundant equal sign in return value assignment",
             RuleSeverity.WARNING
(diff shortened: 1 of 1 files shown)


robocop/checkers/misc.py
```diff
@@ -32,3 +33,20 @@ def visit_Keyword(self, node):  # noqa
...
+            "redundant-equal-sign",
+            "Redundant equal sign in return value assignment",
+            RuleSeverity.WARNING
+        )
+    }
+
+    def visit_KeywordCall(self, node):  # noqa
+        if node.assign:  # if keyword returns any value
+            if node.assign[-1][-1] == '=':
+                print(node.data_tokens[0])
```
bhirsz, line 50: is this print supposed to be there?

robocop/checkers/misc.py
```diff
@@ -32,3 +33,20 @@ def visit_Keyword(self, node):  # noqa
...
+                equal_position = [x for x in node.data_tokens if x.type == 'ASSIGN'][-1].end_col_offset
```
bhirsz, line 51: isn't this enough:
```
equal_position = node.assign[-1][-1].end_col_offset
```
? Same as you did in line 49

bhirsz (approved):

PR #141, opened 2020-09-02
Fix indent and capitalized keyword checker to work with old for loop

bhirsz:


robocop/checkers/naming.py
```diff
@@ -87,7 +87,6 @@ def visit_Teardown(self, node):
         self.check_if_keyword_is_capitalized(node.name, node)
 
     def visit_TestCase(self, node):  # noqa
-        self.check_if_keyword_is_capitalized(node.name, node)
```
bhirsz, line 90: removed this line as it was throwing keyword not capitalized on test case name

mnojek (approved):

PR #142, opened 2020-09-02
Spacing rules

### LINKED ISSUE #91: Rule: no of lines between sections, test cases, keyword definitions

mnojek:
Closes #91

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -32,28 +32,23 @@ class InvalidSpacingChecker(RawFileChecker):
 class MissingTrailingBlankLineChecker(VisitorChecker):
     """ Checker for invalid spacing. """
     rules = {
-        "1002": (
-            "missing-trailing-blank-line",
-            "Missing trailing blank line at the end of file",
-            RuleSeverity.WARNING
-        ),
         "1003": (
             "empty-lines-between-sections",
             "Invalid number of empty lines between sections (%d/%d)",
             RuleSeverity.WARNING,
-            ("empty_lines_between_sections", "empty_lines_between_sections", int)
+            ("empty_lines", "empty_lines_between_sections", int)
         ),
         "1004": (
             "empty-lines-between-test-cases",
             "Invalid number of empty lines between test cases (%d/%d)",
             RuleSeverity.WARNING,
-            ("empty_lines_between_test_cases", "empty_lines_between_test_cases", int)
+            ("empty_lines", "empty_lines_between_test_cases", int)
         ),
         "1005": (
             "empty-lines-between-keywords",
             "Invalid number of empty lines between keywords (%d/%d)",
             RuleSeverity.WARNING,
-            ("empty_lines_between_keywords", "empty_lines_between_keywords", int)
+            ("empty_lines", "empty_lines_between_keywords", int)
         )
     }
 
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -39,19 +39,19 @@ class MissingTrailingBlankLineChecker(VisitorChecker):
         ),
         "1003": (
             "empty-lines-between-sections",
-            "Invalid number of spaces between sections (%d/%d)",
+            "Invalid number of empty lines between sections (%d/%d)",
             RuleSeverity.WARNING,
             ("empty_lines_between_sections", "empty_lines_between_sections", int)
         ),
         "1004": (
             "empty-lines-between-test-cases",
-            "Invalid number of spaces between test cases (%d/%d)",
+            "Invalid number of empty lines between test cases (%d/%d)",
             RuleSeverity.WARNING,
             ("empty_lines_between_test_cases", "empty_lines_between_test_cases", int)
         ),
         "1005": (
             "empty-lines-between-keywords",
-            "Invalid number of spaces between keywords (%d/%d)",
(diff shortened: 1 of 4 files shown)


mnojek:
Closes #87

robocop/checkers/spacing.py
```diff
@@ -0,0 +1,119 @@
...
+        if stripped_line:
+            if stripped_line[-1] == ' ':
+                self.report("trailing-whitespace", lineno=lineno, col=len(stripped_line))
+
+
+class MissingTrailingBlankLineChecker(VisitorChecker):
+    """ Checker for invalid spacing. """
+    rules = {
+        "1002": (
+            "missing-trailing-blank-line",
```
bhirsz, line 36: is this checked somewhere?

robocop/checkers/spacing.py
```diff
@@ -0,0 +1,119 @@
...
+    """ Checker for invalid spacing. """
+    rules = {
+        "1001": (
+            "trailing-whitespace",
+            "Trailing whitespace at the end of line",
+            RuleSeverity.WARNING
+        )
+    }
+
+    def check_line(self, line, lineno):
```
bhirsz, line 25: We can reuse that checker for perfomance and to implement missing-trailing-blank-line or mixed spaces/tabs rules. 

For example given this:
```
    def __init__(self, *args):
        self.lines = []
        super().__init__(*args)

   def visit_File(self, node):
       self.lines = []
       super().visit_File(node)
       # check self.lines[-1] for missign trailing line, or too much of it

       # do stuff
       self.lines.append(line)

File: robocop/checkers/spacing.py [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+        "1002": (
+            "missing-trailing-blank-line",
+            "Missing trailing blank line at the end of file",
+            RuleSeverity.WARNING
+        ),
+        "1003": (
+            "empty-lines-between-sections",
+            "Invalid number of empty lines between sections (%d/%d)",
+            RuleSeverity.WARNING,
+            ("empty_lines_between_sections", "empty_lines_between_sections", int)
```
bhirsz, line 44: Configurables are in format of ("public name", "internal_name", type). Since if we're configuring we're already giving rule name or id we end up with this:
``
--configure empty-lines-between-sections:empty_lines_between_sections:1
```
How about making public name bit shorter? ie
```
("empty-lines", "empty_lines_between_sections", int)
```
```
--configure empty-lines-between-sections:empty-lines:2
```
(or different name). Public name needs to be only unique within rule(as we can have multiple configurables in each rule too) while private name within whole checker

File: robocop/checkers/spacing.py (hunk 1 above)
Comment (mnojek) [reply robocop/checkers/spacing.py:36]: This is nit implemented yet. I can remove it and will do it in another PR.

File: robocop/checkers/spacing.py (hunk 2 above)
Comment (mnojek) [reply robocop/checkers/spacing.py:25]: I thought about storing lines in a list but didn't have time yesterday to do it. Good idea 😃

File: robocop/checkers/spacing.py (hunk 3 above)
Comment (mnojek) [reply robocop/checkers/spacing.py:44]: Do we really need to provide `internal_name` since we have the mapping always in the messages dictionary? Maybe let's make it easier and just give `public_name` and the internal one will be resolved automatically? Or am I missing something here?

Comment (bhirsz) [reply robocop/checkers/spacing.py:44]: Internal name is later referring to attribute in checker, that's why internal_name should be unique. It could be resolved automatically only if we give some rule to it. For example <name_of_the_rule>+<public_name> = internal name. Otherwise we would need to make public name unique which would make this impossible:
```
        "0504": (
            "too-long-test-case",
            "Test case is too long (%d/%d)",
            RuleSeverity.WARNING,
            ('max_len', 'testcase_max_len', int)
```
(I can use "max_len" for all lengths rules in given checker thanks to this)

How it works in general:
Class that is handling configuration finds what rule we want to configure and then call this method to find our tuple (public, private, type) in rule:
```
configurable = msg.get_configurable(param)
```
then we're calling configure method from checker directly:
```
checker.configure(configurable[1], configurable[2](value))
```
As you can see we're using internal_name here. And configure method is updating attributes directly:
```
    def configure(self, param, value):
        self.__dict__[param] = value
```

Comment (mnojek) [reply robocop/checkers/spacing.py:44]: Ok, now I see. We can make the public_name also unique but then they would need to be longer to maintain uniqueness. I think the solution we have is better, but we need to provide the name of the rule also. It's fine.

**Comment by mnojek** [comment]
Closes #86

File: robocop/checkers/spacing.py [hunk 4]
```diff
@@ -0,0 +1,131 @@
...
+        super().parse_file()
+        if self.lines and not self.lines[-1].endswith('\n'):
+            self.report("missing-trailing-blank-line", lineno=len(self.lines), col=0)
+
+    def check_line(self, line, lineno):
+        self.lines.append(line)
+
+        stripped_line = line.rstrip('\n')
+        if stripped_line:
+            if stripped_line[-1] == ' ':
```
bhirsz, line 45:
```suggestion
        if stripped_line and stripped_line[-1] == ' ':
```

bhirsz (approved):

robocop/checkers/spacing.py
mnojek: Yeah, we talked about it and I forgot to change it. Thanks 👍

Issue #118, opened 2020-08-30
Rule: Duplicated variables


mnojek:
Verify that variables are not defined twice in variables section

PR #143, opened 2020-09-02
Add duplicates rules

### LINKED ISSUE #89: Rule: duplicated settings

### LINKED ISSUE #113: Handle warnings reported by RF in log file
Some problems with code are not reported as syntax errors but are only visible in a log file, e.g. duplicated test names or keywords.
We should check if it's possible to catch these warnings and report as rules violation.

bhirsz:
Closes #89 
Closes #113


bhirsz:
Closes #118 
Unless 118 should cover bigger scope than variables section? @mnojek

bhirsz:
I've discovered that I need to normalize keyword names too, I will update this PR

mnojek:
> Closes #118
> Unless 118 should cover bigger scope than variables section? @mnojek

I think variables section is a good start. We can always extend it.

robocop/checkers/duplications.py
```diff
@@ -0,0 +1,96 @@
...
+    def __init__(self, *args):
+        self.test_cases = defaultdict(list)
+        self.keywords = defaultdict(list)
+        self.variables = defaultdict(list)
+        self.resources = defaultdict(list)
+        self.libraries = defaultdict(list)
+        super().__init__(*args)
+
+    def visit_File(self, node):
+        self.test_cases = defaultdict(list)
```
mnojek, line 53: Do you need to assign it again? You did it in init before.

robocop/checkers/duplications.py
```diff
@@ -0,0 +1,96 @@
...
+        self.check_duplicates(self.variables, "duplicated-variable")
+        self.check_duplicates(self.resources, "duplicated-resource")
+        self.check_duplicates(self.libraries, "duplicated-library")
+
+    def check_duplicates(self, container, rule):
+        for name, nodes in container.items():
+            if len(nodes) == 1:
+                continue
+            for duplicate in nodes:
+                self.report(rule, duplicate.name, node=duplicate)
```
mnojek, line 70: Very clever

tests/duplicates.robot
```diff
@@ -0,0 +1,50 @@
+*** Settings ***
+Documentation  Suite for duplicated test cases and keywords
+Library   PythonLibrary  1
+Library   PythonLibrary  1
+Resource  resource.robot
+Library   PythonLibrary  2
+Library   PythonLibrary
+Resource  resource2.robot
+Resource  resource.robot
```
mnojek, line 9: What about Metadata?

robocop/checkers/duplications.py
bhirsz: I need to init it in init and Im resetting them again in visit_File because I want to have empty containers before every file scan. Otherwise It would compare names between files

tests/duplicates.robot
bhirsz: I was still thinking how I should handle it (for example there could be multiline metadata that is duplicate of metadata with one long line). I will update this PR with metadata checks

mnojek (approved):

Issue #150, opened 2020-09-04
Keyword name should be capitalized is too strict


bhirsz:
Keyword name should be capitalized rule is being reported also on names with embedded arguments:
```
Keyword Name ${var} And So On
This Is Embedded ${Var} Argument
```
I think it would be best to ignore embedded arguments in name (even if it's not capitalized). What do you think? @mnojek 

This rule will also report on:
```
Get Data With ID
Login To UI With User
```
What about ignoring words that are all uppercase?

I believe we also miss rule for _ in name, it could be included with Keyword name checker (together with reserved words). 
```
This_Should_Not_Happen
Because Is Looks Better
```

bhirsz:
We also need to test it with number, ie 
```
Get 3 Accounts
```
Would it pass or it will be reported since 3 is not uppercase?

bhirsz:
I've also noticed that this rule reports on parsing errors (since most syntax issues are being parsed as keywords):
```
For  ${i}  IN  ${list}
    Log  ${i}
END
```
It will report on For and END, since it's invalid loop. 
Also here
```
My Keyword
...  ${arg}
... ${not_enough_spaces_after_eol_mark}
```

Those two problems will be covered by reserved words rules.

bhirsz:
> We also need to test it with number, ie
> 
> ```
> Get 3 Accounts
> ```
> 
> Would it pass or it will be reported since 3 is not uppercase?

It will not report for numbers

PR #154, opened 2020-09-04
Extend naming checkers

### LINKED ISSUE #140: Bug: Some commented lines are parsed as test case/keyword
This
```
*** Keywords ***
Keyword
   Some Keyword
   Log  1

 ### I am comment ###
```
Parses to two keyword definitions: "Keyword" and " ### I am comment ###"

### LINKED ISSUE #145: Rule: Keyword name is reserved word
This rule could be breached while defining new keyword/using it:
```
*** Keywords ***
FOR
    Pass
```
Or as a typo when trying to use Robot Framework syntax:
```
FOr  ${i}  IN  ${elems}
    Log  ${i}
end
```
'FOr' and 'end' will be interpreted as two keywords "FOr" and "end". During test execution it will throw an error (with the exception for this bug):
https://github.com/robotframework/robotframework/issues/3648

### LINKED ISSUE #151: Support for various ways of parsing settings header
From RF user guide:

> Different sections are recognized by their header row. The recommended header format is *** Settings ***, but the header is case-insensitive, surrounding spaces are optional, and the number of asterisk characters can vary as long as there is one asterisk in the beginning. In addition to using the plural format, also singular variants like Setting and Test Case are accepted. In other words, also *setting would be recognized as a section header.

We need rules to ensure *** Settings ***, *** Test Case *** and so on is used, not *setting ;)

### LINKED ISSUE #152: Rule: Nested for loops are not supported
Nested for loop are not supported. You can only call keyword with another for loop inside your for loop.

### LINKED ISSUE #153: Rule: Setting name should be capitalized
Settings names like Suite Setup or Library are case insensitive. It should be reported as warning if used as not capitalized or upper case:
```
*** Settings ***
Suite Setup  Keyword
suite Setup  Keyword
SUITE SETUP  Keyword
```

bhirsz:
Closes #151 
Closes #153 
Closes #145 
Closes #140 
Partly relates to #150


bhirsz:
Closes #145

robocop/checkers/naming.py
```diff
@@ -58,48 +61,197 @@ def visit_KeywordName(self, node):  # noqa
...
+    """ Checker for keyword naming violations. """
     rules = {
         "0302": (
             "not-capitalized-keyword-name",
             "Keyword name should be capitalized",
             RuleSeverity.WARNING
+        ),
+        "0303": (
+            "keyword-name-is-reserved-word",
+            "'%s' is a reserved keyword%s",
```
mnojek, line 74: Shouldn't there be a space before injected second string?

robocop/checkers/naming.py
```diff
@@ -58,48 +61,197 @@ def visit_KeywordName(self, node):  # noqa
...
+                    )
+        elif self.check_if_keyword_is_reserved(keyword_name, node):
+            return
         words = keyword_name.replace('_', ' ').split(' ')
-        if any(not word.istitle() for word in words):
+        if any(not (word.istitle() or word.isupper()) for word in words if not self.is_variable(word)):
             self.report("not-capitalized-keyword-name", node=node)
+
+    def is_variable(self, name):
+        if len(name) < 3:
```
mnojek, line 167: I think it can be even `< 4`

robocop/checkers/naming.py
mnojek, line 74: Lacking space before injected str?

robocop/checkers/naming.py
```diff
@@ -58,48 +61,197 @@ def visit_KeywordName(self, node):  # noqa
...
+            return False
+        reserved_type = self.reserved_words[keyword_name.lower()]
+        suffix = self.prepare_reserved_word_rule_message(keyword_name, reserved_type)
+        self.report("keyword-name-is-reserved-word", keyword_name, suffix, node=node)
+        return True
+
+    @staticmethod
+    def prepare_reserved_word_rule_message(name, reserved_type):
+        return f". It must be in uppercase ({name.upper()}) when used as a marker with '{reserved_type}'. " \
+               f"Each marker should have minimum of 2 spaces as seperator." if reserved_type else ''
```
mnojek, line 182:
```suggestion
               f"Each marker should have minimum of 2 spaces as separator." if reserved_type else ''
```

robocop/checkers/naming.py
```diff
@@ -58,48 +61,197 @@ def visit_KeywordName(self, node):  # noqa
...
+class SettingsNamingChecker(VisitorChecker):
+    rules = {
+        "0306": (
+            "setting-name-not-capitalized",
+            "Setting name should be capitalized or upper case",
+            RuleSeverity.WARNING
+        ),
+        "0307": (
+            "section-name-invalid",
+            "Section name should should be in format *** Capitalized *** or *** UPPERCASE ***",
```
mnojek, line 194:
```suggestion
            "Section name should should be in format '*** Capitalized ***' or '*** UPPERCASE ***'",
```

robocop/checkers/naming.py
```diff
@@ -58,48 +61,197 @@ def visit_KeywordName(self, node):  # noqa
...
+        ),
+        "0304": (
+            "not-enough-whitespace-after-newline-marker",
+            "Provide at least two spaces after '...' marker",
+            RuleSeverity.ERROR
+        ),
+        "0305": (
+            "invalid-comment",
+            "Invalid comment. '#' needs to be first character in the cell. "
+            "For block comments you can use *** Comments *** section",
```
mnojek, line 85:
```suggestion
            "For block comments you can use '*** Comments ***' section",
```

robocop/checkers/naming.py
bhirsz: Last injected string can be empty - for 'while' and 'continue' reserved words (which don't have 'reserved_type' like for -> For loop, else -> Run keyword if etc).

robocop/checkers/naming.py
bhirsz: You're right, ${a} is already 4, and ${} is not variable (probably, I will test ;))

mnojek (approved):

PR #165, opened 2020-09-06
Add empty return rule and modify other return rules

### LINKED ISSUE #164: Rule: Empty [Return]

bhirsz:
Closes #164 

I also modified how the 'no keywords calls after return statement' works. I learned that [Return] only set variable to be returned and keywords past it are being executed. I changed severity to Warning and splitted "Return From Keyword" to separate rule (with severity Error).


robocop/checkers/misc.py
```diff
@@ -10,41 +10,64 @@
...
         "0901": (
             "keyword-after-return",
-            "Keyword call after %s statement",
+            "[Return] is not defined at the end of keyword. "
+            "Note that [Return] does not return from keyword but only set returned variables",
+            RuleSeverity.WARNING
+        ),
+        "0902": (
+            "keyword-after-return-from",
+            "Keyword call after Return From Keyword keyword",
```
mnojek, line 28:
```suggestion
            "Keyword call after 'Return From Keyword' keyword",
```

mnojek (approved):

PR #167, opened 2020-09-06
Remove register method

### LINKED ISSUE #163: Rules explorer should autoregister checker classes
Currently we need to provide register method inside file with checker classes. Instead our method for discovering checkers could also traverse class available in module and automatically register all with BaseChecker parent and non empty class attribue rules (to avoid registering utils classess).

### LINKED ISSUE #166: Remove register() method from checker files
Robocop needs register() methods in files with checkers in order to 'register' checker classes. It's easy to forgot to add your class to register method. It should be done automatically in checkers init() method. 

Only register classes with BaseChecker parent and non empty class attribue rules (to avoid registering utils classess).

bhirsz:
Closes #163


mnojek (approved):

mnojek:
Great change!

PR #173, opened 2020-09-07
Ignore path matching pattern

### LINKED ISSUE #17: Allow to ignore file
Add a way to ignore particular files ie given directory. Support glob patterns:

tests/
  test.robot
  resource.robot
  test2.robot
  othersuite.robot

If robocop parse whole directory with --ignore test*.robot it should only run tests on resource.robot and othersuite.robot

bhirsz:
Closes #17


robocop/config.py
```diff
@@ -2,6 +2,7 @@
 import argparse
 import re
 import fnmatch
+from pathlib import Path
```
mnojek, line 5: I think specific module imports should be located under general import instructions

mnojek (approved):

robocop/config.py
bhirsz: Actually I've checked and this import statement was no longer used - I already pass Path() object to config so I didn't need to create new Path() object. Removed the import

PR #179, opened 2020-09-08
Add rule for checking if keyword name contains underscores

### LINKED ISSUE #150: Keyword name should be capitalized is too strict

mnojek:
Closes #150

### DIFF
--- a/robocop/checkers/naming.py
+++ b/robocop/checkers/naming.py
@@ -78,6 +78,11 @@ class KeywordNamingChecker(VisitorChecker):
             "Invalid comment. '#' needs to be first character in the cell. "
             "For block comments you can use '*** Comments ***' section",
             RuleSeverity.ERROR
+        ),
+        "0306": (
+            "underscore-in-keyword-name",
+            "Underscores in keyword name can be replaced with spaces",
+            RuleSeverity.WARNING
         )
     }
     reserved_words = {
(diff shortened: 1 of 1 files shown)


robocop/checkers/naming.py
```diff
@@ -143,7 +148,10 @@ def check_keyword_naming(self, keyword_name, node):  # noqa
         if keyword_name.startswith('...'):
             self.report("not-enough-whitespace-after-newline-marker", node=node)
             return
-        if normalize_robot_name(keyword_name) == 'runkeywordif':
+        normalized_name = normalize_robot_name(keyword_name)
+        if normalized_name != keyword_name.replace(' ', '').lower():
```
bhirsz, line 152: how about simple 
```
if '_' in keyword_name:
```

mnojek: People sometimes like to complicate easy things 😉

bhirsz (approved):

PR #191, opened 2020-09-09
Move comment rule

bhirsz:
And also change col offset in error rule

### DIFF
--- a/robocop/checkers/comments.py
+++ b/robocop/checkers/comments.py
@@ -5,36 +5,8 @@ from robocop.checkers import VisitorChecker
 from robocop.rules import RuleSeverity
 
 
-class CommentBaseChecker(VisitorChecker):
-    """
-    This is base class for comment checkers. Not supposed to be used directly.
-    """
-    rules = {}
-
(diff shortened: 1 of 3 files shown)


robocop/checkers/comments.py
```diff
@@ -5,19 +5,36 @@
 from robocop.rules import RuleSeverity
 
 
-class CommentBaseChecker(VisitorChecker):
```
bhirsz, line 8: I removed CommentBaseChecker since we only had one class that used it

robocop/checkers/comments.py
```diff
@@ -5,19 +5,36 @@
...
-    """
-    This is base class for comment checkers. Not supposed to be used directly.
-    """
-    rules = {}
+class CommentChecker(VisitorChecker):
+    """ Checker for content of comments. It detects invalid comments or leftover todo or fixme in code. """
```
mnojek, line 9:
```suggestion
    """ Checker for content of comments. It detects invalid comments or leftover `todo` or `fixme` in code. """
```

mnojek (approved):

PR #202, opened 2020-09-17
Update info

mnojek:
Changes:
* added metavar for config options
* added badges to readme file
* updated links to GitHub project
* updated names for alt attribute of images
* moved Documentation higher in the readme file

### DIFF
--- a/robocop/config.py
+++ b/robocop/config.py
@@ -138,7 +138,7 @@ class Config:
                                          formatter_class=argparse.RawTextHelpFormatter,
                                          description='Static code analysis tool for Robot Framework',
                                          epilog='For full documentation visit: '
-                                                'https://github.com/bhirsz/robotframework-robocop',
+                                                'https://github.com/MarketSquare/robotframework-robocop',
                                          add_help=False)
         required = parser.add_argument_group(title='Required parameters')
         optional = parser.add_argument_group(title='Optional parameters')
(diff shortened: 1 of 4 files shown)


robocop/config.py
```diff
@@ -138,17 +138,17 @@ def _create_parser(self):
...
+                                                'https://github.com/MarketSquare/robotframework-robocop',
                                          add_help=False)
         required = parser.add_argument_group(title='Required parameters')
         optional = parser.add_argument_group(title='Optional parameters')
 
         required.add_argument('paths', metavar='paths', type=str, nargs='*', default=['.'], help=self.HELP_MSGS['help_paths'])
 
         optional.add_argument('-i', '--include', action=ParseDelimitedArgAction, default=self.include,
-                              help=self.HELP_MSGS['help_include'])
+                              metavar='RULES', help=self.HELP_MSGS['help_include'])
```
bhirsz, line 149: what does metavar does, it is kind of extra documentation or method of grouping our vars?

bhirsz (approved):

robocop/config.py
mnojek: It just changes the display name of the option variable in the help message.
So for example instead of:
`--include INCLUDE`
it is:
`--include RULES`
By default the name of the variable is the same as long option name.

PR #227, opened 2020-11-13
Enable all reports by using all

### LINKED ISSUE #204: --report all to enable all reports
For lazy typers it would be nice to have "all" option to include all reports instead of typing their names separetly

bhirsz:
closes #204

### DIFF
--- a/robocop/config.py
+++ b/robocop/config.py
@@ -80,7 +80,11 @@ class Config:
         'help_exclude':     'Ignore specified rules. You can define rule by its name or id.\n'
                             'Glob patterns are supported',
         'help_ext_rules':   'List of paths with custom rules',
-        'help_reports':     'Run reports',
+        'help_reports':     'Generate reports after scan. You can enable reports by listing them in comma\n'
+                            'separated list:\n'
+                            '--reports rules_by_id,other_report\n'
+                            'To enable all reports use all:\n'
+                            '--report all',
         'help_format':      'Format of output message. '
                             'You can use placeholders to change the way an issue is reported.\n'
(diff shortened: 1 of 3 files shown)


robocop/config.py
```diff
@@ -80,7 +80,11 @@ def __init__(self):
         'help_exclude':     'Ignore specified rules. You can define rule by its name or id.\n'
                             'Glob patterns are supported',
         'help_ext_rules':   'List of paths with custom rules',
-        'help_reports':     'Run reports',
+        'help_reports':     'Generate reports after scan. You can enable reports by listing them in comma\n'
+                            'separated list:\n'
+                            '--reports rules_by_id,other_report\n'
```
mnojek, line 85: What about displaying in the help message all available options? It would be very useful instead of this example. The instruction about providing them in comma-separated list is enough, I think.

mnojek (changes requested):

robocop/config.py
bhirsz: Should it be hardcoded or dynamically loaded? The latter would be probably bit too complex for this (especially since we only have 2 reports right now...), and the first could became obsolete if any report would be added/modified

Final version:
```
        'help_reports':     'Generate reports after scan. You can enable reports by listing them in comma\n'
                            'separated list:\n'
                            '--reports rules_by_id,rules_by_error_type\n'
                            'To enable all reports use all:\n'
                            '--report all',
```

mnojek: I think your code suggestion is fine now 😃 And I think that it does not need to be dynamically loaded. It's common that the CLI text needs to be updated when new features are introduced to the software.

PR #235, opened 2020-11-16
Improve rule 0509

### LINKED ISSUE #234: 0509 fails at Comments section
**Describe the bug**
I started to play with [issue #174](https://github.com/MarketSquare/robotframework-robocop/issues/174) and I ran into an issue.
If data is defined before the first data section, that part counts as `Comments` section (lines are considered as `COMMENT` tokens).
Currently, the mentioned `0509 Section is empty` rule will fail for `Comments` section unless there is only one line which starts with `robocop:` string due to the `check_if_empty` function.

```python
for child in node.body:
    if isinstance(child, Comment):
        if 'robocop:' in child.tokens[0].value and len(node.body) == 1:
            break
        elif not isinstance(child, EmptyLine):
            break
    else:
        self.report("empty-section", node=node)
```

**To Reproduce**
You can reproduce the issue with this test code:

```robotframework
Data before the first section is ignored in Robot Framework.
Second line in Comment section.

*** Settings ***
Documentation    Dummy documentation.


*** Test Cases ***
Dummy Test Case
    [Documentation]    Dummy documentation.
    No Operation
```

This one will pass.

```robotframework
robocop:Data before the first section is ignored in Robot Framework.
*** Settings ***
Documentation    Dummy documentation.


*** Test Cases ***
Dummy Test Case
    [Documentation]    Dummy documentation.
    No Operation
```

**Expected behavior**
I suppose we should handle `Comments` section like the others.

**Additional context**
Is there any particular reason why the `check_if_empty` function was implemented this way?
If this is a bug, it should be an easy fix by removing the `if 'robocop: ...` part, I can do a pull request for it.

szabolcstoth:
- If the section is Comments, just check if there is anything
  except EmptyLines.
- If the section is any other section, it should trigger warning if
  there are only EmptyLines and Comments.

Closes #234

### DIFF
--- a/robocop/checkers/lengths.py
+++ b/robocop/checkers/lengths.py
@@ -1,6 +1,7 @@
 """
 Lengths checkers
 """
+from robot.parsing.model.blocks import CommentSection
 from robot.parsing.model.statements import KeywordCall, Comment, EmptyLine, Arguments
 from robocop.checkers import VisitorChecker, RawFileChecker
 from robocop.rules import RuleSeverity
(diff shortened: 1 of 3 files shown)


bhirsz (approved):
Great! After you merge I will release new version of Robocop - together with other changes that were merged recently

PR #243, opened 2020-11-19
Add rule related to ignored data

### LINKED ISSUE #174: Rule: Check if there is data before first section
Any data in a file that is located before the first section is ignored in RF. I think it may be useful to catch such cases and report that it is not contained by any section thus ignored by RF.

szabolcstoth:
Added rule "0704": checker for ignored data.

Closes #174

### DIFF
--- a/robocop/checkers/comments.py
+++ b/robocop/checkers/comments.py
@@ -1,7 +1,7 @@
 """
 Comments checkers
 """
-from robocop.checkers import VisitorChecker
+from robocop.checkers import RawFileChecker, VisitorChecker
 from robocop.rules import RuleSeverity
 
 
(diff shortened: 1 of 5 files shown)


robocop/checkers/comments.py
```diff
@@ -58,3 +58,34 @@ def check_comment_content(self, comment_token):
...
+
+    def __init__(self, *args):
+        self.section_header = False
+        self.ignored_data = False
+        super().__init__(*args)
+
+    def parse_file(self):
+        with open(self.source) as file:
+            for lineno, line in enumerate(file, 1):
+                self.check_line(line, lineno)
```
bhirsz, line 81: to reduce amount of code you can return boolean value from this function:

```
    def parse_file(self):
        with open(self.source) as file:
            for lineno, line in enumerate(file, 1):
                if self.check_line(line, lineno)
                    break

        if line.startswith('***'):
            return True
        if not line.startswith('***') and not line.startswith('# robocop:'):
            self.report("ignored-data", lineno=lineno, col=0)
           return True
```
(init not needed in case). But code clarity will worsen a bit - your call

tests/atest/rules/ignored-data/test.robot
```diff
@@ -0,0 +1,11 @@
+# robocop: disable=1001
```
bhirsz, line 1: you can also create more robot files inside "ignored-data" directory, it scan whole directory. For example happy-path.robot (without ignored data), and the one without # robocop: disable line

robocop/checkers/comments.py
szabolcstoth: Agree, I refactored the code.

tests/atest/rules/ignored-data/test.robot
szabolcstoth: Agree, I added more test scenarios. ;)

Issue #169, opened 2020-09-06
Support JSON output format


bhirsz:
It will be needed for #168. It could be also useful for integrating Robocop with SonarQube which can also use JSON formatted values.

bhirsz:
Closed with #252

Issue #251, opened 2021-02-04
API to treat output results outside of robocop


andreagubellini:
*Feature*:
It would be useful to implement an API that returns robocop output results.

For example:

```
from robocop.run import run_robocop

test = run_robocop()
```

In this specific scenario it would be very useful to retrieve the output in order to manage data outside of the robocop context (for example reports, badges and more).

bhirsz:
Sounds good! And it will be usefull for other planned features. There are different ways of implementing this though:

1) We can just return list of formatted messages after run. Easiest to implement but not that usefull:
```
issues = run_robocop()
print(issues)
['C:\OCP_project\tests\test.robot:7:0 [W] 0509 Section is empty',
'C:\OCP_project\tests\test.robot:22:0 [E] 0801 Multiple test cases with name "Simple Test" in suite']
```

2) We can use internal `Message` class to return list of messages:
https://github.com/MarketSquare/robotframework-robocop/blob/master/robocop/rules.py#L116
```
issues = run_robocop()
print(issues)
[Message(), Message()]

3) Implement special class for returning issues (simple dataclass or similiar with msg, severity, source path and line numbers)

4) Return json like output. It's possible to create report for that
```
issues = run_robocop()
print(issues)
[{'source': <>, 'msg': 'Stuff', 'lineno': 112 ...}, {}]
```

In 1-3 cases we can reuse reports - since they already consume our messages and could just accumulate messages in our desired format. We would just need to add method to robocop for retrieving the results. https://github.com/MarketSquare/robotframework-robocop/blob/master/robocop/reports.py
ie
```
from robocop.run import Robocop

robocop = Robocop()
robocop.run()
data = robocop.get_results() # or whatever
```

andreagubellini:
alright!

I can try working on this on my spare time, if you want

bhirsz:
Sure, go ahead :)

andreagubellini:
@bhirsz I opened a #252 work-in-progress pull request to show you a first implementation of this task.

Here's an example of the returned output:

[...]

PR #252, opened 2021-02-05
add export_results()

andreagubellini:
WIP

#251


robocop/run.py
```diff
@@ -32,6 +33,26 @@ def __init__(self, from_cli=False):
         self.load_reports()
         self.configure_checkers_or_reports()
 
+    def export_results(self, *issues):
+        """ Gather results from issues and reports, merge them in a list and return the data"""
+        payload = []
+        for report in self.reports:
+            output = report.get_report()
+            if output is not None:
+                payload.append(output)
```
bhirsz, line 42: Do we need report output in payload? Because it will be inconsisted, ie
```
# issues alone 
[{'source': '/home/andrea/Documents/black-manta/test.robot', 'line': 0, 'col': 0, 'severity': 'W', 'rule_id': '0203', 'desc': 'Missing documentation in suite', 'msg_name': 'missing-doc-suite'}, {'source': '/home/andrea/Documents/black-manta/test.robot', 'line': 1, 'col': 0, 'severity': 'W', 'rule_id': '0509', 'desc': 'Section is empty', 'msg_name': 'empty-section'}, {'source': '/home/andrea/Documents/black-manta/test.robot', 'line': 1, 'col': 0, 'severity': 'W', 'rule_id': '1002', 'desc': 'Missing trailing blank line at the end of file', 'msg_name': 'missing-trailing-blank-line'}]
# and then reports
['Issues by ids:
W1007 (uneven-indent)                : 5
E0904 (nested-for-loop)              : 4
W0302 (not-capitalized-keyword-name) : 4

Found 12 issue(s): 9 WARNING(s), 4 ERROR(s).']
```
After receiving such list you would need to recognize and parse what's inside. It would be better to have only issues in output and then additionaly you can call report and get their output. Optionally reports could report json-like output to ie:
```
data = report.get_json_report()
print(data)
['W1007 (uneven-indent)': 5, 'E0904 (nested-for-loop)': 4]
# or different report
['total': 12, 'error': 4, 'warning': 8]
```

robocop/run.py
```diff
@@ -32,6 +33,26 @@ def __init__(self, from_cli=False):
...
+        for issue in self.issues:
+            issue = {
+                "source": issue.source,
```
bhirsz, line 45: You can add serializing method to Message class so it would return dict:
```
class Message:
    (...)
    def to_json(self):
        return {
            "source": self.source,
            # and so on
        }
```

robocop/run.py
```diff
@@ -91,6 +113,7 @@ def run_checks(self):
                 checker.issues.clear()
             found_issues.sort()
             for issue in found_issues:
+                self.issues.append(issue)
```
bhirsz, line 116: It would collect issues every time, even if we don't need it (for example for normal cli run). What about creating report that would collect issues, parsing them to json like you did? For that:
1) It would be better to keep reports as dict not a list so it's easier to access them (I can do that though, so we can stick to current implementation for now)
2) Create report inside reports 
```
class JsonReport(Report):
    def __init__(self):
        self.name = 'JsonReport'
        self.issues = []

    def add_message(self, message, **kwargs):  # pylint: disable=unused-argument
        self.issues.append(message.to_json()) # granted you add to_json() method to message
        
     def get_report(self):
         return ''
```

You can add then this method for retrieving issues to Robocop class:
```
def export_results(self): # not sure about naming
    for report in self.reports:
        if report.name == 'JsonReport':
            return report.issues
  
        
```

bhirsz: There could be some gotchas or we would need to automatically register this report to get it working

robocop/run.py
```diff
@@ -32,6 +33,11 @@ def __init__(self, from_cli=False):
         self.load_reports()
         self.configure_checkers_or_reports()
 
+    def get_json_results(self, *issues):
+        """ Gather results from issues and reports, merge them in a list and return the data"""
+        payload = [issue.to_json() for issue in self.issues]
+        return(payload)
```
bhirsz, line 39:
```suggestion
        return [issue.to_json() for issue in self.issues]
```

bhirsz:
@andreagubellini hi, what's the status? If there is review or update from my side needed let me know - I could forgot about it in meantime :(

andreagubellini:
> @andreagubellini hi, what's the status? If there is review or update from my side needed let me know - I could forgot about it in meantime :(

Hey!

I've finished to implement all the requested changes, now we just need a method (or a run() refactor) to return the output. I believe you said you would do that, do you want me to take a look or?

bhirsz:
@andreagubellini Ok now I remember the whole context. Since I'm doing some updates in Robocop over the weekend I will also take care of that and I will let you know when its done

bhirsz:
I've decided to update some of the core code to support this feature. I've created JsonReport report class which is accumulating issues in json format (but only for non cli runs). And of course I've added tests. 

After this change if you want to use Robocop in python code you can:
```
        import robocop

        config = Config()
        config.include = {'1003'}
        config.paths = ['tests\\atest\\rules\\section-out-of-order']

        issues = robocop_runner.run()
        print(issues)
```
run() returns list of issues in json format:
```
[{'col': 0,
  'desc': 'Invalid number of empty lines between sections (0/2)',
  'line': 17,
  'rule_id': '1003',
  'rule_name': 'empty-lines-between-sections',
  'severity': 'W',
  'source': 'C:\\Users\\barte\\Desktop\\repo\\forks\\robotframework-robocop\\tests\\atest\\rules\\section-out-of-order\\test.robot'},
 {'col': 0,
  'desc': 'Invalid number of empty lines between sections (1/2)',
  'line': 20,
  'rule_id': '1003',
  'rule_name': 'empty-lines-between-sections',
  'severity': 'W',
  'source': 'C:\\Users\\barte\\Desktop\\repo\\forks\\robotframework-robocop\\tests\\atest\\rules\\section-out-of-order\\test.robot'}]
```

@mnojek  @andreagubellini

robocop/reports.py
```diff
@@ -41,7 +41,7 @@ def __init__(self):
         self.name = 'rules_by_id'
         self.message_counter = defaultdict(int)
 
-    def add_message(self, message, **kwargs):  # noqa
```
bhirsz, line 44: Those kwargs were left 'just in case this method will be extended in the future' but it doesn't make sense to keep it that way.

robocop/rules.py
```diff
@@ -137,3 +137,14 @@ def __lt__(self, other):
...
+
+    def to_json(self):
+        return {
+                "source": self.source,
+                "line": self.line,
+                "col": self.col,
+                "severity": self.severity.value,
+                "rule_id": self.rule_id,
+                "desc": self.desc,
+                "name": self.name
```
bhirsz: I'm still not 100% sure about the naming here but it looks ok so we can go with it.

robocop/run.py
```diff
@@ -16,21 +16,40 @@
 
 
 class Robocop:
-    def __init__(self, from_cli=False):
+    """
+    Main class for running the checks.
+
+    If you want to run checks with non default configuration create your own ``Config`` and pass it to ``Robocop``.
```
bhirsz, line 22: Suggestions how I could write it better are welcome :)

andreagubellini:
Looks fine to me!

bhirsz:
Closes #169

robocop/rules.py
```diff
@@ -137,3 +137,14 @@ def __lt__(self, other):
...
         return f"{self.severity.value}{self.rule_id} ({self.name})"
+
+    def to_json(self):
+        return {
+                "source": self.source,
+                "line": self.line,
+                "col": self.col,
+                "severity": self.severity.value,
+                "rule_id": self.rule_id,
+                "desc": self.desc,
```
mnojek, line 148: Hmm, why not to put here the whole word (`description`)?

robocop/run.py
```diff
@@ -16,21 +16,40 @@
...
+    Use ``Robocop.run()`` method to start analysis. If ``from_cli`` is set to False it will return
+    list of found issues in json format.
```
mnojek, line 24:
```suggestion
    list of found issues in JSON format.
```

robocop/run.py
```diff
@@ -16,21 +16,40 @@
 
 
 class Robocop:
-    def __init__(self, from_cli=False):
+    """
+    Main class for running the checks.
+
+    If you want to run checks with non default configuration create your own ``Config`` and pass it to ``Robocop``.
+    Use ``Robocop.run()`` method to start analysis. If ``from_cli`` is set to False it will return
```
mnojek, line 23:
```suggestion
    Use ``Robocop.run()`` method to start analysis. If ``from_cli`` is set to ``False`` it will return
```

robocop/rules.py
andreagubellini: you mean something like `"description": self.desc,` right?

bhirsz: Not Mateusz but I will answer: yes exactly. If you want to update it you will need also to update tests (under tests / utests/test_reports.py)

mnojek: Yes. Same goes with the "col" key. I think we don't need to save space by reducing the names. Would be more clear for everyone that "desc" is "description" and "col" is "column" IMO.

PR #257, opened 2021-02-25
Move pylama checks to separate github action

### LINKED ISSUE #256: Separate pylama from unit test runs
There is no need for running pylama together with unit tests. Right now it's executed on every os * every python version  * every rf version even if Python code is the same.

bhirsz:
Closes #256

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,27 @@
+# This workflow will run pylama checks on PRs against master
+
+name: Linter
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


bhirsz:
I opted for manual include list instead of using array matrix. That way instead of 12 builds (with redundant settings) we have 5 builds:
- one build on windows-latest
- one build for 3.2* RF version
- builds on ubuntu-latest, RF 4.0 and 3.7, 3.8, 3.9 python versions

mnojek (approved):

PR #260, opened 2021-02-26
Add new rule for checking too many blank lines at EOF

### LINKED ISSUE #250: Robocop can't detect too many empty lines at the end of file.
Checker `missing-trailing-blank-line` can only detect the lack of blank line at the end of file but it can't detect if there are too many of them. It's not exactly a bug, but I believe I wanted it to work like that when writing this.

mnojek:
Closes #250 

- Added new rules for checking too many blank lines at the end of the file
- Added one more case to trailing whitespace atest

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -28,6 +28,11 @@ class InvalidSpacingChecker(RawFileChecker):
             "missing-trailing-blank-line",
             "Missing trailing blank line at the end of file",
             RuleSeverity.WARNING
+        ),
+        "1010": (
+            "too-many-trailing-blank-lines",
+            "Too many blank lines at the end of file",
+            RuleSeverity.WARNING
         )
     }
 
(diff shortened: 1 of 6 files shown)


robocop/checkers/spacing.py
```diff
@@ -38,8 +43,12 @@ def __init__(self, *args):
     def parse_file(self):
         self.lines = []
         super().parse_file()
-        if self.lines and not self.lines[-1].endswith('\n'):
-            self.report("missing-trailing-blank-line", lineno=len(self.lines), col=0)
+        if self.lines:
+            last_line = self.lines[-1]
+            if not last_line.endswith('\n'):
+                self.report("missing-trailing-blank-line", lineno=len(self.lines), col=0)
+            if last_line == '\n':
```
bhirsz, line 50: what happens for this case (extra spaces before new line char):
```
    Keyword\n
    \n
```

mnojek: Probably it fails :) Good catch. I will cover this case.

bhirsz: Just in case - you can add several files to single atest (for example file with trailing line, file without, file with spaces and \n..). expected_output will then contain merged issues from all files.

bhirsz (approved):

Issue #161, opened 2020-09-06
Config files for modyfing Robocop options per directory


bhirsz:
Robocop could support config files. Many popular cli tools does support configuration files. Global options provided from cli can be overwritten by local file with specific extension like ".robocop". It would allow to use different set of settings for given directory. Hovewer not all Robocop options could be supported that way. Example:

```
[disable-rules]
some-rule
0102
other-rule
```
It would not report on given rules. Including rules would be not possible with our current architecture since rules ignored at the start are often removed altogether from the Robocop during the run and cannot be accesssed later. 

```
[main]
message-template={source}:{line}:{col} [{severity}] {rule_id} {desc}
ignore_files=*.resource
threshold=W
filetypes=.robot,.txt
recursive=False
report=some_report  # reports are aggregated globally, so this issue would mostly disable sending messages to reports for given directory
```

```
[configure-rules]
some-rule:max_count=10
```

To implement this every file should probably carry instance of config class (None if no configuration file found in same directory).  If the config is None the global one should be used, otherwise some option should temporary overwrite global options.

mnojek:
Will it be possible to set some options that are not possible to set using command line interface?

bhirsz:
If there is need for it - we need to thinkg which options are needed and which cannot be used in config files. If you have some proposals go ahead. Parsing config files will be bit different (yet it will be connected to some variables from Config class if it's possible)

mnojek:
I think I will give it a thought after 1.0.0 is released but nonetheless it's a good idea.

PR #264, opened 2021-02-27
Load arguments from .robocop file

### LINKED ISSUE #253: Default config file
**Is your feature request related to a problem? Please describe.**
There is no simple and unified way for a team to run the same set of linter rules (other than copy/pasting a long list of exclusions) 

**Describe the solution you'd like**
Introduce a default config file that will allow a team to configure their linter (this will allow team members to run a simple command and be unified with the rest of the team) 

**Describe alternatives you've considered**
none, sorry 

**Additional context**
image

bhirsz:
Closes #253 

If no arguments are provided in command line Robocop will try to find '.robocop' file and load it from there. It will start looking from current directory and go up until it found it or '.git' file is found.

`.robocop` file use the same syntax as CLi arguments:
```
--include rule
--report all
src/to/be/scanned.robot
```


bhirsz:
It's bit different but I think it closes #161 .

mnojek (approved):

PR #268, opened 2021-02-28
Variable naming rules

mnojek:
Added 2 new rules:

- Detect if variables defined in variables section are uppercase
- Detect if variables set using `Set Test Variable`, `Set Task Variable`, `Set Suite Variable`, `Set Global Variable` are uppercase


robocop/checkers/naming.py
```diff
@@ -227,3 +229,44 @@ def visit_Documentation(self, node):  # noqa
...
+                            col=token.col_offset)
+
+    def visit_KeywordCall(self, node):  # noqa
+        if not node.keyword:
+            return
+        if normalize_robot_name(node.keyword) in self.set_variable_variants:
+            if len(node.data_tokens) < 2:
+                return
+            token = node.data_tokens[1]
+            if token.type == Token.ARGUMENT and not token.value.isupper():
```
bhirsz, line 271: Are numbers in variable names ignored? Ie ${VERSION2}

bhirsz (approved):

robocop/checkers/naming.py
mnojek: Yes, they are. Is it OK?

bhirsz: Yeah that's good then

bhirsz (approved):

PR #273, opened 2021-03-07
Add # noqa as alternative to # robocop: disable

### LINKED ISSUE #269: Support # noqa as alternative to # robocop: disable

bhirsz:
Closes #269

### DIFF
--- a/robocop/utils/disablers.py
+++ b/robocop/utils/disablers.py
@@ -36,22 +36,12 @@ class DisablersFinder:
         """
         if not self.any_disabler:
             return False
-        if 'all' in self.rules:
-            disabled = self.is_line_disabled(rule_msg.line, 'all')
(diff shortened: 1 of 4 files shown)


robocop/utils/disablers.py
```diff
@@ -74,14 +66,16 @@ def _parse_file(self, source):
 
     def _parse_line(self, line, lineno):
         statement, comment = line.split('#', maxsplit=1)
+        if '# noqa' in line:
```
mnojek, line 69: What about `#noqa`? Shouldn't it be supported?

bhirsz: We could support it but it would be rule volation - comments needs space after # (missing-space-after-comment)

mnojek (approved):

PR #277, opened 2021-03-12
Fix mixed-tabs-and-spaces to work with multiple files

### LINKED ISSUE #274: mixed-tabs-and-spaces works only on first found issue
When Robocop is executed on more than 1 file (ie 1000) that contains mixed-tabs-and-spaces issue, only first time is reported.

bhirsz:
Closes #274

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -193,17 +193,19 @@ class InconsistentUseOfTabsAndSpacesChecker(VisitorChecker, ModelVisitor):
     }
 
     def __init__(self, *args):
-        self.found = False
-        self.tabs = False
-        self.spaces = False
+        self.found, self.tabs, self.spaces = False, False, False
         super().__init__(*args)
+
+    def visit_File(self, node):  # noqa
+        self.found, self.tabs, self.spaces = False, False, False
+        super().visit_File(node)
 
     def visit_Statement(self, node): # noqa
         if self.found:
             return
         for token in node.get_tokens(Token.SEPARATOR):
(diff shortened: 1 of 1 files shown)


robocop/checkers/spacing.py
```diff
@@ -193,17 +193,19 @@ class InconsistentUseOfTabsAndSpacesChecker(VisitorChecker, ModelVisitor):
...
+        self.found, self.tabs, self.spaces = False, False, False
+        super().visit_File(node)
+
     def visit_Statement(self, node): # noqa
         if self.found:
             return
         for token in node.get_tokens(Token.SEPARATOR):
-            self.tabs = True if '\t' in token.value else self.tabs
-            self.spaces = True if ' ' in token.value else self.spaces
+            self.tabs = '\t' in token.value or self.tabs
```
bhirsz, line 207: simplified condition

robocop/checkers/spacing.py
```diff
@@ -193,17 +193,19 @@ class InconsistentUseOfTabsAndSpacesChecker(VisitorChecker, ModelVisitor):
...
 
     def __init__(self, *args):
-        self.found = False
-        self.tabs = False
-        self.spaces = False
+        self.found, self.tabs, self.spaces = False, False, False
         super().__init__(*args)
 
+    def visit_File(self, node):  # noqa
+        self.found, self.tabs, self.spaces = False, False, False
```
bhirsz, line 200: without that, if the issues was found in first file it would stop checking for all other files

mnojek (approved):

PR #287, opened 2021-03-22
LSP integration

bhirsz:
It should be now possible to:
```
# create config class which will autoloads .robocop config file

config = Config(root='C:\repository\robot_stuff')
```
Run scan on ast model:
```
import robocop

(...)
issues = robocop_runner.run_check(ast_model, filename, source)
```

List of issues can be converted to LSP friendly diagnostic:
```
from robocop.utils import issues_to_lsp_diagnostic

issues = issues_to_lsp_diagnostic(issues)
```
(Note that it's not full support for Diagnostic yet because we don't preserve the full range, only starting line and col. In the future we need to update our rules reporting with that information).

### DIFF
--- a/robocop/config.py
+++ b/robocop/config.py
@@ -54,8 +54,9 @@ class SetListOption(argparse.Action):
 
 
 class Config:
-    def __init__(self):
+    def __init__(self, root=None):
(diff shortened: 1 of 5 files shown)


bhirsz:
Documentation needs to be added

PR #288, opened 2021-03-23
Add tests and docs for Robocop API

### LINKED ISSUE #168: Add client-server mode with API
Client-server mode with API could be used by IDEs or separate programs for file scanning "on the fly" without need to wrap Robocop. 

Example:
```
robocop = Robocop.server(*args)
robocop.scan_file(file)
robocop.scan_file(file)
...
robocop.scan_file(file)
```

Returned output would need to be different than std.out (for example json object) so it would be handled by other tools.

bhirsz:
Closes #168

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,53 @@
+.. _api:
+
+API
+===
+
+``robocop`` package can be imported and used directly in Python code.
+You need to create instance of ``Config`` class that will contain your configuration::
+
+    import robocop
+
+    config = robocop.Config()
+    config.include = {'1003'}
+    config.paths = ['tests\\atest\\rules\\section-out-of-order']
+
+``Config`` can also load configuration automatically from ``.robocop`` file if it exists in the project root::
+
+    import robocop
+
+    config = robocop.Config(root='C:\your_project\')
+
+Following ``Config`` is used to create ``Robocop`` class::
+
+    robocop_runner = robocop.Robocop(config=config)
+
+It is recommended to reload configuration before running the checks (it should register rules and reports based on your
+configuration)::
+
+    robocop_runner.reload_config()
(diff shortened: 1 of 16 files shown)


docs/api.rst
```diff
@@ -0,0 +1,53 @@
+.. _api:
```
bhirsz, line 1: Far from perfect but we can start with it - there are basic examples that should be sufficient.

robocop/utils/utils.py
```diff
@@ -69,11 +69,11 @@ def issues_to_lsp_diagnostic(issues):
     return [{
         'range': {
             'start': {
-                'line': issue.line,
+                'line': issue.line - 1,
```
fabioz, line 72: This needs to be `'line': issue.line - 1 if issue.line > 0 else 0` (it seems there are some cases where 0 is reported and it became negative... in particular the: `Missing documentation in suite`).

docs/api.rst
```diff
@@ -0,0 +1,53 @@
...
+    robocop_runner.reload_config()
+
+
+Scan files supplied in configuration
+------------------------------------
+To scan all files provided in ``Config`` invoke::
+
+    robocop_runner.run()
+
+``run()`` method by default return issues in JSON format.
```
mnojek, line 37:
```suggestion
``run()`` method by default returns issues in JSON format.
```

docs/api.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Scan existing robot ast model
+-----------------------------
+If you have ast model loaded before you can use ``robocop`` to scan it::
```
mnojek, line 41:
```suggestion
If you have ast model loaded before, you can use ``robocop`` to scan it::
```

PR #298, opened 2021-03-25
Fix uneven-indent rule

### LINKED ISSUE #296: Uneven-indent rule reports comments as rule violation

### LINKED ISSUE #297: Uneven-indent rule ignores If blocks

bhirsz:
Closes #296 
Closes #297


tests/atest/rules/uneven-indent/test.robot
```diff
@@ -74,3 +103,22 @@ Keyword With Uneven NewLines
     Keyword 1
     Keyword
       ...  ${2}
+
+Keyword With Assignments
+    ${arg}    ${arg}    Keyword
+    ...  multiline
+    IF    ${condition}
+        ${value}    Correct Indent
+          ${arg}  Incorrent Indent
```
mnojek, line 112: Do we want to force 4 spaces always inside nested structures?

bhirsz: No, we don't do it. The indent checker check for two things: whether you have correct indent (so if you're inside IF or FOR block your indent should be at least +1 from FOR/IF header) and whether all indents in block are the same. 
For example:
```
    Line
      OverIndent
    Line
```
The most common indent is used by Line keyword, so the Overindent keyword will throw warning. 
In case there are only two lines (like in above block) the first one will be labelled as "correct" and the second will be compared to it. That's why we have overindent warning in line 112.

mnojek (approved):

tests/atest/rules/uneven-indent/test.robot
mnojek: Then you have my upvote, sir.

PR #300, opened 2021-03-25
Fix file_stats report

### LINKED ISSUE #299: File statistic report count files with issues that were excluded from CLI
If I run Robocop for ie 10 files, while 5 of the contains issues -> but issues from 3 files are excluded (--exclude, --treshold etc), file_stats report should print "Processed 10 file(s) from which 2 file(s) contained issues" but it does print "Processed10 file(s) from which 5 file(s) contained issues".

bhirsz:
Closes #299

### DIFF
--- a/tests/atest/rules/misaligned-continuation/tests.robot
+++ b/tests/atest/rules/misaligned-continuation/tests.robot
@@ -35,3 +35,8 @@ Misaligned headers
       ...  ${2}
         Keyword
     END
+
+Misaligned Arguments
+    [Arguments]    ${arg}
+   ...  ${arg}
+    Keyword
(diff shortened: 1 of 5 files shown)


tests/atest/rules/misaligned-continuation/tests.robot
```diff
@@ -34,4 +34,9 @@ Misaligned headers
     FOR  ${value}  IN  ${1}
       ...  ${2}
         Keyword
-    END
\ No newline at end of file
+    END
```
bhirsz, line 37: Not connected to PR topic but I added some test data for one of the rules

mnojek (dismissed):

mnojek (approved):

PR #303, opened 2021-03-25
Fix Robocop not using configuration when running using API

### LINKED ISSUE #302: .robocop not picked up when using API
When using the API as:

```
    import robocop
    from robocop.utils import issues_to_lsp_diagnostic

    config = Config(root=project_root)
    robocop_runner.reload_config()

    issues = robocop_runner.run_check(ast_model, filename, source)
    diag_issues = issues_to_lsp_diagnostic(issues)
    return diag_issues
```

The `.robocop` file is not being loaded based on the `project_root`.

I debugged a bit and it seems like `load_default_config_file` is never called. I tried to manually call: `config.parse_opts([])`, right after creating the `Config` and it seems to pick things up on the debugger, but I still couldn't get it to ignore what I expected.

I tested with a file with:

```
*** Test Cases ***
Test
    Fail
```

and a `.robocop` file with:

```
--exclude missing-doc-testcase
--include missing-doc-suite
```

But the missing doc in the testcase is still reported.

So, I'm waiting on some pointers on what I'm doing wrong here...

bhirsz:
Fixes #302


robocop/run.py
```diff
@@ -45,9 +45,8 @@ def __init__(self, from_cli=False, config=None):
         self.root = os.getcwd()
         self.config = Config() if config is None else config
         self.from_cli = from_cli
-        if from_cli:
-            self.config.parse_opts()
-        else:
+        self.config.parse_opts()
```
fabioz, line 48: Note: doing this will use `sys.argv`, which is not what's wanted in the API (so, there should be a way for the API to pass the default arguments here).

bhirsz: I've actually updated it locally later -> I'm passing from_cli instead, and if from_cli is set to False it will not load from sys.argv (and this flag is True only if we're running robocop from the cli).

mnojek (approved):

PR #306, opened 2021-03-28
List configurables without severity

### LINKED ISSUE #304: Do not list severity in --list-configurables
When executing `robocop --list-configurables` the output is very long. For each rule we can change severity and this can be noted only once at the beginning of the output. The rest of the custom fields should be printed normally with related rules.

bhirsz:
Closes #304

### DIFF
--- a/robocop/run.py
+++ b/robocop/run.py
@@ -159,7 +159,7 @@ class Robocop:
         if not (self.config.list or self.config.list_configurables):
             return
         if self.config.list_configurables:
-            print("All following rules have configurable parameter severity. Allowed values are:"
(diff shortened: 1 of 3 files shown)


robocop/run.py
```diff
@@ -158,6 +158,9 @@ def load_checkers(self):
     def list_checkers(self):
         if not (self.config.list or self.config.list_configurables):
             return
+        if self.config.list_configurables:
+            print("All following rules have configurable parameter severity. Allowed values are:"
```
mnojek, line 162:
```suggestion
            print("All following rules have configurable parameter 'severity'. Allowed values are:"
```

mnojek (approved):

PR #319, opened 2021-04-02
Fix #318 TypeError at KeywordCall with variable, but without keyword

### LINKED ISSUE #318: TypeError at KeywordCall with variable, but without keyword
if you have a .robot file with this content:

```robotframework
*** Test Case ***
first
    ${a}    ${b}    # no Keyword here!!!

*** Keywords ***
get something
    [Return]    1    2
```

and run robocop with that file, an TypeError is raised:

      File "c:\develop\robot\robotframework-robocop\robocop\checkers\errors.py", line 64, in visit_KeywordCall
      match = self.setting_pattern.match(node.keyword)
    
    TypeError: expected string or bytes-like object.

If I fix this, another error occurs:

      File "c:\develop\robot\robotframework-robocop\robocop\utils\utils.py", line 50, in normalize_robot_name
        return name.replace(' ', '').replace('_', '').lower()
    AttributeError: 'NoneType' object has no attribute 'replace'

d-biehl:
This should fix #318


bhirsz:
I've added your test data from the issue to our repo - it works now after your fix. Thanks!

bhirsz (approved):

PR #321, opened 2021-04-02
Add rules for inconsistent equal sign

### LINKED ISSUE #295: [Rule] Equal signs convention should be configurable
I see some people use equal signs `=` in variable assignment, both in Variables section or in local assignments. 
I think we should consider adding a rule that can detect inconsistent use of equal signs. Right now we only detect if the equal sign appears and then report it as a redundant but not all RF users perceive them useless. Some like to use equal signs for readability.
There are two ways equal signs can be written - with `" ="` our without `"="` preceding whitespace. If we detect such inconsistency, we should report it. 
Also some people use them only in local variables assignments, some use them only in variables section and some use them in both places. We need to think if we want to separate such cases for different rules or we want to treat them globally. For people who only use them in one of these places, we would need to ignore other places or force them to use it everywhere. Hard decision how to handle that.
We can also create one configurable rule where users can decide whether they want to use them with preceding whitespace, without whitespace, without equal sign at all or maybe just ignore it and do not report any issues about that.

bhirsz:
Deprecate 0906 (redundant-equal-sign)

Closes #295


tests/atest/checkers/comments/invalid-comment/expected_output.txt
```diff
@@ -1 +0,0 @@
-${rules_dir}${\}test.robot:12:0 [E] 0703 Invalid comment. '#' needs to be first character in the cell. For block comments you can use '*** Comments ***' section
```
bhirsz, line 1: Few files were duplicated in previous PR - removing them

robocop/checkers/misc.py
```diff
@@ -134,3 +127,86 @@ def visit_KeywordCall(self, node):  # noqa
...
+        --configure inconsistent-assignment-sign-variables:assignment_sign_type:{sign_type}
+
+    ``${sign_type}` can be one of: ``autodetect`` (default), ``equal_sign`` ('='), ``none`` (''),
+    ``space_and_equal_sign`` (' =').
+
+    """
+    rules = {
+        "0909": (
+            "inconsistent-assignment-sign",
+            "The assignment sign is not consistent thorough the file. Expected '%s' but got '%s' instead",
```
mnojek, line 151:
```suggestion
            "The assignment sign is not consistent through the file. Expected '%s' but got '%s' instead",
```

robocop/checkers/misc.py
```diff
@@ -134,3 +127,86 @@ def visit_KeywordCall(self, node):  # noqa
...
+        --configure inconsistent-assignment-sign:assignment_sign_type:{sign_type}
+        --configure inconsistent-assignment-sign-variables:assignment_sign_type:{sign_type}
+
+    ``${sign_type}` can be one of: ``autodetect`` (default), ``equal_sign`` ('='), ``none`` (''),
+    ``space_and_equal_sign`` (' =').
+
+    """
+    rules = {
+        "0909": (
+            "inconsistent-assignment-sign",
```
mnojek, line 150: I think I would just name it `inconsistent-assignment`

robocop/checkers/misc.py
```diff
@@ -134,3 +127,86 @@ def visit_KeywordCall(self, node):  # noqa
...
+            "The assignment sign is not consistent thorough the file. Expected '%s' but got '%s' instead",
+            RuleSeverity.WARNING,
+            ('assignment_sign_type', 'keyword_assignment_sign_type', parse_assignment_sign_type)
+        ),
+        "0910": (
+            "inconsistent-assignment-sign-variables",
```
mnojek, line 156: I think I would just name it `inconsistent-assignment-in-variables`

robocop/run.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, from_cli=False, config=None):
         if not from_cli:
             self.config.reports.add('json_report')
         self.out = self.set_output()
+        if from_cli:
+            print("### DEPRECATION WARNING: The rule 0906 (redundant-equal-sign) is "
```
mnojek, line 53: Maybe we should suggest excluding this rule? So people would be prepared for its removal.

robocop/checkers/misc.py
bhirsz: Sweet, a lot of names will get shorter that way :D

mnojek (approved):

PR #322, opened 2021-04-03
Fix unicode keyword names throwing not capitalized warning

### LINKED ISSUE #314: Local characters cause problem with capitalization rule
**Describe the bug**
When writing a keyword where the second character is a local character robocop will complain that the capitalization rule is not followed, even though it is

**To Reproduce**
1. Create a keyword named Eäi Saa Peittää
2. Robocop will complain that it does not follow the rule
If you instead write "Ei Saa Peittää" it will be accepted, it only seems to happen when the second character is a local character

**Expected behavior**
Keyword should validate

bhirsz:
Fix #314 
Miscellaneous fixes for report printing (added missing new line)

### DIFF
--- a/tests/atest/rules/naming/not-capitalized-keyword-name/test.robot
+++ b/tests/atest/rules/naming/not-capitalized-keyword-name/test.robot
@@ -44,6 +44,7 @@ Keyword-With_Special $ Chars 10
     Edit Group's Report Logo
     Take2 Is Capitalized
     Quoted "Strings" Are Fine Too
+    Keyword_With_underscores
 
 Keyword With Library Import
     SeleniumLibrary.Input Text    ${txt_welcome}    Hello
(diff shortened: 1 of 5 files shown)


tests/atest/rules/naming/not-capitalized-keyword-name/test.robot
```diff
@@ -44,11 +44,17 @@ Keyword-With_Special $ Chars 10
...
     Input Text    ${txt_welcome}    Hello
 
 Keyword With API Abbreviation Should Pass
     Keyword With Number Ins1de 0r In Fr0nt
-    I Will Love U 4 Ever
\ No newline at end of file
+    I Will Love U 4 Ever
+
+Keyword With Unicode And Non Latin
+    Eäi Saa Peittää
```
mnojek, line 58: I see what you did here 😉

mnojek (approved):

PR #324, opened 2021-04-03
Add verbose flag

### LINKED ISSUE #72: Add verbose output
Add CLI option to turn on verbose output and also a method that write lines to output if verbose is enabled.

bhirsz:
Closes #72

### DIFF
--- a/robocop/config.py
+++ b/robocop/config.py
@@ -260,11 +260,11 @@ class Config:
         args = self.preparse(args) if from_cli else None
         if not args or args == ['--verbose']:
             loaded_args = self.load_default_config_file()
-            if loaded_args is not None:
+            if loaded_args is None:
+                self.load_pyproject_file()
+            else:
                 # thanks for this we can have config file together with some cli options like --verbose
                 args = [*args, *loaded_args] if args is not None else loaded_args
-        if not args or args == ['--verbose']:
-            self.load_pyproject_file()
         if args:
             args = self.parser.parse_args(args)
             for key, value in dict(**vars(args)).items():
--- a/robocop/config.py
+++ b/robocop/config.py
@@ -40,6 +40,8 @@ def parse_toml_to_config(toml_data, config):
             config.output = open(value, 'w')
         elif key == 'no_recursive':
             config.recursive = not value
+        elif key == 'verbose':
+            config.verbose = value
         else:
             raise InvalidArgumentError(f"Option '{key}' is not supported in pyproject.toml configuration file.")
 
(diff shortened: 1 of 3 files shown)


robocop/config.py
```diff
@@ -140,19 +144,20 @@ def __init__(self, root=None, from_cli=False):
...
-        'help_argfile':      'Path to file with arguments',
-        'help_ignore':       'Ignore file(s) and path(s) provided. Glob patterns are supported',
-        'help_info':         'Print this help message and exit',
-        'help_version':      'Display Robocop version',
+        'help_recursive':    'Use this flag to stop scanning directories recursively.',
+        'help_argfile':      'Path to file with arguments.',
+        'help_ignore':       'Ignore file(s) and path(s) provided. Glob patterns are supported.',
+        'help_info':         'Print this help message and exit.',
+        'help_version':      'Display Robocop version.',
+        'verbose':           'Display extra information.',
```
mnojek, line 160: Let's be consistent
```suggestion
        'help_verbose':           'Display extra information.',
```

robocop/config.py
```diff
@@ -245,32 +250,40 @@ def _create_parser(self):
         optional.add_argument('-h', '--help', action='help', help=self.HELP_MSGS['help_info'])
         optional.add_argument('-v', '--version', action='version', version=__version__,
                               help=self.HELP_MSGS['help_version'])
+        optional.add_argument('--verbose', action='store_true', help=self.HELP_MSGS['verbose'])
```
mnojek, line 253:
```suggestion
        optional.add_argument('--verbose', action='store_true', help=self.HELP_MSGS['help_verbose'])
```

robocop/config.py
```diff
@@ -245,32 +250,40 @@ def _create_parser(self):
...
         args = self.preparse(args) if from_cli else None
-        if not args:
-            args = self.load_default_config_file()
-        if not args:
+        if not args or args == ['--verbose']:
+            loaded_args = self.load_default_config_file()
+            if loaded_args is not None:
+                # thanks for this we can have config file together with some cli options like --verbose
+                args = [*args, *loaded_args] if args is not None else loaded_args
+        if not args or args == ['--verbose']:
```
mnojek, line 266: This is a duplicate of the condition above

bhirsz: It's not - args can be empty in the line 261 (no cli arguments, or only --verbose) but then there could be .robocop file with arguments. Depending if we were able to find .robocop file, args could be still empty or not. That's why we need to check for it again before looking for pyproject.toml

bhirsz: But when I think about it, I think load.pyproject_file should be called only if loaded_args is None - we will avoid the situation where there .robocop with only --verbose inside.

PR #325, opened 2021-04-03
Add toml docs

bhirsz:


docs/user_guide.rst
```diff
@@ -23,7 +23,39 @@ You can load arguments for Robocop from file with::
...
+    exclude = ["0203"]
+    reports = [
+        "rules_by_id",
+        "scan_timer"
+    ]
+    ignore = ["ignore_me.robot"]
+    ext_rules = ["path_to_external\\dir"]
+    filetypes = [".txt", ".csv"]
+    threshold = "E"
+    format = "{source}:{line}:{col} [{severity}] {rule_id} {desc} (name)1"
```
mnojek, line 52:
```suggestion
    format = "{source}:{line}:{col} [{severity}] {rule_id} {desc} (name)"
```

docs/user_guide.rst
```diff
@@ -23,7 +23,39 @@ You can load arguments for Robocop from file with::
...
-It will start looking from current directory and go up until it founds it or '.git' file is found.
+It will start looking from current directory and go up until it founds it or '.git' file is found. ``.robocop`` file
+supports the same syntax as given from cli::
+
+    --include rulename
+    --reports all
+
+
+If there is no ``.robocop`` file present it will try to load ``pyproject.toml`` file (if there is toml module installed).
+Robocop use [tool.robocop] section. Options have the same names as CLI arguments but hyphens '-' are replaced by underscores '_'.
```
mnojek, line 34: What exactly is replaced by underscores?

docs/user_guide.rst
```diff
@@ -23,7 +23,39 @@ You can load arguments for Robocop from file with::
...
 If no arguments are provided to Robocop it will try to find ``.robocop`` file and load it from there.
-It will start looking from current directory and go up until it founds it or '.git' file is found.
+It will start looking from current directory and go up until it founds it or '.git' file is found. ``.robocop`` file
+supports the same syntax as given from cli::
+
+    --include rulename
+    --reports all
+
+
+If there is no ``.robocop`` file present it will try to load ``pyproject.toml`` file (if there is toml module installed).
```
mnojek, line 33:
```suggestion
If there is no ``.robocop`` file, it will try to load ``pyproject.toml`` file (if there is toml module installed).
```

mnojek: BTW can we somehow force to use  toml instead of .robocop file?

docs/user_guide.rst
bhirsz: hyphens. For example if you have --ext-rules option, it became ext_rules in toml. Although I've added this line:
```
toml_data = {key.replace('-', '_'): value for key, value in toml_data.items()}
```
So this line is probably not necessary.

PR #329, opened 2021-04-05
[Rule] wrong import order

### LINKED ISSUE #313: [Rule] Order of imports similar to Python PEP8
Robocop should detect if importing libraries is done in specific order. Best will be to follow official Python PEP8:

- Standard library imports.
- Related third party imports.
- Local application/library specific imports.

bhirsz:
Closes #313 

It only check if BuiltIn imports are placed before other imports (without distinction between local and third party import for now). 

New SettingsOrder checker can be later extended with more order related rules (for example suite settings like Test Setup should be placed after imports).


robocop/checkers/misc.py
```diff
@@ -210,3 +211,38 @@ def auto_detect_assignment_sign(node):
         auto_detector = AssignmentTypeDetector()
         auto_detector.visit(node)
         return auto_detector
+
+
+class SettingsOrderChecker(VisitorChecker):
+    """ Checker for settings order.
+
+    BuiltIn library imports should be always placed first before other library imports.
```
mnojek, line 219:
```suggestion
    BuiltIn libraries imports should always be placed first before other libraries imports.
```

mnojek (approved):

PR #335, opened 2021-04-06
Big refactor

mnojek:
- Rewritten README from .rst to markdown
- Added FAQ section to README
- Prettified README
- Updated documentation
- Added shortcuts for some CI arguments
- Changed default values for quality gates (W:100 -> W:0)
- Changed defaults for LengthChecker:
  - `keyword_max_calls` 8->10
  - `keyword_min_calls` 2->1
  - `testcase_max_calls` 8->10
- Enabled severity in rules to be case-insensitive
- Return status is now calculated on difference between found issues and quality gates limits
- `--list-configurables` now displays only rules that have configurables available
- `--list-reports` gives nice output
- Changes version to 1.7.0


README.md
```diff
@@ -0,0 +1,348 @@
+[![Unit-tests](https://img.shields.io/github/workflow/status/MarketSquare/robotframework-robocop/Unit%20tests/master)](https://github.com/MarketSquare/robotframework-robocop/actions?query=workflow%3A%22Unit+tests%22 "GitHub Workflow Unit Tests Status")
+![Codecov](https://img.shields.io/codecov/c/github/MarketSquare/robotframework-robocop/master "Code coverage on master branch")
+![PyPI](https://img.shields.io/pypi/v/robotframework-robocop?label=version "PyPI package version")
+![Python versions](https://img.shields.io/pypi/pyversions/robotframework-robocop "Supported Python versions")
+![Licence](https://img.shields.io/pypi/l/robotframework-robocop "PyPI - License")
+
+---
+
+<img style="float:right" src="https://raw.githubusercontent.com/MarketSquare/robotframework-robocop/master/docs/images/robocop_logo_small.png">
```
bhirsz, line 9: the image is not loading for me:
image

README.md
```diff
@@ -0,0 +1,348 @@
...
+- [Introduction](#introduction)
+- [Documentation](#documentation)
+- [Values](#values)
+- [Requirements](#requirements)
+- [Installation](#installation)
+- [Usage](#usage)
+- [Example](#example)
+- [FAQ](#faq)
+
+<a href="http://www.youtube.com/watch?feature=player_embedded&v=VIDEO_ID"
```
bhirsz, line 23: It is placeholder?

README.md
```diff
@@ -0,0 +1,348 @@
...
+
+</details>
+
+<details>
+  <summary>Can I define custom rules?</summary>
+
+  **Yes**, you can define and include custom rules using `-rules / --ext_rules` command line option
+  by providing a path to a file containing your rule(s). The option accepts comma-separated list
+  of paths to files or directories, e.g. 
+  ```robocop -rules my/own/rule.py --ext_rules rules.py,external_rules.py```.
```
bhirsz, line 259:
```commandline
...
```
otherwise it's displayed in the same line:
image

README.md
mnojek: Yes, maybe today the video will be online already and I will fix the url.

docs/index.rst
```diff
@@ -5,10 +5,6 @@ Robocop documentation
    :maxdepth: 2
    :caption: Contents:
 
-.. include:: ../README.rst
```
bhirsz, line 8: Now we need to add some text here otherwise the entry page will be empty: 
https://robocop.readthedocs.io/en/bamf-refactor/

docs/requirements.txt
```diff
@@ -1,2 +1,2 @@
 sphinx-rtd-theme
-robotframework>=3.2.1
\ No newline at end of file
+robotframework>=3.2.2
```
bhirsz, line 2: This is requirement only for building the docs. The reason is that we're using Robocop while building the docs (for example to get configurables) and when importing it, we need to also import rf. It can be 3.2.2 or even 4.0.0 since it only affects docs building

mnojek: Should I update it to 4.0.0?

docs/index.rst
mnojek: Yeah, I know.

mnojek:
New changes:
- Changed option name `--ext_rules` to `--ext-rules`
TODO:
- Add some text to entry page of docs (since readme is not included there)
- Add utests for return status calculation
- Update URL for YouTube video of our RoboCon talk

robocop/config.py
```diff
@@ -233,32 +233,32 @@ def _create_parser(self):
...
                               help=self.HELP_MSGS['help_threshold'])
         optional.add_argument('-A', '--argumentfile', metavar='PATH', help=self.HELP_MSGS['help_argfile'])
-        optional.add_argument('--ignore', action=ParseDelimitedArgAction, default=self.ignore,
+        optional.add_argument('-g', '--ignore', action=ParseDelimitedArgAction, default=self.ignore,
                               metavar='PATH', help=self.HELP_MSGS['help_ignore'])
         optional.add_argument('-h', '--help', action='help', help=self.HELP_MSGS['help_info'])
         optional.add_argument('-v', '--version', action='version', version=__version__,
                               help=self.HELP_MSGS['help_version'])
-        optional.add_argument('--verbose', action='store_true', help=self.HELP_MSGS['help_verbose'])
+        optional.add_argument('-vv', '--verbose', action='store_true', help=self.HELP_MSGS['help_verbose'])
```
bhirsz, line 253: Usually -vv, or -vvv .. means "verbose but more" (for example while single --verbose print only few extra message, -vv prints every issue even if ignored etc). So it could be confusing for some people. Although we could stick to -vv if you want it to be consistent (short and long options) with other options

mnojek: Yeah, I had the same issue with that, but I think that our tool will never produce so much additional output so this -vv option is still a verbose output and it is different than -v which returns tool version. What do you think? I don't have a better idea.

README.md
mnojek: Changed

docs/requirements.txt
bhirsz: Yeah let's update it

robocop/config.py
bhirsz: Yes, I don't want to add more levels of verbose so I believe we can go with -vv for --verbose.

docs/index.rst
mnojek: @bhirsz  Added back in https://github.com/MarketSquare/robotframework-robocop/pull/335/commits/d7fa8735b9b0d4b923bd2a1c15f0e573f14d95fb
Can you please check?

bhirsz: The url I gave earlier rebuilds with every commit (https://robocop.readthedocs.io/en/bamf-refactor/). It looks good!

bhirsz (approved):

bhirsz (approved):

PR #343, opened 2021-04-09
fix #342 InvalidSpacingChecker does not respect Tabs and Windows/Mac line endings

### LINKED ISSUE #342: InvalidSpacingChecker does not respect Tabs and Windows/Mac line endings
Take this piece of code:

```python
from robocop.run import Robocop
from robot.api import get_model
import io

filename = "test.robot"
text = """*** Settings *** \t\r\n  \r\n  \r\n"""  # here we have Tabs and Windows line endings \r\n

with io.StringIO(text) as data:
    model = get_model(text)
    model.source = filename
    config = Config(root=".", from_cli=False)
    config.include = set(["1001", "1002", "1010"])
    robocop = Robocop(from_cli=False, config=config)

    robocop.reload_config()
    for e in robocop.run_check(model, filename, text):
        print(f"{e.severity.value}{e.rule_id}: {e.desc} at {e.line}:{e.col}")
```

I would expect an output something like:
    
    W1001: Trailing whitespace at the end of line at 1:18
    W1001: Trailing whitespace at the end of line at 2:2
    W1001: Trailing whitespace at the end of line at 3:2
    W1010: Too many blank lines at the end of file at 3:0

But i get no output

d-biehl:
this is a fix for #342 

If you use the Robocop API you can add files with Windows and Mac line endings. It does not happen if you use the command line tool, because the files where loaded with python open, wich uses universal new lines (see python doc).


bhirsz:
Looks good and thank you for also adding utests for API :)

robocop/checkers/spacing.py
```diff
@@ -42,7 +42,7 @@ def parse_file(self):
         super().parse_file()
         if self.raw_lines:
             last_line = self.raw_lines[-1]
-            if last_line == '\n':
+            if last_line in ['\n', '\r']:
```
mnojek, line 45: What about common line ending for Windows - '\r\n'?

bhirsz (approved):

robocop/checkers/spacing.py
d-biehl: I add a new commit to fix this

robocop/checkers/spacing.py
```diff
@@ -54,14 +54,14 @@ def parse_file(self):
                 if empty_lines > 1:
                     self.report("too-many-trailing-blank-lines", lineno=len(self.raw_lines), col=0)
                     return
-            if not empty_lines and not last_line.endswith('\n'):
+            if not empty_lines and not last_line.endswith(('\n','\r')):
```
mnojek, line 57: and here (`'/r/n'`)?

tests/utest/test_api.py
```diff
@@ -29,9 +29,41 @@ def test_run_check_in_memory(self):
...
+        in_memory = "*** Settings *** \r\r"
+        ast_model = get_model(in_memory)
+        issues = robocop_runner.run_check(ast_model, r'C:\directory\file.robot', in_memory)
         expected_issues = {
             'Missing documentation in suite',
             'Section is empty',
+            'Trailing whitespace at the end of line',
             'Too many blank lines at the end of file'
         }
         assert all(issue.desc in expected_issues for issue in issues)
```
bhirsz, line 69: I think we need to reverse condition - right now if the issues is empty or only contains 1 issue it will be true (as long as given issue is inside expected issues). We actually rather want to check that "all the expected issues are in issues". That's why we missed 'Too many blank lines at the end of file' but the test passed.

robocop/checkers/spacing.py
```diff
@@ -54,14 +54,14 @@ def parse_file(self):
...
                 self.report("missing-trailing-blank-line", lineno=len(self.raw_lines), col=0)
 
     def check_line(self, line, lineno):
         self.raw_lines.append(line)
 
-        stripped_line = line.rstrip('\n')
-        if stripped_line and stripped_line[-1] == ' ':
+        stripped_line = line.rstrip('\n\r')
```
mnojek, line 63: And here it's inverted

robocop/checkers/spacing.py
d-biehl: \r\n also ends with \n, so it is not needed here

robocop/checkers/spacing.py
d-biehl: rstips removes all characters that are given in the parameter, regardless of the order

tests/utest/test_api.py
bhirsz, line 69:
```suggestion
                actual_issues = {issue.desc for issue in issues}
                assert expected_issues == actual_issues
```

tests/utest/test_api.py
```diff
@@ -29,9 +29,41 @@ def test_run_check_in_memory(self):
...
+        in_memory = "*** Settings *** \r\n\r\n"
+        ast_model = get_model(in_memory)
+        issues = robocop_runner.run_check(ast_model, r'C:\directory\file.robot', in_memory)
+        expected_issues = {
+            'Missing documentation in suite',
+            'Section is empty',
+            'Trailing whitespace at the end of line',
+            'Too many blank lines at the end of file'
+        }
+        assert all(issue.desc in expected_issues for issue in issues)
```
bhirsz, line 53:
```suggestion
        actual_issues = {issue.desc for issue in issues}
        assert expected_issues == actual_issues
```

tests/utest/test_api.py
```diff
@@ -29,9 +29,41 @@ def test_run_check_in_memory(self):
         in_memory = "*** Settings ***\n\n"
         ast_model = get_model(in_memory)
         issues = robocop_runner.run_check(ast_model, r'C:\directory\file.robot', in_memory)
+        expected_issues = {
+            'Missing documentation in suite',
+            'Section is empty',            
+            'Too many blank lines at the end of file'
+        }
+        assert all(issue.desc in expected_issues for issue in issues)
```
bhirsz, line 37:
```suggestion
        actual_issues = {issue.desc for issue in issues}
        assert expected_issues == actual_issues
```

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
mnojek, line 1: Probably added by mistake

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
```
mnojek, line 1: Probably added by mistake

.vscode/settings.json
d-biehl: i reverted the commit, sorry

bhirsz (approved):

mnojek (approved):

PR #363, opened 2021-04-21
Include license in distributions

bollwyvl:
Hi, thanks for `robocop`!

This PR adds the LICENSE to the source distribution, as per the Apache 2.0.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 recursive-include robocop *
+include LICENSE


bhirsz (approved):
Good idea with adding licencje to distribution :)!

bollwyvl:
Backstory: looking to ship this with [conda-forge](https://github.com/conda-forge/staged-recipes/pull/14658) so that we can test it with [jupyterlab-lsp](https://github.com/krassowski/jupyterlab-lsp/pull/493). Thanks again!

A packaging quick note: if i'm maintaining a package, i like to run the unit tests. i'm having to patch out `test_find_project_root_missing_but_git` because (i believe) it's looking for the repo of my _packaging_ script... otherwise everything looking good!

bhirsz:
> Backstory: looking to ship this with [conda-forge](https://github.com/conda-forge/staged-recipes/pull/14658) so that we can test it with [jupyterlab-lsp](https://github.com/krassowski/jupyterlab-lsp/pull/493). Thanks again!
> 
> A packaging quick note: if i'm maintaining a package, i like to run the unit tests. i'm having to patch out `test_find_project_root_missing_but_git` because (i believe) it's looking for the repo of my _packaging_ script... otherwise everything looking good!

Right, it's not the best test out there.. I will think about it's design (it's like you mentioned - it's looking for .git file, and the next .git file is one from your project). Most likely I will create temporary directory with temporary files for the sake of test.

bhirsz:
And for sake of integrating with jupyterlab-lsp you can of course create separate issue and get help if there is need for it. I saw there is https://github.com/krassowski/jupyterlab-lsp/blob/master/scripts/lint.py but robocop is integrated with lsp already so maybe there is no need to call it from commad line separatetly like pylint or rflint tools. But you probably want to configure it a bit (ignore some rules etc).

bollwyvl:
Right. Diagnostics are a quick win for testing "is this thing on" in the
browser, and we try to bring along the linter (or cop, in this case)
supported by the language-server-under-test. And of course we'd use against
our (not exactly tiny) robot suite!

mnojek (approved):

Issue #357, opened 2021-04-15
Unify spelling of "test case" in names of rules.


MoreFamed:
Mentions of test cases in names of default rules are written with hyphen (e.g. "too-long-test-case") or without it ("missing-doc-testcase"). This is confusing, especially when Robocop does not warn about unknown names of rules. Please, unify the spelling of the phrase.

bhirsz:
Additionally from fixing the naming, "especially when Robocop does not warn about unknown names of rules" - do you want Robocop to print out the warning? It should fail when trying to configure not existing rule. But like you said it will ignore `robocop --include idontexist`. We could either print simple warning at the start of the run or throw exception (probably better).

MoreFamed:
Yes, I wish Robocop to warn about unknown names of rules. I will open another issue regarding this, so the topic of this issue could remain just unification of spelling. Thanks.

PR #364, opened 2021-04-22
not-capitalized-keyword-name rule enhancements and rule naming refactors

### LINKED ISSUE #341: Improve description of rule W0302
Now it says that 'Keyword name should be capitalized' which is misleading because we expect each word of the keyword name to be capitalized, which is sometimes related to **title case**.
There is a difference between upper case, title case and capitalization. In my opinion (and most of the examples follow it):
- **upper case** is all letters are uppercased, e.g. `THIS IS MY KEYWORD`
- **title case** is first letters of each word are uppercased, e.g. `This Is My Keyword`
- **capitalization** is only first letter is uppercased, e.g. `This is my keyword`

That being said, I think what we expect is **title case**, not **capitalization**.
We need to make the rule description more accurate.

### LINKED ISSUE #359: Loosen check of capitalization of keyword names optionally
**Problem**
Rule 0302 checks whether every word of a keyword name starts with capital letter. This is too strict for two reasons:

- For other languages than English might be natural to name things with only the first letter of their name being capital.
- Even in the (English) [documentation](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html) of RF, we can find many keywords with only the first letter of their names being capital. 

**Suggestion**
I'd like the rule 0302 to be configurable: the default value of its parametr would check for every first letter being capital, as it does up to now. When the other value of the parameter will be specified, check for first letter of the first word only would be performed.

bhirsz:
* allow to check only first word in keyword capitalization
* capitalized -> title case in W0302

Closes #359 
Closes #341


robocop/checkers/naming.py
```diff
@@ -75,8 +75,9 @@ class KeywordNamingChecker(VisitorChecker):
     rules = {
         "0302": (
             "not-capitalized-keyword-name",
```
bhirsz, line 77: 'capitalized' is still in rule name, but changing it would be not backward compatible

robocop/checkers/naming.py
```diff
@@ -75,8 +75,9 @@ class KeywordNamingChecker(VisitorChecker):
...
-            "Keyword name should be capitalized",
-            RuleSeverity.WARNING
+            "Keyword name should use title case",
+            RuleSeverity.WARNING,
+            ('check_only_first_word', 'check_only_first_word', bool)
```
mnojek, line 80: How about naming this parameter as `convention` (or whatever fits better) and allowing values like:
- `first_word_capitalized` - only first word needs to be capitalized (like in `This is my keyword`)
- `each_word_capitalized` - each word in keyword name needs to be capitalized (like in `This Is My Keyword`)
- `none` - no rules for case in keyword naming, so even names like `this_is_my_keyword` are allowed

robocop/checkers/naming.py
mnojek: We can provide new rule instead of this one and this can be tagged as deprecated and then we can use whatever is more suitable. What do you think?

bhirsz: I agree, and we can add print with deprecation warning only if someone includes / excludes deprecated rule (thanks for the PR with verification if rule exist if used in config it's now possible)

robocop/checkers/naming.py
bhirsz: I agree. Does none means all lower case? So it will report on 'First word', and 'All Words' but not on "all lowercase'. Because if it means 'don't check case' then it would be the same as disabling this rule.

mnojek: I actually meant that 'none' means that nothing is required from the keyword name but you're right that it is the same as disabling this rule. Should we have also some other values then `first_word_capitalized` and `each_word_capitalized`?

bhirsz: `first_word_capitalized`, `each_word_capitalized` sounds fine, although it could be also convention: `capitalize`, `title_case`, `lower_case`

mnojek: Can you elaborate on those modes? Do you consider them as something different or they are just different names for the ones I explained?

bhirsz: Just different names

mnojek: I'm still not sure... Mine are more descriptive, but yours are shorter 😁

bhirsz: I will go with yours, it's better to be more descriptive

robocop/checkers/naming.py
bhirsz: I've replaced the names but now I wonder if it's too explicit. 'not-title-case-keyword-name' with convention param 'first_word_capitalized' would not actually check for title case but capitalization. So maybe more general `wrong-capitalization-in-keyword-name` or similiar?

mnojek: I think the name `wrong-case-in-keyword-name` for the parameter should work fine.

bhirsz:
I've also renamed missing-doc-testcase to missing-doc-test-case which should resolve #357

robocop/checkers/naming.py
```diff
@@ -212,8 +220,8 @@ class SettingsNamingChecker(VisitorChecker):
     """ Checker for section naming violations. """
     rules = {
         "0306": (
-            "setting-name-not-capitalized",
-            "Setting name should be capitalized or upper case",
+            "setting-name-not-title-case",
```
mnojek, line 223: What about?
```suggestion
            "setting-name-not-in-title-case",
```

mnojek (approved):

PR #374, opened 2021-05-05
Do not check empty lines between test cases for templated suites

### LINKED ISSUE #367: Automatically ignore certain rules when using Templates
When a suite uses templates, the empty-lines-between-test-cases rule should be skipped (maybe others too) to allow the following structure

image

bhirsz:
Fixes #367

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -186,7 +186,7 @@ class EmptyLinesChecker(VisitorChecker):
             if not isinstance(child, node_type):
                 continue
             empty_lines = self.verify_empty_lines(child)
-            if empty_lines != allowed_empty_lines and allowed_empty_lines != -1 and index < last_index:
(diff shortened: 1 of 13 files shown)


robocop/checkers/spacing.py
```diff
@@ -186,14 +186,14 @@ def verify_empty_lines_between_nodes(self, node, node_type, issue_name, allowed_
             if not isinstance(child, node_type):
                 continue
             empty_lines = self.verify_empty_lines(child)
-            if index < last_index and empty_lines != allowed_empty_lines:
+            if empty_lines != allowed_empty_lines and allowed_empty_lines != -1 and index < last_index:
```
mnojek, line 189: I think I will split those conditions into separate lines.

bhirsz: I refactored it this way:
```
if allowed_empty_lines not in (empty_lines, -1) and index < last_index
```

mnojek (approved):

PR #377, opened 2021-05-12
Fix #376 - Rule '0801 duplicated-test-case' does not work correctly with normalized names

### LINKED ISSUE #376: Rule '0801 duplicated-test-case' does not work correctly with normalized names
given this robot file:

```robotframework
*** Test Cases ***
Test    
    No Operation
   
test
    No Operation


Test 1
    [Documentation]  doc
    [Tags]  sometag   
    No Operation
   

test1
    No Operation

Te s t_1
    No Operation

*** Keywords ***
Keyword
    [Documentation]  this is doc
    No Operation
  
```

I would expect that robocop gives error about "Multiple test cases with name *", because running the test with robot gives us a warning.

d-biehl:
this should fix #376

### DIFF
--- a/robocop/checkers/duplications.py
+++ b/robocop/checkers/duplications.py
@@ -85,7 +85,8 @@ class DuplicationsChecker(VisitorChecker):
                 self.report(rule, duplicate.name, node=duplicate)
 
     def visit_TestCase(self, node):  # noqa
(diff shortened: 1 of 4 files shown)


bhirsz (approved):
Looks good!

PR #388, opened 2021-07-10
Fix for reading format option in .robocop config file

### LINKED ISSUE #387: -f /--format argument gives FileError when put in config file
**Describe the bug**
When I create a config file with the all the configuration i want to use for our projects i also include a --format command.
When i copy and paste the format line directly in the command line it is working. But when i call the the config file with the -a argument it raises a FileError: File does not exist. After some testing it seems that it cant parse a space in the config file.
My command is as follows:  

--format "line:{line}, column:{col}, {rule_id}:{desc}[{severity}]"

This gives back the following error
Traceback (most recent call last):
  File "c:\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Python37\Scripts\robocop.exe\__main__.py", line 7, in <module>
  File "c:\python37\lib\site-packages\robocop\run.py", line 288, in run_robocop
    linter.run()
  File "c:\python37\lib\site-packages\robocop\run.py", line 78, in run
    self.recognize_file_types()
  File "c:\python37\lib\site-packages\robocop\run.py", line 97, in recognize_file_types
    for file in self.get_files(files, self.config.recursive):
  File "c:\python37\lib\site-packages\robocop\run.py", line 233, in get_files
    yield from self.get_absolute_path(Path(file), recursive)
  File "c:\python37\lib\site-packages\robocop\run.py", line 237, in get_absolute_path
    raise robocop.exceptions.FileError(path)
robocop.exceptions.FileError: File "column:{col}" does not exist

After i removed the all the spaces from the line it works. But makes it very unreadably.
Also when i Copy paste that exact --format command into the command line it works as expected.

**To Reproduce**
1. Have a Config file with a -f or --format command. 
2. Put in the string after the --format command one space between the text.
3. Run the config file via a '-A' argument via the command line.
4. it will give back a FileError

**Expected behavior**
I Expected it to behave exactly the same as i would run it from the command line

**Screenshots**
Robocop_bug

mnojek:
Fixes #387

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -179,7 +179,7 @@ Acceptance tests
 Acceptance tests are dynamically generated for every rule in Robocop. Test data
 should be located in ``tests\atest\rules\{rules_category}`` directory. If your rule has name "rule-name"
 it will expect ``rule-name`` directory with ``expected_output.txt`` file inside.
-You can put one or more *.robot files inside - it will be autoscanned with ``--include "rule-name"`` option.
+You can put one or more \*.robot files inside - it will be autoscanned with ``--include "rule-name"`` option.
 Robocop output will be compared with content of ``expected_output.txt`` file.
 
 When updating ``expected_output.txt`` file you can use two macro variables: ``${rules_dir}`` and ``${/}``.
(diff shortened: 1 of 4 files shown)


mnojek:
I still need to decide if we want to keep .robocop file and if yes, then I need to add more unit tests for this part.

bhirsz (approved):

Issue #386, opened 2021-07-07
[Bug] KeywordCall without keyword will cause fatal error


bhirsz:
This code:
```
*** Keywords ***
Keyword
    Some Keyword
${EXECDIR}\\target\\New.txt        ${file}
    Another Keyword
```
will cause fatal error -> but it should report (not existing yet) "missing keyword name" or similar issue. 
simplified example, just variable is enough:
```
*** Keywords ***
Keyword
    Some Keyword
    ${var}
    Another Keyword
```

mnojek:
@bhirsz I fixed the issue with fatal error and now Robocop can parse it. I also added new rule for missing keyword name, but I am thinking how it should be discovered. Considering following code:
```
1 *** Keywords ***
2 Keyword
3     var
4     ${var}
5     ${one}      ${two}
6     ${var}var
7     ${var} var
8     ${var}Var
9     ${var} Var
```
Robocop discovers these issues:
```
test.robot:3:0 [W] 0302 Keyword name should use title case (wrong-case-in-keyword-name)
test.robot:4:0 [E] 0403 Missing keyword name when calling some values (missing-keyword-name)
test.robot:5:0 [E] 0403 Missing keyword name when calling some values (missing-keyword-name)
test.robot:6:0 [W] 0302 Keyword name should use title case (wrong-case-in-keyword-name)
test.robot:7:0 [W] 0302 Keyword name should use title case (wrong-case-in-keyword-name)
```
While lines 4 and 5 (E0403) are obvious and caught what I expected to, the lines 3, 6 and 7 (W0302) show that keyword does not use title case. Also, lines 8 and 9 are accepted. Do you think that this is expected and we should leave it like this?

I will add a PR related to this soon...

bhirsz:
Robocop is not able to differentiate between "not enough spaces between assign and keyword" and "keyword with embedded arguments" so this output look ok for me

PR #389, opened 2021-07-10
Handle missing keyword name and add rule for it

mnojek:
Fixex #386


bhirsz (approved):
Looks good! You can also try to run robocop only with new rule over while tests directory - to see if it doesn't have any false negatives.

PR #390, opened 2021-07-10
Add favicon to docs, update html static path

### LINKED ISSUE #385: Add favicon to docs
Example robotidy docs with favicon and robocop without:
image

mnojek:
Closes #385


docs/conf.py
```diff
@@ -61,7 +61,8 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = ['images']
+html_favicon = "images/robocop.ico"
```
mnojek, line 65: @bhirsz I saw that you use `_static` as a directory name for images. Is it important to use this name or you just used what was in the [sphinx documentation](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_static_path)?

bhirsz: The name is not important, I just used default from docs

bhirsz (approved):

PR #391, opened 2021-07-10
Add rule suggesting to use resource file

### LINKED ISSUE #380: [Rule] extension could be changed to .resource
There could be an info level rule that suggests to change file extension to .resource if there are no tests define inside.

mnojek:
Closes #380


robocop/checkers/misc.py
```diff
@@ -258,3 +260,25 @@ def visit_LibraryImport(self, node):  # noqa
...
+    """ Checker for resource files. """
+    rules = {
+        "0912": (
+            "can-be-resource-file",
+            "No tests in '%s' file, consider renaming to '%s.resource'",
+            RuleSeverity.INFO
+        )
+    }
+
+    def visit_File(self, node):  # noqa
```
bhirsz, line 275: There could be problems with API - it would be good to test if it works with API (for example in intellij plugin)

robocop/checkers/misc.py
```diff
@@ -258,3 +260,25 @@ def visit_LibraryImport(self, node):  # noqa
...
+        source = node.source if node.source else self.source
+        if source:
+            extension = Path(source).suffix
+            if '.robot' in extension and not any([isinstance(section, TestCaseSection) for section in node.sections]):
+                self.report("can-be-resource-file",
+                            Path(source).name,
+                            Path(source).stem,
+                            node=node)
+        super().visit_File(node)
```
bhirsz, line 284: Do we need to visit it further?

mnojek: No, we don't. I accidentally copy-pasted.

robocop/checkers/misc.py
```diff
@@ -258,3 +260,25 @@ def visit_LibraryImport(self, node):  # noqa
...
+            "No tests in '%s' file, consider renaming to '%s.resource'",
+            RuleSeverity.INFO
+        )
+    }
+
+    def visit_File(self, node):  # noqa
+        source = node.source if node.source else self.source
+        if source:
+            extension = Path(source).suffix
+            if '.robot' in extension and not any([isinstance(section, TestCaseSection) for section in node.sections]):
```
bhirsz, line 279: What if someone is using ".txt"? It could be good to always recommend resource extension (especially if someone is not using default robot)

robocop/checkers/misc.py
mnojek: There are, but they appear only when running tests on GitHub, not locally. Do you have an idea why, @bhirsz?

bhirsz: Paths are resolved differently:
```
E         "No tests in 'file.robot' file, consider renaming to 'file.resource'"
E         Extra items in the right set:
E         "No tests in 'C:\\directory\\file.robot' file, consider renaming to 'C:\\directory\\file.resource'"
```
It most likely occurs because we're passing Windows-like path in this line: https://github.com/MarketSquare/robotframework-robocop/blob/master/tests/utest/test_api.py#L29 (it would be best to construct Path() and then convert it to str - the slashes will be converted to be compatible with running OS).

PR #392, opened 2021-07-11
[Rule] Variables in Variables section should be left aligned

### LINKED ISSUE #293: [Rule] Variables in Variables section should be left aligned

yo-ga:
- Adjust `VariableNamingChecker`
- Add `LeftAlignedChecker`
- Add Rule 1014
- Add test case

- Close #293


tests/atest/rules/spacing/variable-should-left-aligned/test.robot
```diff
@@ -0,0 +1,13 @@
+*** Settings ***
+Library   String
+Documentation   test
+
+
+*** Variables ***
```
bhirsz, line 6: Verify also with only one space - its not recognized as separator but part of the name do it could cause problems (`` ${var} value``)

bhirsz: Also add gold cases - correctly aligned to test it doesn't trigger rule

robocop/checkers/spacing.py
```diff
@@ -472,3 +472,23 @@ def get_indent(node):
...
+            RuleSeverity.ERROR
+        )
+    }
+
+    def visit_VariableSection(self, node):  # noqa
+        for child in node.body:
+            if not child.data_tokens:
+                continue
+            tokens = child.data_tokens
+            if tokens[0].type == Token.VARIABLE and tokens[0].value == '':
```
bhirsz, line 492: It looks like it will always trigger even for correctly aligned - because first token will be non data separator. It also doesnt work for RF 3.2 for some reason - check cli results in github

yo-ga:
The results of parsing are different between 3.2 and 4.0.

RF 3.2:
```py
[Token(ERROR, '', 7, 0, "Invalid variable name ''.")]
```

RF 4.0
```py
[Token(VARIABLE, '', 7, 0), Token(ARGUMENT, '${MY VAR}', 7, 2), Token(ARGUMENT, '10', 7, 13)]
```

Is there any solution to handle this situation?

robocop/checkers/duplications.py
```diff
@@ -96,6 +96,8 @@ def visit_VariableSection(self, node):  # noqa
         self.generic_visit(node)
 
     def visit_Variable(self, node):  # noqa
+        if node.name == '':
```
bhirsz, line 99: ``if not node.name`` could be better, if name is None (can't think of example when it's None but just in case)

robocop/checkers/spacing.py
```diff
@@ -472,3 +473,29 @@ def get_indent(node):
...
+        "1014": (
+            "variable-should-left-aligned",
+            "Variable in variable section should be left aligned.",
+            RuleSeverity.ERROR
+        )
+    }
+
+    def __init__(self):
+        super().__init__()
+        self.disabled = not IS_RF4
```
bhirsz, line 490: Great that you've discovered this 'feature' - I think it's not covered in any contribution guides (and I definietly need to describe it)

bhirsz, line 490: You can also try multiline variables (golden and negative). For example:
```
  ${NOT_ALIGNED}  1
...  2
${ALIGNED}  1
...  2
```

bhirsz:
Thanks for the PR - looks good and I apologize for doing review in 'turns' but sometimes I recall some extra edge case only after some time

tests/atest/rules/spacing/variable-should-left-aligned/test.robot
```diff
@@ -0,0 +1,17 @@
...
+ ${MY VAR}  10
+  ${MY VAR1}  10
+   ${MY VAR2}  10
+    ${MY VAR3}  10
+${GOLD VAR}  10
```
bhirsz, line 11: You can also try multiline variables (golden and negative). For example:
```
  ${NOT_ALIGNED}  1
...  2
${ALIGNED}  1
...  2
```

robocop/checkers/spacing.py
```diff
@@ -472,3 +473,29 @@ def get_indent(node):
...
+
+    def visit_VariableSection(self, node):  # noqa
+        for child in node.body:
+            # print(child.data_tokens)
```
bhirsz, line 494: leftover prints

robocop/checkers/spacing.py
yo-ga: In RF 3.2, the error is handled by 0401 parsing-error rule.

bhirsz (approved):

PR #393, opened 2021-07-11
[Rule] Variable assigned but not used

### LINKED ISSUE #209: Rule: variable assigned but not used
If the variable was assigned from keyword call but not used in current scope (method, test case).

yo-ga:
- Add the checker
- Add test case
- Close #209


robocop/checkers/misc.py
```diff
@@ -258,3 +258,52 @@ def visit_LibraryImport(self, node):  # noqa
...
+        self.check_unused()
+
+    def visit_TestCase(self, node):  # noqa
+        self.variable_list.clear()
+        self.generic_visit(node)
+        self.check_unused()
+
+    def use_variable(self, node):
+        node_nomalized = self.normalize(node.value)
+        for i, token in enumerate(self.variable_list):
```
bhirsz, line 296: Instead of storing multiple copies of the same variable (if it's used twice) and renornalizing whole list of variables, maybe it would be better to save it to set or dict (and already normalized)?

robocop/checkers/misc.py
```diff
@@ -258,3 +258,52 @@ def visit_LibraryImport(self, node):  # noqa
...
+            RuleSeverity.WARNING
+        )
+    }
+
+    def __init__(self):
+        self.variable_list = []
+        super().__init__()
+
+    def visit_KeywordCall(self, node):  # noqa
+        for child in node.data_tokens:
```
bhirsz, line 278: You can assign variables with the same name as your arguments:
```
${var} Keyword ${var} 
```
And variable after redeinition would be not used. Thats why you should first get only arguments (and save assign to temporary list for now), call use_variaes on it,  and then process assign.

variable_list should also pop/remove variables when using them - so you will catch redefined but not used vars

tests/atest/rules/misc/variable-not-used/test.robot
```diff
@@ -0,0 +1,30 @@
...
+Documentation    This is suite doc
+
+
+*** Test Cases ***
+Test
+    [Documentation]   doc
+    ${var}    Keyword
+
+
+*** Keywords ***
```
bhirsz, line 11: There should be also example with [Return] - since it also count as using var

bhirsz, line 11: What about tests for test cases?

bhirsz:
Misc comments: how about checking [Arguments] as well? Also variable can be passed as kwarg l:
```
Keyword param=${value} 
```

bhirsz:
Another cases I can think of:
```
Keyword  argument value with ${value} embedded
```
(arguments would need to be parsed through some kind of regex)

Class instances or dict variables :
```
${class_instance} Keyword 
$(dict} Keyword 
Another Keyword ${class_instance.attribute}
Another Keyword ${dict['key']} 
```

bhirsz:
You code matches most of the cases - I listed missing ones in previous comments.

Advice for architecture: like I mentioned in other comment you can store token in dictionary, for example:
```
{
  '${normalized_name}: Token(),
  ${other_normalized_name}: Token()
}
```
when calling use_variables, you can pop from dict (and ignore if it's not present - most likely variable is global). Remember to correctly recognize variable name - match anything before `.` or `[` for example.

When adding to dictionary (ie. on assign, [Return], [Arguments]) you can extract varable name (in case its "Some ${embedded} variable" or key=${value}), normalize it and check if it already exist in dictionary - if yes, you're trying to redefine variable that was not used so it should be reported. If it doesn't exist yet - just add it.

At the end of Keyword/Test Case if can call for check_unused like you're doing now.

robocop/checkers/misc.py
yo-ga: If I change `list` to `dict`, the rule would only catch once in the following situation. Is this expect result correct?
```robotframework
** Settings ***
Documentation    This is suite doc


*** Test Cases ***
Test
    [Documentation]   doc
    ${var}    Keyword
    ${var}    Keyword 2
```

yo-ga: If I change `list` to `dict`, the rule would only catch once in the following situation. Is this expect result correct?
```robotframework
** Settings ***
Documentation    This is suite doc


*** Test Cases ***
Test
    [Documentation]   doc
    ${var}    Keyword
    ${var}    Keyword 2
```

tests/atest/rules/misc/variable-not-used/test.robot
yo-ga: There is only one cases for `Test Cases` now.
I'll copy the cases in the keywords and adjust them for test cases.

robocop/checkers/misc.py
bhirsz: In this case it should work as dict - because you're redefining variable. 

ie:
```
    ${var}    Keyword
    ${var}    Keyword 2
    Log  ${var}
```
Your current checker will not detect that first ${var} was never used and just redifined. 

and in this case
```
    ${var}    Keyword
    ${var}    Keyword 2
```
If you're redefining not used variable - it should throw warning. Then it should throw warning on second ${var} because it's also not used anywhere.

yo-ga:
> Misc comments: how about checking [Arguments] as well? Also variable can be passed as kwarg l:
> 
> ```
> Keyword param=${value} 
> ```

Should the error message and error number be different? Such as `argument-variable-not-used`

bhirsz:
> > Misc comments: how about checking [Arguments] as well? Also variable can be passed as kwarg l:
> > ```
> > Keyword param=${value} 
> > ```
> 
> Should the error message and error number be different? Such as `argument-variable-not-used`

Yes, good idea to split it into separate rule.

bhirsz:
Another case I came up with (yeah it's one of the most trickiest rules so far): variable can be used in IF condition or FOR LOOP, not only in keyword call.

yo-ga:
There are still same cases needed to handle
- `Keyword    ${var_one + var_another}`
- `Keyword    ${var by ${place}}`

bhirsz:
Hi, what's the status, are you working on it or you're awaiting our further review? I apologise but I got lost in the middle of conversation so you could be waiting for review etc and I prefer to confirm (it's fine if you're working on it, just checking if you need any input from me).

yo-ga:
Sorry for not updating the code or status
I would close the PR until I finish adjusting the code

PR #400, opened 2021-07-28
Add rule for checking if variable is without value

### LINKED ISSUE #294: [Rule] Variable value should not be empty
Special variable ${EMPTY} can be used instead (depending if it's desired behaviour)
```
*** Variables ***
${VARIABLE}
```

mnojek:
Closes #294


tests/atest/rules/misc/empty-variable/test.robot
```diff
@@ -0,0 +1,9 @@
+*** Variables ***
```
bhirsz, line 1: add also:
```
${MULTILINE_WITH_EMPTY}
...    ${EMPTY}
${MULTILINE_NO_VALUE}
...
${MULTILINE_WITH_EMPY_LINES}   Value
...  # I think it should be just empy line, without value - but I'm not sure
...    ${EMPTY}
```

mnojek: I tested it and it is... strange. You're right that empty string appeared in the second line of `${MULTILINE_WITH_EMPY_LINES}` but the same happened to `${MULTILINE_NO_VALUE}`. So my part of the code `if not node.value` will not catch it because actually this node has `value` field with 1 value which is `''` (empty string). 
I will adjust the code to catch those cases. Tell me what you think.

robocop/checkers/misc.py
```diff
@@ -243,3 +243,20 @@ def visit_LibraryImport(self, node):  # noqa
...
+            "empty-variable",
+            "Use built-in variable ${EMPTY} instead of leaving variable without value",
+            RuleSeverity.INFO
+        )
+    }
+
+    def visit_Variable(self, node):  # noqa
+        if not node.name:
+            return
+        if not node.value or all(not val for val in node.value):
```
bhirsz, line 261: It will only raport those variables that have all values empty, ie:

```
@{ARRAY}
...
...
```
but it will not catch mixed empty:
```
@{ARRAY}
...
...  value
```
It will also not catch ``MULTILINE_WITH_EMPTY_LINES`` from your test data - it's actually array of two values, one empty and second ${EMPTY}, with comment in between them. It should be catched.

The easiest way of achieving this would be problably calling get_tokens on variable looking for Token.Argument and check them one by one (it's better to check on tokens because then you can point to exact place where there is empty value). So in this case:
```
@{ARRAY}
...  1
...  2
...  3
...  a
...  b
...  c
...
...  5
```
It will only raport on ``...  `` and not whole variable

bhirsz:
Also, what about variables declared in other places?

```
${variable}    Set Variable   1
...  
```
(I don't know if this particular syntax is possible though)

bhirsz:
See https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html and search for "Handling empty values". This rule should also report on `\ ` used instead of `${EMPTY}`

bhirsz (approved):

PR #412, opened 2021-08-13
External rules from module

### LINKED ISSUE #404: import external rules through python package
**Is your feature request related to a problem? Please describe.**
Managing external rules is tedious when the latter need to be shared across multiple projects.

**Describe the solution you'd like**
At least, I'd like to vendor my external rules as Python packages. I'd like to specify a package name instead of a path to files/directory to load external rules.

At best, I'd like Robocop to provides a plugin system similar to [flake8](https://github.com/PyCQA/flake8).

bhirsz:
Closes #404

### DIFF
--- a/docs/external_rules.rst
+++ b/docs/external_rules.rst
@@ -69,7 +69,7 @@ Configurable parameter can be referred by its :code:`public_name` in command lin
 
 Import from external module
 ----------------------------
-Robocop rules can be written in seperate, distributed module. For example using ``RobocopRules`` module name and following
+Robocop rules can be written in separate, distributed module. For example using ``RobocopRules`` module name and following
 directory structure::
 
     RobocopRules/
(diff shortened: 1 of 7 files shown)


docs/external_rules.rst
```diff
@@ -66,3 +66,47 @@ Rules can have configurable values. You need to specify them in rule body after
...
+    RobocopRules/
+    RobocopRules/__init__.py
+    RobocopRules/some_rules.py
+    setup.py
+
+inside ``__init__.py``::
+
+    from .some_rules import CustomRule
+
+    all = ['CustomRule']
```
fdaguin, line 84: To comply with the class defined in `RobocopRules/some_rules.py`:

```suggestion
    from .some_rules import MissingKeywordName

    all = ['MissingKeywordName']
```

bhirsz: Good eye! But it should be other way around - since MissingKeywordName checker already exist in default ones, I should rename class to CustomRule.

fdaguin: Good point!

PR #413, opened 2021-08-15
Allow to configure section order in section-out-of-order

### LINKED ISSUE #384: Allow to configure section-out-of-order with custom order
From #382

bhirsz:
Closes #384 
Also fix unintended feature where you could configure severity of multiple rules in one --configure statement: it was breaking parsing for any other configurable with commas. 
If there will be need for this in the future we can add special parsing of severity (splitting on ``,``) to ``configure_checkers_or_reports`` method.


robocop/checkers/duplications.py
```diff
@@ -146,18 +153,54 @@ class SectionHeadersChecker(VisitorChecker):
...
+        self.configure('sections_order', 'settings,variables,testcases,keywords')
         self.sections_by_order = []
         self.sections_by_existence = set()
         super().__init__()
 
-    def visit_File(self, node):
+    def configure(self, param, value):
+        section_map = {
+            'settings': Token.SETTING_HEADER,
+            'variables': Token.VARIABLE_HEADER,
```
mnojek, line 166: @bhirsz  Why no option with singular 'variable' and 'setting'?

bhirsz: I guess it doesn't hurt to add it if I covered other cases - I will do it in separate PR

PR #417, opened 2021-08-18
Add summary after listing rules

### LINKED ISSUE #416: Print summary with rules count after --list and --list-configurables command

bhirsz:
Closes #416

### DIFF
--- a/robocop/run.py
+++ b/robocop/run.py
@@ -194,10 +194,10 @@ class Robocop:
                     print(f"{rule_def}\n"
                           f"    {configurables}")
                     severity_counter[rule_def.severity.value] += 1
-        suffix = '.' if self.config.list else ' with configurable parameters.'
-        print(f"Altogether {sum(severity_counter.values())} rule(s): "
-              f"{severity_counter['E']} error(s), {severity_counter['W']} warning(s) and {severity_counter['I']} info"
-              f"{suffix}")
+        print(f"Altogether {sum(severity_counter.values())} rule(s) with following severity:\n"
+              f"    {severity_counter['E']} error rule(s),\n"
+              f"    {severity_counter['W']} warning rule(s),\n"
+              f"    {severity_counter['I']} info rule(s).")
         sys.exit()
 
     def load_reports(self):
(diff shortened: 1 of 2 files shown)


mnojek:
I think I will make the counter information more precise, e.g.:
```
"Altogether 3 rule(s) with following severity:
    1 Error rule(s),
    1 Warning rule(s),
    1 Info rule(s).\n"
```
Also, I think we can drop the part "with configurable parameters." since I don't see a big value in it. But you can try to convince me, because maybe I don't see it clearly 😉

bhirsz:
> I think I will make the counter information more precise, e.g.:
> 
> ```
> Altogether 3 rule(s) with following severity:
>     1 Error rule(s),
>     1 Warning rule(s),
>     1 Info rule(s).
> ```
> 
> Also, I think we can drop the part "with configurable parameters." since I don't see a big value in it. But you can try to convince me, because maybe I don't see it clearly 😉

It's bit of "just a remined those are only rules with configurable parameters, not all of them" but I see point we don't really need them - especially we're not reminding if some rules were filtered out, ie:
--list missing-doc*

mnojek:
Yes, I think we can drop it. If someone wants to list only configurable rules, I guess one remembers what was passed several seconds ago and what is expected. No need to add it, in my opinion.

robocop/run.py
```diff
@@ -194,10 +194,10 @@ def list_checkers(self):
                     print(f"{rule_def}\n"
                           f"    {configurables}")
                     severity_counter[rule_def.severity.value] += 1
-        suffix = '.' if self.config.list else ' with configurable parameters.'
-        print(f"Altogether {sum(severity_counter.values())} rule(s): "
-              f"{severity_counter['E']} error(s), {severity_counter['W']} warning(s) and {severity_counter['I']} info"
-              f"{suffix}")
+        print(f"Altogether {sum(severity_counter.values())} rule(s) with following severity:\n"
```
mnojek, line 197: Have you checked if there is one line of space between this summary and the list of rules?

bhirsz: Local version example:
image

mnojek: I mean if there is a blank line between this summary and the actual list of rules while calling `--list`?

bhirsz: Ah no, summary is directly under list of rules. If I were to add black line it would be like this:
image
And now it is:
image

Which version is better?

mnojek: I think the first one with one blank line of space for readability.

PR #418, opened 2021-08-18
Small code clean up

bhirsz:


robocop/config.py
```diff
@@ -328,30 +322,11 @@ def load_pyproject_file(self):
...
 
-    @staticmethod
-    def replace_in_set(container, old_key, new_key):
-        if old_key not in container:
-            return
-        container.remove(old_key)
-        container.add(new_key)
-
     def validate_rule_names(self, rules):
-        deprecated = {
```
mnojek, line 339: Instead of removing this dict, maybe we can leave it empty? So that we don't need to come up with the same concept again in the future?

bhirsz: Good idea, it may be usefull in the future

mnojek (approved):

PR #422, opened 2021-08-25
Detect same variables with similar names

### LINKED ISSUE #120: Rule: Detect same variables
RF ignores case, spaces and underscore in variable names and it's sometimes confused that
```
${some var}
${someVar}
${some_var}
${SOME_VAR}
```
are treated like the same variable. This can lead to bugs when someone overwrites the variable using very similar but different name.
It should be reported by robocop to avoid potential bugs.

mnojek:
Closes #120


mnojek:
@bhirsz I added support for detecting similar variables in arguments.
Also now it checks also different types of variables like scalar, list and dict.
Have a look.

robocop/checkers/naming.py
```diff
@@ -378,3 +376,39 @@ def visit_KeywordCall(self, node):  # noqa
...
+    def visit_TestCase(self, node):  # noqa
+        self.check_similar_variables(node)
+
+    def check_similar_variables(self, node):
+        """
+        Creates a dictionary `variables` with normalized variable name as a key
+        and ads a list of all detected variations of this variable in the node as a value,
+        then it checks if similar variable was found.
+        """
+        variables = defaultdict(list)
```
bhirsz, line 403: why not set instead of list?

tests/atest/rules/naming/possible-variable-overwriting/test.robot
```diff
@@ -0,0 +1,27 @@
...
+    ${MYVAR}        Set Variable    5
+    ${M Y V A R}    Set Variable    6
+    ${mYvAr}        Some Keyword    123
+
+
+*** Keywords ***
+Keyword With Overwritten Variables
+    [Arguments]     ${my_var}
+    ${someVar}      Set Variable    1
+    ${SOME_VAR}     Set Variable    2
```
bhirsz, line 17: Add more testing with:
- for loops (especially if nothing breaks and nested for loops are visited)
- if conditions
- Templated tests (the ones with [Template])

tests/atest/rules/naming/possible-variable-overwriting/expected_output.txt
```diff
@@ -0,0 +1,12 @@
+${rules_dir}${\}test.robot:5:4 [I] 0316 Variable '${MY VAR}' may overwrite similar variable inside 'Test With Overwritten Variables' TestCase
```
bhirsz, line 1: Do you think it will be clear what's this issue about? Should we either a) mention for which other var it's similar or b) mention how Robot works (ie normalize Robot name)

mnojek: I was thinking about both but:
a) providing variable can be tricky, because sometimes it can be similar to several variables and
b) this makes the message very long but I don't think it's a big issue, so I can add it.

robocop/checkers/naming.py
```diff
@@ -378,3 +380,70 @@ def visit_KeywordCall(self, node):  # noqa
...
+                tokens = child.get_tokens(Token.ASSIGN)
+                self.find_similar_variables(tokens, variables, node)
+
+            # check FOR LOOPS
+            elif isinstance(child, For):
+                # look for nested for loops
+                while child.body and isinstance(child.body[0], For):
+                    for var in child.variables:
+                        variables[normalize_robot_var_name(var)].add(var)
+                    child = child.body[0]
```
bhirsz, line 429: What if its nested for loop and then two adjacent for loops? I think it would be better to create variable default dict globally, reset it before entering keyword / test case and then gather and check variables using visitors. It will take care of nested structures for you.

robocop/checkers/naming.py
```diff
@@ -378,3 +380,70 @@ def visit_KeywordCall(self, node):  # noqa
...
+                for token in child.body:
+                    tokens = token.get_tokens(Token.ASSIGN)
+                    self.find_similar_variables(tokens, variables, node)
+
+            # check IF CONDITIONS
+            elif isinstance(child, If):
```
bhirsz, line 435: Will it work with else if?

mnojek: Probably no, good point.

robocop/checkers/naming.py
mnojek: Yeah, you're right. I tried using visitors but I again failed with nested structures so I did this. Now I see, that I should stick with visitors. I'm tired of this issue already 😛

bhirsz (approved):

PR #447, opened 2021-08-30
Improve test coverage for comment tests, add support for missing cases

### LINKED ISSUE #378: More variants of 'todo' in comment should be accepted
Consider following code snippet:
```
*** Settings ***
Documentation  Suite doc
# TODO found


*** Keywords ***  # TODO not found
Different Place Todos    # TODO found
    [Documentation]  Some doc  # TODO not found
    No Operation
    [Return]        somevalue       # TODO not found
# TODO found

Different Case Todos
    [Documentation]  Some doc
    No Operation    # TODO found
    No Operation    # todo found
    No Operation    # TODO: found
    No Operation    # todo: found
    No Operation    # to-do not found
    No Operation    # TO-DO not found
    No Operation    # to do not found
    No Operation    # TO DO not found

```
I think we need to decide where we should also detect 'todo' in comment and also what variants of it we want to accept. What do you think, @bhirsz ?

mnojek:
Closes #378


tests/atest/rules/comments/missing-space-after-comment/expected_output.txt
```diff
@@ -1 +1,26 @@
...
+${rules_dir}${\}test.robot:27:0 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:29:18 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:30:9 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:31:34 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:32:25 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:33:19 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:34:18 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:38:16 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:40:29 [W] 0702 Missing blank space after comment character
+${rules_dir}${\}test.robot:42:10 [W] 0702 Missing blank space after comment character
```
mnojek, line 25: I don't know why it discovers everything from RF4 execpt for `test.robot:45:30 [W] 0702 Missing blank space after comment character` which is:
```
    FOR  ${var}  IN  @{list}  #invalid comment
```

robocop/checkers/comments.py
```diff
@@ -50,6 +50,93 @@ def visit_KeywordCall(self, node):  # noqa
...
+
+    def visit_ForceTags(self, node):  # noqa
+        self.find_comments(node)
+        self.generic_visit(node)
+
+    def visit_DefaultTags(self, node):  # noqa
+        self.find_comments(node)
+        self.generic_visit(node)
+
+    def visit_IfHeader(self, node):  # noqa
```
bhirsz, line 101: You don't need to check for IS_RF4 if whole node was introduced in 4.0+ - visitor will just not match this node if it doesn't exist

tests/atest/rules/comments/missing-space-after-comment/expected_output.txt
bhirsz: Isn't it because you're limiting this check to IS_RF4?
```
    def visit_ForHeader(self, node):  # noqa
        if IS_RF4:
            self.find_comments(node)
            self.generic_visit(node)

```

bhirsz: Ok I know why. It's called `ForLoopHeader` in 3.2 (and that's what you should visit for RF 3.2). So you need both visit_ForHeader and visit_ForLoopHeader

robocop/checkers/comments.py
```diff
@@ -50,6 +50,93 @@ def visit_KeywordCall(self, node):  # noqa
         self.find_comments(node)
         self.generic_visit(node)
 
+    def visit_Return(self, node):  # noqa
```
bhirsz, line 53: I've checked your code and I've noticed that we actually don't need visit_Keyword and visit_TestCase visitors. I've removed them and it worked fine. Can you also verify it?

robocop/checkers/comments.py
mnojek: Yeah, I thought the same but I was trying to understand why it visits some of the nodes even if they haven't existed in RF 3.2. For example lines: test.robot:38:16, test.robot:40:29, test.robot:42:10.

bhirsz: It will visit those nodes but it will not recognize it as for loop - just as keyword call (with weird indentation ;) )

robocop/checkers/comments.py
mnojek: I needed to add it back because it couldn't discover one of the cases. Now it works fine.

tests/atest/rules/comments/missing-space-after-comment/expected_output.txt
mnojek: Added missing visitor and also removed condition for checking RF version. Now it works fine.

bhirsz (approved):

PR #451, opened 2021-08-31
Remove variants of robot vars from keyword name before check

### LINKED ISSUE #433: Robot Framework 4.1 support
- do not warn on `robot:continue-on-failure` and `robot:recursive-continue-on-failure ` new tags
- variable can be also `${var}[key] ` (for naming rules that remove variables before check)

bhirsz:
It will deal with this syntax:
```
${var}['key']['key2']
```
But it will also fix minor issues with embedded variables (for example it was not working correctly for ${var.attr}). 

I've decided to replace regex pattern with python method for replacing robot variables - it was more clear than monstrous regex that would be required to cover nested robot variables and dictionary subscriptions.

Closes #433

### DIFF
--- a/robocop/utils/misc.py
+++ b/robocop/utils/misc.py
@@ -248,7 +248,8 @@ def remove_robot_vars(name):
     replaced = ''
     index = 0
     while index < len(name):
-        if not started and name[index] in var_start and index + 1 < len(name) and name[index+1] == '{':
+        if not started and name[index] in var_start and index + 1 < len(name) and name[index+1] == '{' and not \
+                (index and name[index-1] == '\\'):
             started = True
             open_bracket = '{'
             close_bracket = '}'
--- a/robocop/utils/misc.py
+++ b/robocop/utils/misc.py
@@ -248,6 +248,7 @@ def remove_robot_vars(name):
     replaced = ''
     index = 0
     while index < len(name):
+        # it looks for $ (or other var starter) and then check if next char is { and previous is not escape \
         if not started and name[index] in var_start and index + 1 < len(name) and name[index+1] == '{' and not \
                 (index and name[index-1] == '\\'):
             started = True
--- a/robocop/utils/misc.py
+++ b/robocop/utils/misc.py
@@ -240,39 +240,36 @@ def last_non_empty_line(node):
     return node.lineno
 
 
+def next_char_is(string, i, char):
+    if not i < len(string) - 1:
+        return False
+    return string[i + 1] == char
+
+
 def remove_robot_vars(name):
     var_start = set('$@%&')
-    started, after_var = False, False
     brackets = 0
     open_bracket, close_bracket = '', ''
     replaced = ''
     index = 0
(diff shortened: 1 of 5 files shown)


robocop/utils/misc.py
```diff
@@ -234,3 +234,42 @@ def last_non_empty_line(node):
...
+
+def remove_robot_vars(name):
+    var_start = set('$@%&')
+    started, after_var = False, False
+    brackets = 0
+    open_bracket, close_bracket = '', ''
+    replaced = ''
+    index = 0
+    while index < len(name):
+        # it looks for $ (or other var starter) and then check if next char is { and previous is not escape \
```
mnojek, line 247:
```suggestion
        # it looks for $ (or other var starter) and then checks if next char is { and previous is not escape \
```

robocop/utils/misc.py
```diff
@@ -234,3 +234,42 @@ def last_non_empty_line(node):
...
+        if not started and name[index] in var_start and index + 1 < len(name) and name[index+1] == '{' and not \
+                (index and name[index-1] == '\\'):
```
mnojek, line 249: I think I will split these conditions into separate lines for readability.

robocop/utils/misc.py
```diff
@@ -234,3 +234,42 @@ def last_non_empty_line(node):
         if not isinstance(child, EmptyLine):
             return child.lineno
     return node.lineno
+
+
+def remove_robot_vars(name):
```
mnojek, line 239: Wow, we have some nice IFology here 😄

bhirsz: I didn't spent last week on learning algorithms and compelx data structures for no reason!

bhirsz: Refactored a bit

robocop/utils/misc.py
```diff
@@ -234,3 +234,39 @@ def last_non_empty_line(node):
...
+            if name[index] == open_bracket:
+                brackets += 1
+            elif name[index] == close_bracket:
+                brackets -= 1
+            if not brackets:
+                # check if next chars are not ['key']
+                if next_char_is(name, index, '['):
+                    brackets += 1
+                    index += 1
+                    open_bracket, close_bracket = '[', ']'
```
mnojek, line 262:
```suggestion
            # check if next chars are not ['key']
            if not brackets and if next_char_is(name, index, '['):
                brackets += 1
                index += 1
                open_bracket, close_bracket = '[', ']'
```

robocop/utils/misc.py
```diff
@@ -234,3 +234,39 @@ def last_non_empty_line(node):
...
+        # it looks for $ (or other var starter) and then checks if next char is { and previous is not escape \
+        elif name[index] in var_start and next_char_is(name, index, '{') and not (index and name[index - 1] == '\\'):
```
mnojek, line 264: Please split it into separate lines for readability.

mnojek (approved):

PR #452, opened 2021-08-31
Rule detecting hyphens inside variable names

### LINKED ISSUE #271: Rule: Catch hyphens usage inside variable names
Using hyphens inside variable names can be sometimes misunderstood by the interpreter as minus sign and break the execution. Better to use underscores instead. Would be nice to have a rule that suggests it.

mnojek:
Closes #271


robocop/checkers/naming.py
```diff
@@ -370,6 +375,10 @@ def visit_VariableSection(self, node):  # noqa
                             col=token.col_offset)
 
     def visit_KeywordCall(self, node):  # noqa
+        for token in node.get_tokens(Token.ASSIGN):
```
bhirsz, line 378: What get_tokens will return if there are not assign tokens? I think it will be None and it will break code

tests/atest/rules/naming/hyphen-in-variable-name/test.robot
```diff
@@ -0,0 +1,16 @@
...
+    ${a-b}  Set Variable  1       # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+    ${-}    Set Variable  1       # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+    ${a-}   Set Variable  1       # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+    ${-b}   Set Variable  1       # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+
+
+*** Keywords ***
+My Keyword
+    ${r}    Set Variable  ${2-1}  # this is fine
+    ${a-b}  Set Variable  1       # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
```
bhirsz, line 13: What about ${a\-b}? (escaped char)

robocop/checkers/naming.py
mnojek: It will be an empty list and it will just not iterate the for loop.

tests/atest/rules/naming/hyphen-in-variable-name/test.robot
mnojek: I think it should also be reported because RF doesn't see that it is escaped (escaping `-` is not listed anywhere in RF documentation).

tests/atest/rules/naming/hyphen-in-variable-name/test.robot
```diff
@@ -0,0 +1,19 @@
...
+    ${a-}   Set Variable  1     # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+    ${-b}   Set Variable  1     # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+    My Keyword    1  2  3
+
+
+*** Keywords ***
+My Keyword
+    ${r}    Set Variable  ${2-1}  # this is fine
+    ${a-b}  Set Variable  1     # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
+    ${a\-b}  Set Variable  1    # this will warn - because if it's later used as ${a-b} it can lead to ambiguous results
```
bhirsz, line 16: The point was that we probably should not var on hypen if its properly escaped. What do you think? We could go with it now but then update in next release

mnojek: I would keep it as it is because this is how RF understands it.

bhirsz (approved):

PR #459, opened 2021-09-02
Fix invalid-char-in-name rule not supporting embedded variables with regexes

### LINKED ISSUE #457: [Bug] Some characters considered invalid in embedded arguments
### What happened?

There are still some issues regarding handling of embedded arguments. Some characters are denoted as invalid by W0301. The behaviour is very strange and depends on some things I consider insignificant. E.g. I have exactly this file:
```
*** Keywords ***
Perform A ${arg:\d+} ${arg2:\d+}
    Log To Console  ${arg}. ${arg2}

Perform ${arg:.*}
    Log To Console  ${arg}

```
Then in line 2, there both '$'s are denoted as invalid, and '{' in line 5 too. However, those keywords are written according to documentation and work perfectly when used in tests.

Besides, characters denoted as invalid depend on length of the keyword . E.g. when changing 'Perform A' to 'Perform ABCD', the first '+' is invalid instead of both '$'s. And when changing (the second) 'Perform' to 'Perform XY', '}' is considered invalid instead of '{'.

My configuration of the rule is `-c invalid-char-in-name:invalid_chars:!.?/;+'"()[]{}#$%^&=<>|\`

### Operating System

Windows

### Robocop version

1.10.0

bhirsz:
And also fix what col is reported.

Fixes #457

### DIFF
--- a/robocop/checkers/naming.py
+++ b/robocop/checkers/naming.py
@@ -56,14 +56,9 @@ class InvalidCharactersInNameChecker(VisitorChecker):
         index = 0
         while index < len(node.name):
             # skip variables
-            i = 0
-            while variables and i < len(variables):
-                start, stop = variables[i]
-                if start <= index < stop:
-                    index += stop - start
-                    variables.pop(i)
-                    break
-                i += 1
+            if variables and variables[0][0] == index:
+                start, stop = variables.pop(0)
+                index += stop - start
             else:
                 if node.name[index] in self.invalid_chars:
                     self.report("invalid-char-in-name", node.name[index], name_of_node, node=node,
--- a/robocop/checkers/naming.py
+++ b/robocop/checkers/naming.py
@@ -10,7 +10,14 @@ from robot.parsing.model.statements import KeywordCall, Arguments
 
 from robocop.checkers import VisitorChecker
 from robocop.rules import RuleSeverity
-from robocop.utils import normalize_robot_name, normalize_robot_var_name, IS_RF4, keyword_col, remove_robot_vars
+from robocop.utils import (
+    normalize_robot_name,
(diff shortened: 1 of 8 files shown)


robocop/checkers/naming.py
```diff
@@ -45,43 +46,35 @@ def visit_File(self, node):
...
+            # skip variables
+            i = 0
+            while variables and i < len(variables):
+                start, stop = variables[i]
+                if start <= index < stop:
+                    index += stop - start
+                    variables.pop(i)
+                    break
+                i += 1
+            else:
```
bhirsz, line 67: Since variables (start, stop) list will be always ascending I'm also thinking about this approach:
```
if variables:
    start, stop = variables[0]
              if start <= index < stop:
                    index += stop - start
                    variables.pop(0)
     else:

```

mnojek (dismissed):

robocop/checkers/naming.py
```diff
@@ -56,14 +56,9 @@ def check_if_char_in_node_name(self, node, name_of_node, is_keyword=False):
...
             # skip variables
-            i = 0
-            while variables and i < len(variables):
-                start, stop = variables[i]
-                if start <= index < stop:
-                    index += stop - start
-                    variables.pop(i)
-                    break
-                i += 1
+            if variables and variables[0][0] == index:
```
mnojek, line 59: Let's assume that I believe you that this works correctly :)

mnojek (approved):

robocop/checkers/naming.py
bhirsz: Thats not the matter of belief if you have good enough tests!  But yeah lets hope it will fix this issue for good.

PR #469, opened 2021-09-04
Add rule for BDD keywords

### LINKED ISSUE #159: Support BDD tests

bhirsz:
Closes #159


robocop/utils/misc.py
```diff
@@ -18,7 +18,7 @@
 from robocop.rules import RuleSeverity
 from robocop.exceptions import InvalidExternalCheckerError
 
-IS_RF4 = VERSION.startswith('4')
+IS_RF4 = VERSION.startswith('4')  # FIXME: We need better version matching - for 5.0.0
 DISABLED_IN_4 = frozenset(('nested-for-loop', 'invalid-comment'))
 ENABLED_IN_4 = frozenset(('if-can-be-used', 'else-not-upper-case', 'variable-should-left-aligned'))
```
mnojek, line 23: BTW Not related but I think we need to change this rule name from `variable-should-left-aligned` to `variable-should-be-left-aligned`

robocop/checkers/naming.py
```diff
@@ -110,6 +111,11 @@ class KeywordNamingChecker(VisitorChecker):
             "keyword-name-is-empty",
             "Keyword name should not be empty",
             RuleSeverity.ERROR
+        ),
+        "0318": (
+            "bdd-without-keyword-call",
+            "BDD reserved keyword '%s' used without keyword%s",
```
mnojek, line 117: What about:
```suggestion
            "BDD reserved keyword '%s' not followed by any keyword%s",
```
?

robocop/checkers/naming.py
```diff
@@ -203,6 +216,14 @@ def check_keyword_naming(self, keyword_name, node):  # noqa
         if any(word[0].islower() for word in words if word):
             self.report("wrong-case-in-keyword-name", node=node)
 
+    def check_bdd_keywords(self, keyword_name, node):
+        if keyword_name.lower() not in self.bdd:
+            return
+        arg = node.get_token(Token.ARGUMENT)
+        suffix = f". Concatenate it with keyword: '{keyword_name.title()} {arg.value}'" if arg else ""
```
mnojek, line 223: ...and then here:
```suggestion
        suffix = f". Use one space between: '{keyword_name.title()} {arg.value}'" if arg else ""
```

mnojek: Just a suggestion but still I am thinking about something better and easier to understand.

robocop/utils/misc.py
bhirsz: Right, it's not backward compatible change though - I will create issue for it.

robocop/checkers/naming.py
bhirsz: Yeah it's easier to understand, I will go with it

robocop/checkers/naming.py
```diff
@@ -129,6 +135,12 @@ class KeywordNamingChecker(VisitorChecker):
         'else',
         'else if'
     }
+    bdd = {
+        'given',
+        'when',
+        'and',
+        'then'
```
mnojek, line 142: One more keyword is also possible: `but`. See [here](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#behavior-driven-style)

bhirsz: Oh good eye, I will update it

mnojek (approved):

Issue #467, opened 2021-09-03
Benchmark rules


bhirsz:
We don't analyse performance of robocop in any structured way. It would be possible to add benchmark (ie for each rule) runnable using tox. 

https://github.com/ionelmc/pytest-benchmark

It would be best to design it so we could run it in previous Robocop versions for comparison.

PR #472, opened 2021-09-04
Add ignore_pattern param to line-too-long rule 

### LINKED ISSUE #348: Allow to ignore patterns in line length checks (ie. urls)
Most linters ignore specific patterns - for example urls like http://... - in too long line checks. We should also implemented it in Robocop.

bhirsz:
Which allows to ignore lines that contains given pattern.

Closes #348 

And it should be checked with benchmakers in the future #467 (to see how pattern checking is impacting performance).


robocop/checkers/lengths.py
```diff
@@ -189,6 +189,12 @@ class LineLengthChecker(RawFileChecker):
                 "max_line_length",
                 int,
                 'number of characters allowed in one line'
+            ),
+            (
+                "ignore_pattern",
+                "ignore_pattern",
+                pattern_type,
+                "Ignore lines that contains configured pattern"
```
mnojek, line 197:
```suggestion
                "Ignore lines that contain configured pattern"
```

robocop/checkers/lengths.py
```diff
@@ -197,9 +203,12 @@ def __init__(self):
         self.max_line_length = 120
         # replace # noqa or # robocop, # robocop: enable, # robocop: disable=optional,rule,names
         self.disabler_pattern = re.compile(r'(# )+(noqa|robocop: ?(?P<disabler>disable|enable)=?(?P<rules>[\w\-,]*))')
+        self.ignore_pattern = re.compile(r'https?://\S+')
```
mnojek, line 206: Shouldn't those backslashes be escaped like `r'https?:\/\/\S+'`?

bhirsz: No, thanks to r'

mnojek (approved):

PR #479, opened 2021-09-05
Improve handling arguments with variable imports

### LINKED ISSUE #434: Recognize variables passed to variable import in duplicated-variables-import rule
Variables import can have variables. Our duplicated-variables-import rule when storing names for checks does not load these. It should work like library import - create unqiue name from import name and variables.

### LINKED ISSUE #435: Verify how variable imports behave with optional arguments
Variables import can have variables:
```
*** Settings ***
Variables  vars.py  arg1  2

```
We should check what happens if we configure robot or yaml type variables with arguments and optionally add rule reporting it.

### LINKED ISSUE #482: Do not report duplicated content on the first occurence
In case of our duplication rules we're reporting issues on every duplicated line (including original). Modify it so it doesn't report on original. Mention original line in the message:
```
test.robot:13:1 [E] 0801 Multiple test cases with name "Test" (first occurrence in line 6)
```

bhirsz:
Closes #434 - now arguments for python variable files are parsed and it's possible to reuse the same variables import but with different arguments
Closes #435 - added variables-import-with-args rule that checks if we provide args to robot or yaml variable file

Closes #482

### DIFF
--- a/robocop/checkers/errors.py
+++ b/robocop/checkers/errors.py
@@ -3,6 +3,8 @@ Errors checkers
 """
 import re
 
+from robot.api import Token
+
 from robocop.checkers import VisitorChecker
 from robocop.rules import RuleSeverity
 from robocop.utils import IS_RF4
@@ -90,3 +92,18 @@ class MissingKeywordName(VisitorChecker):
                 lineno=node.lineno,
                 col=node.data_tokens[0].col_offset + 1
             )
+
+
+class VariablesImportErrorChecker(VisitorChecker):
+    """ Checker for syntax error in variables import. """
+    rules = {
+        "0404": (
+            "variables-import-with-args",
+            "Robot and YAML variable files do not take arguments",
+            RuleSeverity.ERROR
+        )
+    }
+
+    def visit_VariablesImport(self, node): # noqa
+        if node.name and not node.name.endswith('.py') and node.get_token(Token.ARGUMENT):
+            self.report("variables-import-with-args", node=node)
--- a/tests/atest/rules/duplications/duplicated-variables-import/expected_output.txt
+++ b/tests/atest/rules/duplications/duplicated-variables-import/expected_output.txt
@@ -1,8 +1,5 @@
-${rules_dir}${\}test.robot:3:1 [W] 0807 Duplicated variables import with path "variables.py" in suite
-${rules_dir}${\}test.robot:5:1 [W] 0807 Duplicated variables import with path "variables.py" in suite
-${rules_dir}${\}test.robot:9:1 [W] 0807 Duplicated variables import with path "other1.py" in suite
-${rules_dir}${\}test.robot:10:1 [W] 0807 Duplicated variables import with path "other1.py" in suite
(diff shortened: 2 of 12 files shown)


robocop/checkers/errors.py
```diff
@@ -90,3 +92,18 @@ def visit_KeywordCall(self, node):  # noqa
                 lineno=node.lineno,
                 col=node.data_tokens[0].col_offset + 1
             )
+
+
+class VariablesImportErrorChecker(VisitorChecker):
+    """ Checker for syntax error in variables import. """
+    rules = {
+        "0404": (
+            "variables-import-with-args",
```
mnojek, line 101: I am thinking if we should use shortcuts in rule names. What is better - `"variables-import-with-args"` or `"variables-import-with-arguments"`?

tests/atest/rules/duplications/duplicated-variables-import/expected_output.txt
```diff
@@ -1,2 +1,8 @@
 ${rules_dir}${\}test.robot:3:1 [W] 0807 Duplicated variables import with path "variables.py" in suite
-${rules_dir}${\}test.robot:5:1 [W] 0807 Duplicated variables import with path "variables.py" in suite
\ No newline at end of file
+${rules_dir}${\}test.robot:5:1 [W] 0807 Duplicated variables import with path "variables.py" in suite
+${rules_dir}${\}test.robot:9:1 [W] 0807 Duplicated variables import with path "other1.py" in suite
+${rules_dir}${\}test.robot:10:1 [W] 0807 Duplicated variables import with path "other1.py" in suite
+${rules_dir}${\}test.robot:12:1 [W] 0807 Duplicated variables import with path "vars2.yaml" in suite
+${rules_dir}${\}test.robot:13:1 [W] 0807 Duplicated variables import with path "vars2.yaml" in suite
```
mnojek, line 6: I wouldn't like to have this message duplicated for each line. If we have 2 exactly the same lines, only the latter should be reported.

bhirsz: But what if you have it thrice? I know its repeating but there is benefit of easily seeing where the duplicate are. Unless we want to modify issue message to something along "variable import with path x already in x line"

mnojek: I think that reporting each duplicate twice (or more) will lead to unnecessarily big amount of artificial issues. The issue here is only one - duplication, not two. I will opt for having one reported message with an indicator of where the original one is, as you proposed.

robocop/checkers/errors.py
bhirsz: We can go with longer one - I've used shortened form for people who actually type the names instead of ids (--include ...)

mnojek: I think you're right. For name it can be shorter, because people need to actually write it. On the other hand, I think this is another thing we did not standardize and probably we have rules with 'args' and 'arguments', 'vars' and 'variables'...

bhirsz: Right, I will save if for review later

mnojek (approved):

mnojek (approved):

PR #481, opened 2021-09-05
Add misaligned-continuation-row rule

### LINKED ISSUE #438: [Rule] Continuation line over/under indented argument
Currently we're checking whether statement is aligned with other statements:
```
   Keyword
   Keyword
   ...  arg
   ...  arg2
   Also Correct One
```
And if continuation mark are aligned (they doesn't need to stay in the same line with other statements):
```
   Keyword
   Keyword
         ...  arg
           ...  arg2  # this one is not aligned with ...  arg
   Also Correct One
```
We should add check for first argument indent, reporting on:
```
   Keyword
   Keyword
   ...  arg    whatever_else
   ...    arg2  # ... are aligned, but second arg is not in line with first arg
   Also Correct One
```

bhirsz:
Closes #438 

@mnojek I have serious problem with naming today, please check my names/desc of the rule


tests/atest/rules/spacing/misaligned-continuation-row/test.robot
```diff
@@ -0,0 +1,58 @@
...
+*** Variable ***
+${STRING}          This is a long string.
+...                It has multiple sentences.
+...                It does not have newlines.
+${MULTILINE}       SEPARATOR=\n
+...                This is a long multiline string.
+...               This is the second line.
+...                This is the third and the last line.
+@{LIST}            this     list     is      quite    long     and
+...                 items in it can also be long
```
mnojek, line 22: I think this one should also be recognized as a violation of the rule.

tests/atest/rules/spacing/misaligned-continuation-row/test.robot
```diff
@@ -0,0 +1,58 @@
...
+    ...       tags    in    real    life
+    Do X    first argument    second argument    third argument
+    ...     fourth argument    fifth argument    sixth argument
+    ${var} =    Get X
+    ...    first argument passed to this keyword is pretty long
+    ...    second argument passed to this keyword is long too
+
+Test
+    [Tags]    you    probably    do    not    have    this    many
+    ...      tags    in    real    life
```
mnojek, line 38: What about this one?

robocop/checkers/spacing.py
```diff
@@ -441,31 +441,48 @@ class MisalignedContinuation(VisitorChecker, ModelVisitor):
             "misaligned-continuation",
             "Continuation marker should be aligned with starting row",
             RuleSeverity.WARNING
+        ),
+        "1015": (
+            "misaligned-continuation-row",  # FIXME better name
+            "First value in continuation row is not aligned with value from first row",
```
mnojek, line 447: What about:
```suggestion
            "Each next continuation line should be aligned with the previous one",
```

robocop/checkers/spacing.py
```diff
@@ -441,31 +441,48 @@ class MisalignedContinuation(VisitorChecker, ModelVisitor):
             "misaligned-continuation",
             "Continuation marker should be aligned with starting row",
             RuleSeverity.WARNING
+        ),
+        "1015": (
+            "misaligned-continuation-row",  # FIXME better name
```
mnojek, line 446: That looks fine, I think. I don't have a better idea.

mnojek (changes requested):

tests/atest/rules/spacing/misaligned-continuation-row/test.robot
bhirsz: I have problem with this one. I'm thinking how we should process it. In this case it's simple:
```
My Keyword    ${arg}
...  1
...  2
...  3
```
I definietly want to have ... 1, 2, 3 aligned. But what about ${arg}? if I would check for alignment with first line (like in linked example with LIST) it would look like:
```
My Keyword    ${arg}
...           1
...           2
...           3
```
So in this case I think it's better to not align with first line. So how about ignoring first line for keywords etc but checking first line for variables only?:
```
@{LIST}            this     list     is      quite    long     and
...                items in it can also be long
```

bhirsz: Same for [Tags], [Arguments] etc - they should be aligned with first row

mnojek: > So how about ignoring first line for keywords etc but checking first line for variables only?

I agree.

robocop/checkers/spacing.py
```diff
@@ -493,6 +495,21 @@ def get_indent(node):
             indent_len += len(token.value.expandtabs(4))
         return indent_len
 
+    @staticmethod
+    def token_pos(tokens, from_tok, search_for):
```
mnojek, line 499: Please add some docstring explaining what this method does

mnojek (approved):

PR #483, opened 2021-09-08
Map parsing-error to revelant (new) rules

### LINKED ISSUE #439: Format parsing-error
Parsing-error rule catch any parsing errors from Robot Framework and pass it as Robocop one. If the error message contains new lines it will also pass it:
image

Check whether it would be good to strip empty lines from parsing-error output so it doesn't break any tool parsing our log output.

### LINKED ISSUE #445: [Rule] Detect invalid FOR loop header
For example:
```
FOR	${INDEX}	In RANGE	2	17
```
It will throw `has no IN or any valid separator` during execution but we could possibly detect it with Robocop (and we could check if RF have already some validations used in runtime we could copy over to our checks). 

Another possible error:
```
    FOR  val  IN RANGE  10
```
throws `Invalid FOR loop variable 'val'.` in runtime.

### LINKED ISSUE #460: [Bug] Not enough whitespace after ... mark is recognized as new variable
This code:
```
${VARIABLE}=
... """                                            ${\n}
... data:                             ${\n}
... ( a b c ) = ( 1  2  3 ) ${\n}
... """
```
produces:
```
test.robot:23:0 [E] 0803 Multiple variables with name "... """" in Variables section. Note that Robot Framework is case-insensitive (duplicated-variable)
test.robot:26:0 [E] 0803 Multiple variables with name "... """" in Variables section. Note that Robot Framework is case-insensitive (duplicated-variable)
```
I think it should rather be catched by "not-enough-whitespace-after-newline-marker" rule (but it's keyword naming rule.. maybe we should export it to be general rule checking all statements that starts from the beginning of line and contains ...?)

### LINKED ISSUE #461: [Bug] Not all parsing errors are reported
Some parsing errors are not reporting - investigate why. 

From #460

### LINKED ISSUE #462: [Rule] *** Settings *** should be left aligned
I'm not sure if we have anything that covers that but this code:
```
*** Settings ***
   Library  Collections
```
produces
```
0401 Robot Framework syntax error: Non-existing setting ''. (parsing-error)
```
but it should report it as not aligned statement.

### LINKED ISSUE #463: [Rule] Not enough whitespace after setting
It could be detected:
```
0401 Robot Framework syntax error: Non-existing setting 'Library StatusLib'. (parsing-error)
```
as
"Not enough whitespace after Library setting" or similar. Same for other settings (if not-existing setting error is catched, we could try to guess possible errors).

### LINKED ISSUE #464: [Bug] Not correctly recognized variable with not enough whitespace
```
*** Variables ***
${var} 1
```
(1 space after ${var}). Produces:
```
0309 Section variable name should be uppercase (section-variable-not-uppercase)
0401 Robot Framework syntax error: Invalid variable name '${var} 2'. (parsing-error)
0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash (empty-variable)
```
It should not warn on empty-variable, and parsing-error should be probably catched and issued as "not enough whitespace after variable name" or similar rule.

### LINKED ISSUE #478: Rename variable-should-left-aligned to variable-should-be-left-aligned
Use methods for matching both previous and new name (with deprecating message) from config.py

bhirsz:
The PR is quite big but everything was interconnected so I had to implement this in one go.

Closes #439 Fixes #461

New rules:
  - invalid-continuation-mark (for example `..  value` or `....  value`)
  - invalid-argument (`[Arguments]  1`)
  - non-existing-setting (All not recognized tokens from Settings section)
  - setting-not-supported (discovers wrong usage of settings - for example [Arguments] in Test Case or [Template] in keyword)
  - not-enough-whitespace-after-variable (`${variable} 1`) Closes #464
  - not-enough-whitespace-after-suite-setting (`Library BuiltIn`) Closes #463
  - invalid-for-loop Closes #445
  - invalid-if
  - suite-setting-should-be-left-aligned Closes #462

Not compatible changes:
- moved `not-enough-whitespace-after-newline-marker` to errors (and id changed from "0304" to "0406". Thanks for extra error parsing we're discovering all instances of this issue now  Closes #460
- rename `missing-whitespace-after-setting` to not-enough-whitespace-after-setting`
- rename `variable-should-left-aligned` to `variable-should-be-left-aligned` Closes #478 

Removed false positives (for example "Non-existing setting ''" when setting was not left aligned).


mnojek:
Will check it today's evening 😲

robocop/checkers/errors.py
```diff
@@ -7,47 +7,207 @@
 
 from robocop.checkers import VisitorChecker
 from robocop.rules import RuleSeverity
-from robocop.utils import IS_RF4
+from robocop.utils import IS_RF4, find_robot_vars
 
 
 class ParsingErrorChecker(VisitorChecker):
-    """ Checker that returns Robot Framework DataErrors as lint errors. """
+    """ Checker that parse Robot Framework DataErrors. """
```
mnojek, line 14:
```suggestion
    """ Checker that parses Robot Framework DataErrors. """
```

mnojek (commented):
Actually this is the only thing I found :D Nice job!

mnojek (approved):

PR #485, opened 2021-09-09
Black formatting

mnojek:
- Added badge to README with download statistics and updated installation command
- Added 'black' to dev package in setup.py
- Updated RF version for tox (BTW, it fails after your last changes inside disablers, can you check it locally on this branch?)
- Added pyproject.toml file for black configuration
- Fixed some issues reported by pylama
- All other changes done automatically by black

BTW if we have black now, should we get rid of other linters or not? 😛  I see that they do not overlap completely.


robocop/reports.py
```diff
@@ -20,7 +20,8 @@
 class Report:
     def configure(self, name, value):
         raise robocop.exceptions.ConfigGeneralError(
-            f"Provided param '{name}' for report '{self.name}' does not exist")  # noqa
+            f"Provided param '{name}' for report '{self.name}' does not exist"
```
mnojek, line 23: `self.name` is not a member of this class

robocop/checkers/comments.py
```diff
@@ -16,24 +10,21 @@
...
-    """ Checker for comments content. It detects invalid comments or leftovers like `todo` or `fixme` in the code. """
+    """Checker for comments content. It detects invalid comments or leftovers like `todo` or `fixme` in the code."""
+
     rules = {
-        "0701": (
-            "todo-in-comment",
-            "Found %s in comment",
-            RuleSeverity.WARNING
-        ),
+        "0701": ("todo-in-comment", "Found %s in comment", RuleSeverity.WARNING),
```
bhirsz, line 16: This will make rules definitions a bit inconsistent but I think we can accept it since it will be replaced by class in the future anyway.

bhirsz:
A lot of '' -> "" ;) But it looks nice - finally a bit more consistent code base.

> * Updated RF version for tox (BTW, it fails after your last changes inside disablers, can you check it locally on this branch?)
It also fails for me. It's not because of the latest changes - it fails in build dependencies in tox stage. Investigating

> BTW if we have black now, should we get rid of other linters or not? 😛 I see that they do not overlap completely.

I think we don't need to remove them, we can keep them for now since not everything is catched by black.

bhirsz:
Found this in `setup.py`:

```
from robocop.version import __version__
```

Since importing from robocop.version requires to import from robocop (which in turns import few of our packages that depends on Robot Framework) the installation will fail if you don't have Robot Framework installed.

mnojek:
> Found this in `setup.py`:
> 
> ```
> from robocop.version import __version__
> ```
> 
> Since importing from robocop.version requires to import from robocop (which in turns import few of our packages that depends on Robot Framework) the installation will fail if you don't have Robot Framework installed.
> 
> Simplest solution - which would be just to define version in two places (version.py and setup.py). Otherwise we would need to either a) remove importing some of the packages when importing from robocop.version -> hard to do since some could use robocop as python module b) see if it's possible to remove Robot Framework dependency from disablers (also difficult), c) read version from file by parsing the file itself (easy).

Feel free to do any changes in this branch if you have time. I will come back to it later.

bhirsz (approved):

PR #486, opened 2021-09-10
Rewrite the invalid-char-in-name rule and update tests

### LINKED ISSUE #466: Let the rule 'invalid-char-in-name' define allowed characters
When writing keywords in one particular project, one usually wants to use only one natural language for their names. (At least Indo-European) languages are written using just a few dozens characters each. One usually wants to allow keyword names to contain digits and several non-alphanumerical characters (space, underscore, hyphen....) as well -- and that's all. In other words, from thousands of possible characters, only several dozens are usually allowed to be used.

This is why I suggest adding one more option to the rule 'invalid-char-in-name'. Then a set of **allowed** characters would be listed in the rule. For brevity, I suggest implementing notation known from regexps where possible values of a character (given in brackets) could be given just as the first and the last item of a lists separated by hyphen. Example for Polish: :D one could probably have something like this line in their _.robocop_ file:   
`-c invalid-char-in-name:allowed_chars:A-Za-z0-9ĄąĆćĘęŁłŃńÓóŚśŹźŻż _-`

I have no idea how should be solved when defines both invalid and allowed chars for the rule. (Probably the same way how the situation is solved now when someone sets the same option of a rule twice with different values.)

mnojek:
Closes #466


robocop/checkers/naming.py
```diff
@@ -26,20 +27,20 @@ class InvalidCharactersInNameChecker(VisitorChecker):
...
+                pattern_type,
+                "pattern defining characters (not) allowed in a name",
             ),
         )
     }
 
     def __init__(self):
-        self.invalid_chars = {".", "?"}
+        self.pattern = re.compile(r"[\.\?]")
         super().__init__()
```
bhirsz, line 44: We will need to extensively cover this change in release notes

bhirsz (approved):

robocop/checkers/naming.py
mnojek: I already added it. Take a look :)

PR #493, opened 2021-09-15
Fix encoding issues in config files

### LINKED ISSUE #492: [Bug] Rule 'not-allowed-char-in-name' handles non-ASCII chars (?) badly
### What happened?

When I have line 
`-c not-allowed-char-in-name:pattern:[Á]`,
in my config file, Robocop crashes with trackback ending with
```
File "C:\Users\cen89549\AppData\Local\Programs\Python\Python37\lib\encodings\cp1250.py", line 23, in decode
return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 147: character maps to <undefined>
```
It looks like Robocop treats the config file as if it were in CP1250 encoding; however, _file_ utility says it is in UTF-8. This might be source of the error.

### Operating System

Windows

### Robocop version

1.11.0

bhirsz:
Fixes #492


robocop/config.py
```diff
@@ -179,9 +180,9 @@ def preparse(self, args):
 
     def load_args_from_file(self, argfile):
         try:
-            with open(argfile) as arg_f:
+            with FileReader(argfile) as arg_f:
                 args = []
-                for line in arg_f:
+                for line in arg_f.readlines():
```
bhirsz, line 185: It is not generator so I can only use methods provided in FileReader class - either read() or readlines()

mnojek (approved):

mnojek (approved):

PR #508, opened 2021-09-20
Add empty-tags rule

### LINKED ISSUE #440: [Rule] Empty [Tags]
We do not cover empty [Tags]:
```
*** Test Cases ***
Test
    [Tags]
    Log  1

```

If there is empty [Tags] and:
  - there is no Default Tags that are being overwritten
  - there is Default Tags but all test cases have [Tags] (rendering Default Tags useless)
report an issue. 

There could be issue if ``__init__.robot`` contain Default Tags and we're overwriting it in child robot suite.

bhirsz:
Closes #440 

The rule itself is simple - the difficult part is deciding on best warning message ;)

### DIFF
--- a/robocop/checkers/tags.py
+++ b/robocop/checkers/tags.py
@@ -100,6 +100,11 @@ class TagScopeChecker(VisitorChecker):
             "Tags defined in Default Tags are always overwritten",
             RuleSeverity.INFO,
         ),
+        "0608": (
+            "empty-tags",
+            "Empty [Tags]. If you are overwriting Default Tags setting use more explicit NONE",
+            RuleSeverity.WARNING
+        )
     }
 
     def __init__(self):
@@ -151,6 +156,8 @@ class TagScopeChecker(VisitorChecker):
         self.default_tags_node = node
 
     def visit_Tags(self, node):  # noqa
+        if not node.values:
+            self.report("empty-tags", node=node, col=node.end_col_offset)
         self.tags.append([tag.value for tag in node.data_tokens[1:]])
         for tag in node.data_tokens[1:]:
             if tag.value in self.force_tags:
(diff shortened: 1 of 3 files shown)


robocop/checkers/tags.py
```diff
@@ -100,6 +100,11 @@ class TagScopeChecker(VisitorChecker):
             "Tags defined in Default Tags are always overwritten",
             RuleSeverity.INFO,
         ),
+        "0608": (
+            "empty-tags",
+            "Empty [Tags]. If you are overwriting Default Tags setting use more explicit NONE",
```
mnojek, line 105:
```suggestion
            "[Tags] setting without values. Consider using NONE if you want to overwrite the Default Tags",
```

mnojek: Maybe this will be better?

tests/atest/rules/tags/empty-tags/default_and_empty_tags.robot
```diff
@@ -0,0 +1,24 @@
...
+    Keyword
+    One More
+
+Test With Tags
+    [Tags]    dummy
+    Log    2
+
+
+*** Keywords ***
+Keyword
```
mnojek, line 19: Extend tests for keywords: http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#user-keyword-tags

mnojek (changes requested):

tests/atest/rules/tags/empty-tags/default_and_empty_tags.robot
bhirsz: Sure. But " but possible Force Tags and Default Tags setting do not affect them" so I will probably put "Consider using NONE.." part only for tags from test cases.

mnojek (approved):

PR #512, opened 2021-10-18
Load configuration files by default

bithium:
This patch changes the configuration parsing logic to always load
the configuration present in the configuration files first.

Any arguments passed in the CLI append to this default configuration.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -44,7 +44,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install robotframework${{ matrix.rf-version }}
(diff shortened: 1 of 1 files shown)


bhirsz:
Github Actions fails but it's bug in our CI/CD - I've noticed that coverage tool uses `tomli` instead of `toml`. Can you try to udpate dependencies here:
https://github.com/MarketSquare/robotframework-robocop/blob/8493f0f0effb8007dd3c053466c82365f4439bcc/.github/workflows/unit-tests.yml#L47
To:
```
pip install toml pytest pytest-benchmark coverage[tomli] pyyaml
```
? I will consider switching `toml` to `tomli` in Robocop as well - looks like it offers very good perfomance.

PR #532, opened 2021-11-02
Display supported version if rule is disabled in particular RF version

### LINKED ISSUE #507: Tag rules for another RF version in --list
Currently, `--list` option shows all possible rules whether they are compatible with current local RF version or not. Would be nice to have some tag that will show which rules will be turned off by default due to different RF version.
We can do it either next to these rules, like this (assume that we have RF 3.2.*):
```
(...)
Rule - 0907 [E]: nested-for-loop: Nested for loops are not supported. You can use keyword with for loop instead (enabled for RF 4.*)
Rule - 0908 [I]: if-can-be-used: '%s' can be replaced with IF block since Robot Framework 4.0 (enabled for RF 4.*)
(...)
```
or just list them at the bottom of the list:
```
(...)

Rules disabled due to different Robot Framework version:
Rule - 0907 [E]: nested-for-loop: Nested for loops are not supported. You can use keyword with for loop instead (enabled for RF 4.*)
Rule - 0908 [I]: if-can-be-used: '%s' can be replaced with IF block since Robot Framework 4.0 (enabled for RF 4.*)
```

**Side note**: We need to add it also to the docs.

bhirsz:
Closes #507 

Currently the logic is:
- If the rule is enabled, always show `(enabled)` next to rule name (when using `--list`)
- If the rule is disabled but not because of the RF version - show `(disabled)`
- If the rule is disabled because it's not supported in used RF version - show `(disabled - enabled for RF version x)` (x = <4.0 etc)

### DIFF
--- a/robocop/rules.py
+++ b/robocop/rules.py
@@ -190,8 +190,14 @@ class Rule:
     def __str__(self):
         return (
             f"Rule - {self.rule_id} [{self.config['severity'].value}]: {self.name}: {self.message_for_docs} "
-            f'({"enabled" if self.enabled else "disabled"})'
+            f'({self.get_enabled_status_desc()})'
         )
+
+    def get_enabled_status_desc(self):
+        s = "enabled" if self.enabled else "disabled"
+        if not self.enabled and self.supported_version != "All":
+            s += f" - enabled for RF version {self.supported_version}"
+        return s
 
     def configure(self, param, value):
         if param not in self.config:
--- a/robocop/rules.py
+++ b/robocop/rules.py
@@ -196,7 +196,7 @@ class Rule:
     def get_enabled_status_desc(self):
         s = "enabled" if self.enabled else "disabled"
         if not self.enabled and self.supported_version != "All":
-            s += f" - enabled for RF version {self.supported_version}"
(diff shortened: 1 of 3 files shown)


robocop/rules.py
```diff
@@ -190,9 +190,15 @@ def supported_in_rf_version(version: str) -> bool:
...
         return (
             f"Rule - {self.rule_id} [{self.config['severity'].value}]: {self.name}: {self.message_for_docs} "
-            f'({"enabled" if self.enabled else "disabled"})'
+            f'({self.get_enabled_status_desc()})'
         )
 
+    def get_enabled_status_desc(self):
+        s = "enabled" if self.enabled else "disabled"
+        if not self.enabled and self.supported_version != "All":
+            s += f" - enabled for RF version {self.supported_version}"
```
mnojek, line 199: It looks strange and may be confusing that 'disabled' and 'enabled' words are next to each other.
I suggest changing it to:
```suggestion
            s += f" - supported only for RF version {self.supported_version}"
```

mnojek (changes requested):

mnojek (approved):

PR #535, opened 2021-11-04
Use jinja templating in rule messages

### LINKED ISSUE #534: Named placeholders inside rule messages
Currently, we have a %-based string formatting when it comes to rule messages, e.g.
`"Section name should be in format '%s' or '%s'`
which is not very readable in the code. Of course, it is then rendered with some values when the rule is violated, but before then it's sometimes hard to guess what it is.

If we create some custom string formatting, we can have something like:
`"Section name should be in format '{capitalized}' or '{uppercase}'`

Maybe we can use jinja templating for this? And maybe we can even make it so this var is used when reporting rule, e.g. `self.report("name", max_lines=30)`?

bhirsz:
Closes #534 

Accidentally while implementing it it became a lot more powerful than before: you can use full jinja2 syntax (if/for blocks, filtering etc) inside rule messages. But most important benefit for us is improved readiblity over raw '%s' formatters or newly introduced `docs_args` which were not straightforward to use.

When displayed in docs/`--list` output it will look like this:
image

### DIFF
--- a/docs/external_rules.rst
+++ b/docs/external_rules.rst
@@ -87,9 +87,9 @@ When defining rule messages you can use `jinja` templates. The most basic usage 
         )
     }
 
-Variables needs to be passed as kwargs to report() method::
+Variables need to be passed by name to report() method::
 
-    self.report("my-rule", variable="some string", "number"=10, node=node)
+    self.report("my-rule", variable="some string", number=10, node=node)
 
 
 Import from external module
--- a/docs/external_rules.rst
+++ b/docs/external_rules.rst
@@ -35,10 +35,12 @@ This is an example of the file with custom checker that asserts that no test hav
             if 'Dummy' in node.name:
                 self.report("dummy-in-name", node=node, col=node.name.find('Dummy'))
 
+Rule parameters
+---------------
 Rules can have configurable values. You need to specify them using RuleParam class and pass it as not named arg to Rule::
 
     from robocop.checkers import VisitorChecker
-    from robocop.rules import Rule, RuleParam
+    from robocop.rules import Rule, RuleParam, RuleSeverity
 
 
     rules = {
@@ -46,8 +48,8 @@ Rules can have configurable values. You need to specify them using RuleParam cla
             RuleParam(name="param_name", converter=str, default="Dummy", desc="Optional desc"),
             rule_id="9999",
             name="dummy-in-name",
-            msg="There is '%s' in test case name",
-            severity="W"
+            msg="There is '{{ variable }}' in test case name",
+            severity=RuleSeverity.WARNING,
         )
     }
 
@@ -59,7 +61,7 @@ Rules can have configurable values. You need to specify them using RuleParam cla
             if self.private_name in node.name:
                 self.report(
                     "dummy-in-name",
-                    self.param("dummy-in-name", "param_name"),
+                    variable=self.param("dummy-in-name", "param_name"),
                     node=node,
                     col=node.name.find(self.param("dummy-in-name", "param_name")))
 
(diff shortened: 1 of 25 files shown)


tox.ini
```diff
@@ -2,7 +2,7 @@
 envlist = rf3, rf4
 skip_missing_interpreters = true
 [pytest]
-addopts = --benchmark-skip
+addopts = --benchmark-skip --lf
```
bhirsz, line 5: I believe it's fine to leave this option. Those option are also read automatically by pytest if you just run `pytest tests` in cli. `--lf` option run last failed tests -> so if you have 4 failing tests out of 400, it will run only 4. If everything passed last run, it will run all tests.

docs/external_rules.rst
```diff
@@ -71,6 +73,25 @@ Value of the configurable parameter can be retrieved using :code:`param` method:
...
+        "9001": Rule(
+            rule_id="9001",
+            name="my-rule",
+            msg="You can supply variables like {{ variable }} or {{ number }}. "
+                "Basic {% if number==10 %}jinja {% endif %}syntax supported",
+            severity=RuleSeverity.ERROR
+        )
+    }
+
+Variables needs to be passed as kwargs to report() method::
```
mnojek, line 90:
```suggestion
Variables need to be passed by name to report() method::
```

docs/external_rules.rst
```diff
@@ -71,6 +73,25 @@ Value of the configurable parameter can be retrieved using :code:`param` method:
...
+
+    self.report("my-rule", variable="some string", "number"=10, node=node)
```
mnojek, line 92:
```suggestion
    self.report("my-rule", variable="some string", number=10, node=node)
```

robocop/checkers/lengths.py
```diff
@@ -15,57 +15,53 @@
...
         RuleParam(name="max_args", default=5, converter=int, desc="number of lines allowed in a file"),
         rule_id="0507",
         name="too-many-arguments",
-        msg="Keyword has too many arguments (%d/%d)",
+        msg="Keyword has too many arguments ({{ arguments_count }}/{{ max_allowed_count }})",
         severity=RuleSeverity.WARNING,
     ),
     "0508": Rule(
-        RuleParam(name="line_length", default=120, converter=int, desc="number of lines allowed in a file"),
+        RuleParam(name="line_length", default=120, converter=int, desc="number of lines allowed in a file"),  # FIXME
```
mnojek, line 64:
```suggestion
        RuleParam(name="line_length", default=120, converter=int, desc="number of lines allowed in a file"),
```

robocop/checkers/tags.py
```diff
@@ -27,7 +27,8 @@
     "0605": Rule(
         rule_id="0605",
         name="could-be-forced-tags",
-        msg='All tests in suite share those tags: "%s". You can define them in Force Tags in suite settings instead',
+        msg='All tests in suite share those tags: "{{ tags }}". '
```
mnojek, line 30:
```suggestion
        msg='All tests in suite share these tags: "{{ tags }}". '
```

robocop/checkers/tags.py
```diff
@@ -27,7 +27,8 @@
...
+        "You can define them in Force Tags in suite settings instead",
```
mnojek, line 31:
```suggestion
        "You can define them in 'Force Tags' in suite settings instead",
```

mnojek (changes requested):

robocop/checkers/lengths.py
bhirsz: FIXME is because the message content is wrong - it's not "number of lines allowed in a file" but "number of characters allowed in line" (but I didn't have any inspiration for properly naming it so I tagged it with FIXME)

mnojek: "number of characters allowed in line" looks fine 😃

bhirsz: See, it took me 10 mins yesterday and I couldn't figure it what I did here in the mid of conversation ;)

docs/external_rules.rst
```diff
@@ -87,9 +87,9 @@ When defining rule messages you can use `jinja` templates. The most basic usage
         )
     }
 
-Variables needs to be passed as kwargs to report() method::
+Variables need to be passed by name to report() method::
```
mnojek, line 90: I think that looks more understandable:
```suggestion
Variables need to be passed to report() method by their name::
```

mnojek (approved):

PR #540, opened 2021-11-05
Ignore some directories like .git by default

### LINKED ISSUE #475: Ignore selected directories by default
Some directories like `.git` could be omitted from patch searching by default (configurable through options).

bhirsz:
Closes #475 

Depending on the size of the repo (the bigger, the less benefit) or age (the older, the bigger benefit) it can yield between 1-20% time saved on robocop execution.

Most of the patterns from default exclude (`(\.direnv|\.eggs|\.git|\.hg|\.nox|\.tox|\.venv|venv|\.svn)`) should be already ignored in gitignore file. But it does extend this list and ignores `.git` which brings the biggest execution time reduction.


robocop/config.py
```diff
@@ -130,6 +138,9 @@ def __init__(self, root=None, from_cli: bool = False):
         "help_recursive": "Use this flag to stop scanning directories recursively.",
         "help_argfile": "Path to file with arguments.",
         "help_ignore": "Ignore file(s) and path(s) provided. Glob patterns are supported.",
+        "help_ignore_default": f"Paths ignored by default. "
+        f"A regular expression to exclude directories on file searches.\n"
```
mnojek, line 142:
```suggestion
        f"A regular expression to exclude directories on file search.\n"
```

robocop/config.py
```diff
@@ -130,6 +138,9 @@ def __init__(self, root=None, from_cli: bool = False):
...
+        f"An empty value means no paths are excluded. Default: {DEFAULT_EXCLUDES}",
```
mnojek, line 143:
```suggestion
        f"An empty value means no path is excluded. Default: {DEFAULT_EXCLUDES}",
```

mnojek (approved):

PR #542, opened 2021-11-05
Add duplicated-tags rule

### LINKED ISSUE #291: Duplicated tag

bhirsz:
Closes #291


tests/atest/rules/tags/duplicated-tags/expected_output.txt
```diff
@@ -0,0 +1,6 @@
+${rules_dir}${\}test.robot:3:30 [W] 0609 Multiple tags with name 'Tag' (first occurrence at position 3:15)
```
bhirsz, line 1: I'm not sure about "first occurrence at position..". Should we list it, or rather list different kind of info? The message should be easy to understand and useful for users.

mnojek: I think I would change it to "first occurence at line {line} column {column}"

mnojek (approved):

PR #545, opened 2021-11-07
Remove supported version matrix from tests - read from rule class instead

### LINKED ISSUE #544: Support matrix in tests is not needed
Support matrix: https://github.com/MarketSquare/robotframework-robocop/blob/81a3bb7d9ac6680dfab80ffb2908923082c3f1a5/tests/atest/test_rule.py#L11 can be replaced by checking `version` attribute of rule class directly.

bhirsz:
Closes #544

### DIFF
--- a/robocop/checkers/errors.py
+++ b/robocop/checkers/errors.py
@@ -47,7 +47,9 @@ rules = {
         msg="Provide at least two spaces after '...' marker",
         severity=RuleSeverity.ERROR,
     ),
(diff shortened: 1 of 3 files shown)


mnojek (approved):

bhirsz:
Thanks to this change (and failing tests) I have discovered that some rules were enabled for particular RF versions in tests but not in our source code - ie rule who shouldn't work on RF3.2 was still being executed on RF3.2. One more reason to keep this data in one place - to avoid human errors.

mnojek (approved):

PR #551, opened 2021-12-09
Add support for Robot Framework 5.0

bhirsz:
Fixed various tests to get it working with RF 5.0.dev1. This could change in meantime but I prefered to always have set of working tests.


robocop/checkers/naming.py
```diff
@@ -215,6 +217,7 @@ class KeywordNamingChecker(VisitorChecker):
 
     def __init__(self):
         self.letter_pattern = re.compile(r"\W|_", re.UNICODE)
+        self.in_if_block = False
```
mnojek, line 220: Do you mean 'inside IF block'? If yes, then I would recommend renaming it to `inside_if_block`.

robocop/checkers/naming.py
```diff
@@ -246,14 +249,14 @@ def visit_Keyword(self, node):  # noqa
         self.generic_visit(node)
 
     def visit_KeywordCall(self, node):  # noqa
+        if self.in_if_block and node.keyword and node.keyword.lower() in self.else_if:
```
mnojek, line 252: I know it was created earlier but maybe we can also rename this `self.else_if`? What about `else_statements` or `else_variants`? `else_if` is not explanatory.

mnojek (changes requested):

robocop/checkers/naming.py
bhirsz: I will go with `else_statements`

mnojek (approved):

Issue #557, opened 2022-02-03
Add rules for BREAK, CONTINUE


bhirsz:
Handle BREAK and CONTINUE. Possible rules (update old or create new):

1. Do not name keyword after reserved words (#570)
2. Do not use BREAK, CONTINUE outside FOR, WHILE or TRY EXCEPT (check how it works - possibly prohibit in finally) (#562)
3. Recommend to use them inside IF statement (otherwise the code after them will be dead)
4. Label `Exit For Loop`, `Exit For Loop If`, `Continue For Loop`, `Continue For Loop If` as deprecated and suggest using BREAK or CONTINUE. (#576)

mnojek:
Closing, because the point 3. was moved to #810.

PR #562, opened 2022-02-18
Add statement-outside-loop rule

### LINKED ISSUE #477: For loop keywords used outside for loop
Keywords like Continue For Loop should not be used outside for loop.

bhirsz:
Closes #477 

I've decided to only support 5.0 because in 5.0 it's error. Hovewer in <=4.0 using loop keywords (like Exit For Loop) is possible so it would need to be warning instead. It's not possible to have two different severities depending on version hence this decision.

Partly implements #557

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -252,7 +252,7 @@ rules = {
         severity=RuleSeverity.ERROR,
         version=">=5.0",
         docs="""
-        Following loop keywords and statements should only be used inside loop (``WHILE`` or ``FOR``):
+        Following keywords and statements should only be used inside loop (``WHILE`` or ``FOR``):
             - ``Exit For Loop``,
             - ``Exit For Loop If``,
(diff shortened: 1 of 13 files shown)


robocop/checkers/misc.py
```diff
@@ -243,6 +245,23 @@
...
         """,
     ),
+    "0915": Rule(
+        rule_id="0915",
+        name="statement-outside-loop",
+        msg="{{ name }} {{ statement_type }} used outside a loop",
+        severity=RuleSeverity.ERROR,
+        version=">=5.0",
+        docs="""
+        Following loop keywords and statements should only be used inside loop (``WHILE`` or ``FOR``):
```
mnojek, line 255:
```suggestion
        Following keywords and statements should only be used inside loop (``WHILE`` or ``FOR``):
```

mnojek (approved):

mnojek (approved):

PR #564, opened 2022-02-26
Add DeprecatedStatementChecker

### LINKED ISSUE #560: Run Keyword Unless is deprecated
https://github.com/robotframework/robotframework/issues/4174

MalikMlitat:
Closes #560
Add `DeprecatedStatementChecker` and set `Run Keyword If` and `Run Keyword Unless` to deprecated

set  severity of `if-can-be-used  `to WARNING


MalikMlitat:
unclean  commits

robocop/checkers/naming.py
```diff
@@ -662,3 +669,59 @@ def find_similar_variables(self, tokens, node):
...
+    def visit_KeywordCall(self, node):  # noqa
+        self.check_if_keyword_is_deprecated(node.keyword, node)
+
+    def check_if_keyword_is_deprecated(self, keyword_name, node):
+        deprecated_statements = self.deprecated_statements_rf3
+        if ROBOT_VERSION.major == 4:
+            deprecated_statements = self.deprecated_statements_rf4
+        elif ROBOT_VERSION.major == 5:
+            deprecated_statements = self.deprecated_statements_rf5
+        if deprecated_statements.get(keyword_name.lower(), None):
```
bhirsz, line 717: Keywords in RF are not only case insensitive but also can ignore spaces, underscores and can contain prefix with library name. Following are equal:
```
Run Keyword Unless
runkeywordunless
run_keyword_unless
BuiltIn.Run Keyword Unless
builtin.run_KeywoRD_UNLESS
```
But fortunately robocop have helper methods for that, you can use ``normalize_robot_name`` with ``remove_prefix=builtin.`` same as [here](https://github.com/MarketSquare/robotframework-robocop/blob/6c98345f3cb87ca916a00c191667a55a82e2d959/robocop/checkers/naming.py#L378)

bhirsz:
I see that you closed PR temporarily so I will do full review later on but it looks good now!

robocop/checkers/misc.py
```diff
@@ -89,8 +89,8 @@
         rule_id="0908",
         name="if-can-be-used",
         msg="'{{ run_keyword }}' can be replaced with IF block since Robot Framework 4.0",
-        severity=RuleSeverity.INFO,
-        version=">=4.0",
+        severity=RuleSeverity.WARNING,
```
bhirsz, line 92: You can leave it with INFO (since we decided to deprecate this rule for versions above 4).

robocop/checkers/naming.py
```diff
@@ -255,6 +255,13 @@
         Since those words are used for BDD style it's also recommended not to use them within the keyword name.
         """,
     ),
+    "0319": Rule(
+        rule_id="0319",
+        name="deprecated-statement",
+        msg="This keyword '{{ keyword_name }}' is deprecated,"
+            " since version '{{ version }}.*', use {{ alternative }} instead",
+        severity=RuleSeverity.WARNING,
```
bhirsz, line 263: The min required version should be 5.0 since it's when those keywords started becoming truly deprecate

robocop/checkers/naming.py
```diff
@@ -662,3 +669,59 @@ def find_similar_variables(self, tokens, node):
...
+    def visit_TestSetup(self, node):  # noqa
+        self.check_if_keyword_is_deprecated(node.name, node)
+
+    def visit_Setup(self, node):  # noqa
+        self.check_if_keyword_is_deprecated(node.name, node)
+
+    def visit_SuiteTeardown(self, node):  # noqa
+        self.check_if_keyword_is_deprecated(node.name, node)
+
+    def visit_TestTeardown(self, node):  # noqa
```
bhirsz, line 699: One trick I learned recently in Python - if you have the same method body but different method names (typical in visitor pattern) you can do something like:
```
    def visit_SuiteTeardown(self, node):  # noqa
        self.check_if_keyword_is_deprecated(node.name, node)

   visit_TestTeardown = visit_Teardown ... (and other identical methods).. = visit_SuiteTeardown
``

MalikMlitat: was facing issues to apply suggested solution, can post more info later

robocop/checkers/naming.py
MalikMlitat: done

robocop/checkers/naming.py
MalikMlitat: thank you, done

MalikMlitat:
Please follow in : PR: #565

PR #572, opened 2022-03-15
Add duplicated-setting rule

### LINKED ISSUE #556: [Bug] ELSE IF parsing error message changed in RF 5
It's not catched in RF 5.0: `ELSE IF cannot have more than one '
 'condition`

bhirsz:
Also handle additional parsing errors and new error messages. Closes #556

It's possible that checks will fail because I have tested it on dev version - as it will be released.


robocop/checkers/errors.py
```diff
@@ -205,6 +205,21 @@
         severity=RuleSeverity.ERROR,
         version=">=4.0",
     ),
+    "0414": Rule(
+        rule_id="0414",
+        name="duplicated-setting",
+        msg="{{ error_msg }}",
+        severity=RuleSeverity.ERROR,
```
mnojek, line 212: Do we want this rule to be an error or warning? Does this violation break the execution or just makes some code being ignored?

robocop/checkers/errors.py
```diff
@@ -205,6 +205,21 @@
...
+        docs="""
+        Some settings can be only used once in file. Only the first value is used.
```
mnojek, line 214: Please also update tests
```suggestion
        Some settings can be used only once in a file. Only the first value is used.
```

robocop/checkers/errors.py
bhirsz: Right, it would be better if it were warning. Also it should belong to duplications group. However since it's catched parsing error in that case we should catch it there (and ignore) and catch it again in duplications - oor I could try if the new design of rules allow to defining rules in one place and importing elsewhere (here). I will test it out

bhirsz: Stwitched to warning but I had to move ``duplicated-setting`` to duplications while also catching (and ignoring) this error in errors.py. Importing the rule from other module is not that easy and it wasn't the good design to define the rule in one place and only use in the other.

mnojek (approved):

mnojek (approved):

PR #573, opened 2022-03-16
Extend reserved tags

### LINKED ISSUE #559: New reserved tags in RF 5.0
https://github.com/robotframework/robotframework/issues/4161

We should update the rule that warns about using robot: in tag name

bhirsz:
Closes #559

### DIFF
--- a/tests/packages/recent/requirements.txt
+++ b/tests/packages/recent/requirements.txt
@@ -1,1 +1,1 @@
-robotframework==5.0rc1
+robotframework==5.0rc2


mnojek (dismissed):

robocop/checkers/tags.py
```diff
@@ -46,10 +46,22 @@
     "0603": Rule(
         rule_id="0603",
         name="tag-with-reserved",
```
mnojek, line 48: BTW What about changing its name to `tag-with-reserved-word`? I don't know why wanted to keep it so short, that it doesn't make sense.

bhirsz: I agree. ``tag-with-reserved-word`` sounds better

mnojek (changes requested):
You also need to update the rule name here:
- docs/user_guide.rst:84
- tests/atest/rules/tags/tag-with-reserved-in-keyword-doc (change directory name)

mnojek (approved):

PR #576, opened 2022-03-22
Add and update rules for RETURN and BREAK,CONTINUE statements

bhirsz:
* add ``return-in-test-case`` rule
* update (new) deprecated-statement rule to also catch usage of ``Exit For Loop``, ``Exit For Loop If``, ``Continue For Loop``, ``Continue For Loop If``, ``Return From Keyword``. ``Return From Keyword If`` and``[Return]`` (instead of ``RETURN``)


bhirsz:
I will also update ``keyword-after-return`` rule inside this PR before the release (support also ``RETURN``). Docs for this rule will need some adjustments to be comprehensive for both <5 and >=5 RF users.

robocop/checkers/naming.py
```diff
@@ -676,55 +677,51 @@ def find_similar_variables(self, tokens, node):
...
-    def visit_TestSetup(self, node):  # noqa
-        self.check_if_keyword_is_deprecated(node.name, node)
-
-    def visit_Setup(self, node):  # noqa
-        self.check_if_keyword_is_deprecated(node.name, node)
-
-    def visit_SuiteTeardown(self, node):  # noqa
-        self.check_if_keyword_is_deprecated(node.name, node)
+    reports = ("deprecated-statement",)
+    depreccated_st = {5: {"[Return]"}}
```
mnojek, line 681: I would put a longer and more descriptive name for this variable. (also there was a typo)
```suggestion
    deprecated_statements = {5: {"[Return]"}}
```

robocop/checkers/naming.py
```diff
@@ -676,55 +677,51 @@ def find_similar_variables(self, tokens, node):
...
+    deprecated_kws = {
```
mnojek, line 682:
```suggestion
    deprecated_keywords = {
```

robocop/checkers/naming.py
```diff
@@ -676,55 +677,51 @@ def find_similar_variables(self, tokens, node):
 class DeprecatedStatementChecker(VisitorChecker):
     """Checker for deprecated statements."""
 
-    reports = (
-        "deprecated-statement",
-    )
-    # deprecated:alternative
-    deprecated_statements_rf3 = {}
-    deprecated_statements_rf4 = {"runkeywordunless": "IF"}
```
mnojek, line 684: So this rule should no longer be supported for RF 4?

robocop/checkers/naming.py
```diff
@@ -676,55 +677,51 @@ def find_similar_variables(self, tokens, node):
...
+        }
+    }
 
-    def visit_TestTeardown(self, node):  # noqa
-        self.check_if_keyword_is_deprecated(node.name, node)
+    def visit_KeywordCall(self, node):  # noqa
+        self.check_if_keyword_is_deprecated(node.keyword, node)
 
-    def visit_Teardown(self, node):  # noqa
+    def visit_SuiteSetup(self, node):
```
mnojek, line 698: I think you missed `# noqa` here

mnojek (changes requested):

robocop/checkers/naming.py
bhirsz: Good question actually - for Run Keyword If there is ``if-can-be-used`` rule (that works only for ==4.*). There is nothing for ``Run Keyword Unless``. So I can at least make this rule work for 4.0

bhirsz:
@mnojek I've also updated keyword-after-return extensively. It now supports ``RETURN`` statement in RF 5.0. I've also decided to remove ``keyword-after-return-from`` rule and merge it's behaviour into ``keyword-after-return``. It was making sense to have two separate rules before (kind of..) but now it's better to use one rule. WDYT?

robocop/checkers/misc.py
```diff
@@ -270,34 +269,49 @@ class ReturnChecker(VisitorChecker):
...
         return_setting_node = None
         keyword_after_return = False
         return_from = False
+        error = ""
         for child in node.body:
             if isinstance(child, Return):
                 return_setting_node = child
+                error = (
+                    "[Return] is not defined at the end of keyword. "
+                    "Note that [Return] does not return from keyword but only set returned variables"
```
mnojek, line 285: I guess that would make more sense:
```suggestion
                    "Note that [Return] does not quit from keyword but only set variables to be returned"
```

robocop/checkers/misc.py
```diff
@@ -270,34 +269,49 @@ class ReturnChecker(VisitorChecker):
...
+
+            if isinstance(child, KeywordCall):
                 if return_setting_node is not None:
                     keyword_after_return = True
                 if return_from:
                     token = child.data_tokens[0]
                     self.report(
-                        "keyword-after-return-from",
+                        "keyword-after-return",
+                        error_msg="Keyword call after 'Return From Keyword' keyword",
```
mnojek, line 303: I think we can safely remove this redundant word here:
```suggestion
                        error_msg="Keyword call after 'Return From Keyword'",
```

mnojek (approved):

mnojek (approved):

PR #578, opened 2022-03-23
Support inline IF in if-can-be-merged rule

### LINKED ISSUE #546: Add one line IF support
Robot Framework 5.0 will introduce one-liner for IF statements:

```
IF  ${val}==5  Keyword
```

Reference: robotframework/robotframework#4093

We need to support this new syntax and add new or update current rule.

bhirsz:
Closes #546


bhirsz:
I have added inline-if-can-be-used new rule. Nothing better than fresh rule right before release! ;)

robocop/checkers/misc.py
```diff
@@ -525,15 +564,47 @@ def check_adjacent_ifs(self, node):
...
+        for var1, var2 in zip(assign_1, assign_2):
+            if self.normalize_var_name(var1) != self.normalize_var_name(var2):
+                return False
+        return True
+
+    @staticmethod
+    def tokens_length(tokens):
+        return sum(len(token.value) for token in tokens)
+
+    def check_if_should_be_inline(self, node):
```
mnojek, line 597: I think I would rephrase the method name to `check_whether_if_should_be_inline` (or maybe even `check_whether_if_condition_should_be_inline`) because currently it is a bit confusing.

robocop/checkers/misc.py
```diff
@@ -525,15 +564,47 @@ def check_adjacent_ifs(self, node):
...
                 return False
             if_node = if_node.orelse
             other_if_node = other_if_node.orelse
         return if_node is None and other_if_node is None
 
+    @staticmethod
+    def normalize_var_name(name):
+        return name.lower().replace("_", "").replace(" ", "").replace("=", "")
+
+    def compare_assign(self, if_node, other_if_node):
```
mnojek, line 581: Maybe lest change the method name to `compare_assign_tokens`? `compare_assign` sounds strange.

robocop/checkers/misc.py
```diff
@@ -501,22 +528,34 @@ def visit_File(self, node):  # noqa
...
 
     def visit_Keyword(self, node):  # noqa
+        if get_errors(node):
+            return
         self.check_adjacent_ifs(node)
 
     visit_For = visit_If = visit_Keyword  # TODO  While, Try Except?
 
+    @staticmethod
+    def is_inline(node):
```
mnojek, line 549: Here maybe `is_if_inline`?

robocop/checkers/misc.py
```diff
@@ -261,6 +261,33 @@
         
         """,
     ),
+    "0916": Rule(
+        RuleParam(
+            name="max_width",
+            default=80,
+            converter=int,
+            desc="Maximum width of IF below which it will be recommended to use inline IF",
```
mnojek, line 269: Width calculated in which unit? Line length?
Then it should be stated here, like `"Maximum width of IF (in characters) below which it will be recommended to use inline IF" or something like this.

mnojek:
> I have added inline-if-can-be-used new rule. Nothing better than fresh rule right before release! ;)

Great! Thanks.
I put some comments and already updated release notes draft with it.

robocop/checkers/misc.py
bhirsz: Going for ``check_whether_if_should_be_inline`` (since it's whole IF that could be inline IF, not only condition)

robocop/checkers/misc.py
```diff
@@ -266,7 +266,7 @@
             name="max_width",
             default=80,
             converter=int,
-            desc="Maximum width of IF below which it will be recommended to use inline IF",
+            desc="Maximum width of IF (in character) below which it will be recommended to use inline IF",
```
mnojek, line 269: I missed the 's' in the previous suggestion:
```suggestion
            desc="Maximum width of IF (in characters) below which it will be recommended to use inline IF",
```

robocop/checkers/misc.py
```diff
@@ -594,8 +594,8 @@ def compare_assign(self, if_node, other_if_node):
     def tokens_length(tokens):
         return sum(len(token.value) for token in tokens)
 
-    def check_if_should_be_inline(self, node):
-        if self.is_inline(node):
+    def check_whether_if_should_be_inline(self, node):
```
mnojek, line 597: You need to change it also when it is called (line 558)

mnojek (changes requested):

mnojek (approved):

PR #581, opened 2022-03-24
Prepare code for version 2.0

mnojek:


robocop/checkers/naming.py
```diff
@@ -262,7 +262,7 @@
         msg="'{{ statement_name }}' is deprecated since Robot Framework version "
         "{{ version }}, use '{{ alternative }}' instead",
         severity=RuleSeverity.WARNING,
-        version=">=4.0",  # TODO change to 5.0 (and tests too)
+        version=">=5.0",
```
bhirsz, line 265: Switched it back to 4.0 - in previous PR we decided that deprecated-statement can support 4.0 (but only for Run Keyword Unless)

robocop/utils/misc.py
```diff
@@ -182,7 +182,7 @@ def find_similar(self, name, candidates):
         if not matches:
             return ""
         matches = self.get_original_candidates(matches, norm_cand)
-        suggestion = " Did you mean:\n"
```
bhirsz, line 185: Is it okay? It's displayed together with error message. For example:
Did not found X. Did you mean?

So missing space will result in ``Did not found X.Did you mean?``

bhirsz: Ah nevermind, I see you've updated the space in other place.

bhirsz (approved):

PR #597, opened 2022-04-09
Fix rule 0912 to include type of var in rule message

### LINKED ISSUE #592: [Bug] Rule 0912 Incorrectly recommend to use ${EMPTY}  for List and Dictionaries
### What happened?

 Rule 0912 Incorrectly recommend to use of ${EMPTY}  for lists
 Rule 0912 Incorrectly recommend to use of ${EMPTY}  for dictionary

Code below show the problem with ${EMPTY} is not the same as blank for lists ->  [''] != []
It`s  not valid entry for dictionary filled with ${EMPTY} 

*** Variables ***
@{A}    ${EMPTY}
@{B}
&{C}
&{D}    ${EMPTY}     #invalid entry 
*** Test Cases ***
Test
    Log To Console    ${A}
    Log To Console    ${B}
    Should Be Equal    ${A}    ${B}

### Operating System

N/A

### Robocop version

2.0.1

mnojek:
Fixes #592


robocop/checkers/misc.py
```diff
@@ -493,11 +493,12 @@ def visit_Variable(self, node):  # noqa
         if get_errors(node):
             return
         if not node.value:  # catch variable declaration without any value
-            self.report("empty-variable", node=node)
+            self.report("empty-variable", var_type=node.name[0], node=node)
```
bhirsz, line 496: It's too simplistic for our use case - note that if list/dict is empty:
```
@{LIST_NO_VALUE}
```
it' should in fact have ``@{EMPTY}`` value. But list/dict with values, where only some are empty:
```
@{MULTILINE_WITH_MULTIPLE_EMPTIES}
...  1
...
...  a
```
should use scalar syntax:
```
@{MULTILINE_WITH_MULTIPLE_EMPTIES}
...  1
...  ${EMPTY}
...  a
```

bhirsz: Same with list of empty strings - it would be:
```
@{MULTILINE_WITH_MULTIPLE_EMPTIES}
...    ${EMPTY}
...    ${EMPTY}
```
and not 
```
@{MULTILINE_WITH_MULTIPLE_EMPTIES}
...    @{EMPTY}  # this indicate sublist ie [[], []]
...    @{EMPTY}
```

mnojek: You're right, but we cannot predict what type of empty variable user wants, so the only possible output in my opinion is:
- we stick to what we had suggesting always `${EMPTY}` or even we can put an asterisk there with an explanation that type desired type needs to be put by the user,
- we stick to what I proposed and it's user's responsibility to propose expected type.

bhirsz: In this case we can. If we don't have any value at all:
```
@{NO_VALS}
&{NO_VALS}
```
we can suggest using @/&{EMPTY}. If the user actually intended to use list with one empty string he can notice from robocop output that he used wrong syntax. 

If the list/dict contains values but value is empty string - it is interpreted as ${EMPTY} anyway so no harm telling to use it explicitly:
```
@{EMPTY_VALS} 
...  1
...  # should be ${EMPTY}
```
Again, from the error message user will see that it's read as ${EMPTY} so if he wants nested empty list for example, then instead of "" he/she should use @{EMPTY}

mnojek: OK, I added support for handling such cases.

Please note that for this case:
```
&{DICT_NO_VALUE}
...
```
it will not print the rule because it already raises RF error with message:
`"Invalid dictionary variable item ''. Items must use 'name=value' syntax or be dictionary variables themselves."`

mnojek: Please review and let me know if such handling is enough or we still may need to support some edge cases.

bhirsz: It looks like we're still not there. Look:
```
${rules_dir}${\}test.robot:7:3 [I] 0912 Use built-in variable @{EMPTY} instead of leaving variable without value or using backslash
```
It should be ${EMPTY} since it's just empty string (it's inside list but it's inrevelant). 

See examples from linked bug:
```
*** Variables ***
@{A} ${EMPTY}
@{B}
&{C}
&{D} ${EMPTY} #invalid entry
@{E} 
...   # I added it
```
it should:
```
*** Variables ***
@{A} ${EMPTY}   # do nothing
@{B}   # report @{EMPTY} should be used
&{C}  # do nothing since parsing-error should be raised (check)
&{D} ${EMPTY}  # do nothing - parsing-error should be raised
@{E} 
...   # I added it  # report ${EMPTY} should be used
```

mnojek: I don't really get the difference between:
```
@{B}    # report @{EMPTY}
```
and
```
@{E}
...    # report ${EMPTY}
```
When there's only one value for each of those and the only difference is that in the second case the value is on a separate line. I'm not saying, that I am not wrong, but I just don't see a difference.

mnojek: Ok, fixed it and now it should be fine.

bhirsz (approved):

PR #598, opened 2022-04-10
Remove deprecation for 'Run Keyword Unless' in RF4

### LINKED ISSUE #593: [Bug] Weird message of rule 0319
### What happened?

Hi. On RF 4.1.3, the rule says "'Run Keyword Unless' is deprecated since Robot Framework version **4**.*". 

But on RF 5.0, it says "'Run Keyword Unless' is deprecated since Robot Framework version **5**.*" etc. Both cannot be true. :)

### Operating System

_No response_

### Robocop version

2.0.1

mnojek:
Fixes #593 

I also changed the name for RF versions in unit tests to better reflect what is 'stable3' etc. in summary.


robocop/checkers/naming.py
```diff
@@ -262,7 +262,6 @@
         msg="'{{ statement_name }}' is deprecated since Robot Framework version "
         "{{ version }}, use '{{ alternative }}' instead",
         severity=RuleSeverity.WARNING,
-        version=">=4.0",
```
mnojek, line 265: This rule should not have a restriction on which versions to run because the keywords themselves have a version associated with them in line 645.

bhirsz: I agree. Now I see we don't have docs here - should we add more explanation with the list of the deprecated methods/statement and their alternatives per version?

bhirsz: It's how it's displayed in docs now:
image

robocop/checkers/naming.py
```diff
@@ -644,7 +643,6 @@ class DeprecatedStatementChecker(VisitorChecker):
 
     reports = ("deprecated-statement",)
     deprecated_keywords = {
-        4: {"runkeywordunless": "IF"},
```
bhirsz, line 647: It's fine fix for now. But I see the problem (and that's why the bug was raised). If the RF 6 will be released this report will stop displaying deprecation messages even if the keyword was deprecated in RF 5. Perhaps it could be reverse: the deprecated statements should be keys, and values would be (alternative, version). That way you would just check if statement is deprecated, and if get value to display message.

bhirsz: Then we of course would need to perform extra check if actual version is equal or bigger than stored version.

robocop/checkers/naming.py
mnojek: > I agree. Now I see we don't have docs here - should we add more explanation with the list of the deprecated methods/statement and their alternatives per version?

Yes, I think that's a nice idea.

robocop/checkers/naming.py
mnojek: Are we going to merge this one or it should include your proposed changes?

bhirsz: We can merge this one. Proposed change can go in separate PR in the future - we have time before 6.0 release.

bhirsz (approved):

PR #601, opened 2022-04-10
Update docs

mnojek:
- Updated examples in `README.md` and `docs/user_guide.rst`
- Added correct plural forms whenever it was possible
- Added examples to docs in reports

Let me know if these changes are an improvement. I can see a case when some users expect specific output from Robocop run and these changes can break it. If you do not consider this an improvement, we can ignore those changes and abandon them.


bhirsz (approved):
I like the changes - especially that we have some rules that are already more explicit ("Issue found in X statement" rather than vague message like "This tag").

PR #602, opened 2022-04-10
Add detection for documentation in templated test cases

### LINKED ISSUE #526: [Bug] Ignore rule 0202 when using test templates
### What happened?

Hello,

Rule 0202 requires each test to have a documentation.
However when using test templates the goal is to simplify the scripts, hence skipping the documentation (for which the keyword should contain a documentation and the test instance should have a meaningful name)
I can simply disable this rule locally but I think it would be better if it could be disabled automatically when a template is used

Thanks,

Raphael

### Operating System

MacOS Big Sur

### Robocop version

1.12.0

mnojek:
Fixes #526 

@bhirsz Can you help me with tests for this depending on the value of the parameter?
Currently, the `ignore_templated` param for `missing-doc-test-case` is set to `True`, but the tests only work when it is set to `False`. We need to add tests checking both of the values for the param. I need guidance.


robocop/checkers/documentation.py
```diff
@@ -24,6 +24,12 @@
         """,
     ),
     "0202": Rule(
+        RuleParam(
+            name="ignore_templated",
+            default=True,
+            converter=bool,
```
bhirsz, line 30: It's a common pitfall: bool(str). bool("False") is actually true since it evaluates it as non-empty truthy string. You can use this method instead: https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

bhirsz: ^Above will return int, so you need to convert it to bool.

Optionally write your own converter:
def str2bool(v):
  return v.lower() in ("yes", "true", "t", "1")
  
(from SO ;) https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python)

mnojek: Fixed it. Can you take a look?

bhirsz (approved):

Issue #595, opened 2022-04-06
[Bug] Any text before dot in keyword name is acknowledged as library name


p74:
### What happened?

Bug details

When Using gherkin with a  dot . 
And a ppp file '1.log'  is loaded

rule 0302 
-c 0302:convention:first_word_capitalized

Keyword is fine but Robocop complains with error:

Keyword name "And a ppp file '1.log' is loaded''' should use title case  

Robocop expect "Log" to be written as first word capitalized but log should be ok as well in this case

Greetings Ruud

### Robocop version

2.0.1

mnojek:
Thanks @p74 for reporting this issue.
This is due to the fact that we normalize the keyword name before running a rule on it and in this case everything before the dot is removed and we only parse `log is loaded` which leads to the reported output. I will take a look how we can handle such cases in the code.

p74:
I know gherkin is bit difficult  to handle for tools like this as it's difficult to discover if it's gherkin or keyword  but in this case I have some ideas for example ignore what is between ' and "   (variables are defined between quotes) Ignore the last char.  Or make an option to ignore this dot..   I expect more gherkin related issues in  the rules like the number of arguments but this I did not check

mnojek:
The problem here is how we discover keyword's library prefix. It's too simple approach and I need to think about something more precise. Gherkin is already supported and it's not the issue here :) I think I will have some time to look at it today, but I can't promise.

bhirsz:
Yes - the problem is with how we discover keyword prefix. Robot Framework uses quite complex logic here. If you have "Some.Text" keyword, it will check if "Some.Text" keyword exist somewhere, and if not it will check if "Some" library exists, and if so it will check it for "Test" keyword (simplifying a bit). In Robocop we don't have access to libraries resources so we need to guess if something is library name or not. Library names are not checked. In your case with the keyword:
```
And a ppp file '1.og' is loaded
```
we are discarding everything before . and we're only checking:
```
log' is loaded
```

I think we should improve how we recognize library name - for example we can agree that libraries don't have ``'`` in its names - just alpha numeric (however it could be also robot ${Variable}). We need to investigate what is allowed for library name and use as pattern to.

bhirsz:
During last few years there were few revision of the rule that helped a bit with the issue. It cant completely go away but I have decided to accept this limitation as known issue, since I dont want to read library /resource keywords during linting. Closing the issue

PR #615, opened 2022-06-06
added configurable regex pattern

jannek76:
It will be useful to use configurable regex patterns to allow more than just chars in filename, keyword, suite or test case names.

Example we like to allow `foo.bar` in keyword names but not `foo.` or `.bar` or any other words that have `.` in Keyword.
Currently this is not possible and I think that this will be useful for other users also.

Also I made new rule `320` for filenames. So rules can be different for suite names.

With this PR Bug #595 has workaround and accepted filenames can be configured.


robocop/checkers/naming.py
```diff
@@ -263,43 +269,83 @@
...
-                    self.report("not-allowed-char-in-name", character=char, block_name="suite", node=node)
+            for iter in self.param("not-allowed-char-in-filename", "pattern").finditer(suite_name):
+                self.report(
+                    "not-allowed-char-in-filename",
+                    character=iter.group(),
+                    block_name="suite",
+                    node=node,
+                    col=node.col_offset + iter.start(0) + 1,
+                )
         super().visit_File(node)
```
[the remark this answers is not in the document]
bhirsz: Optionally we can just add ``"not-allowed-char-in-filename"`` rule to ``InvalidCharactersInNameChecker``. That way we don't need to introduce new checker (one checker can check for two or more related rules).

robocop/checkers/naming.py
```diff
@@ -263,43 +269,83 @@
...
-                if self.param("not-allowed-char-in-name", "pattern").search(node.name[index]):
-                    self.report(
-                        "not-allowed-char-in-name",
-                        character=node.name[index],
-                        block_name=f"'{node.name}' {name_of_node}",
-                        node=node,
-                        col=node.col_offset + index + 1,
-                    )
-                index += 1
+        node_name = node.name
```
bhirsz, line 324: I still vividly remember I had to introduce some hacks here to handle variables, patterns and still be able to pinpoint location of the issue. I now need to checkout your branch and debug that part a bit to fully get it :)

bhirsz:
Thanks for the PR, the change looks good - the separate rule for filename & extra pattern possible. Give me few days though as I want to read the code and run it. ``wrong-case-in-keyword-nam`` and similar rules are most problematic rules in Robocop history after all :D

robocop/checkers/naming.py
jannek76: I'll add some comments to code.
Your code was looping `node.name` char by char and dropping variables off if there was one. I had to change variables searching, because in my proposal will search pattern from fully string.

bhirsz (approved):
LGTM! I've tried your code and it works :)

PR #616, opened 2022-06-06
Add Robocop version report

rikerfi:
Hi,

it would be helpful for test developers and DevOps staff to know from reports the actual RoboCop version used. Sometimes experienced a mismatch between setups which gave differences in results.

Regards,
Risto


bhirsz:
Thanks for the PR - it looks good and useful!

mnojek (approved):

PR #630, opened 2022-07-11
Introducing timestamp report

rikerfi:
Hi,

here we go with a draft PR for the timestamp report. Please, feel free to comment especially regarding documentation. Needs testing on different systems (Win, Mac, Linux).

-Risto


rikerfi:
@bhirsz, @mnojek: Seems that installing dependencies are missing `pytz` and `python-dateutil`. How do I do updating the dependencies?

bhirsz:
Thanks for your question I've noticed we can improve dependency handling a bit (I did it recently in robotidy). But I will open separate issue for it. Right now if you want to update dependencies you need to do it in 2 places:

1) setup.py : https://github.com/MarketSquare/robotframework-robocop/blob/2.2.0/setup.py
2) github unitest workflow https://github.com/MarketSquare/robotframework-robocop/blob/1bc9b9c728fd17cfff9668a5b07cfdf30d7074df/.github/workflows/unit-tests.yml#L50 (add to line)

robocop/reports.py
```diff
@@ -228,3 +232,95 @@ def add_message(self, *args):
...
+        self.description = "Returns Robocop execution timestamp."
+        self.timezone = "local"
+        self.format = "%Y-%m-%d %H:%M:%S %z"
+
+    def configure(self, name, value):
+        if name != "conf":
+            super().configure(name, value)
+        value_err = True
+        if value.startswith("timezone="):
+            conf = value.removeprefix("timezone=").split(":", maxsplit=1)
```
bhirsz, line 291: Didn't know this useful utility method - though it's introduced in Python 3.9 so we can't use it (we support 3.7+).

robocop/reports.py
```diff
@@ -228,3 +232,95 @@ def add_message(self, *args):
...
+    """
+
+    def __init__(self):
+        self.name = "timestamp"
+        self.description = "Returns Robocop execution timestamp."
+        self.timezone = "local"
+        self.format = "%Y-%m-%d %H:%M:%S %z"
+
+    def configure(self, name, value):
+        if name != "conf":
```
bhirsz, line 287: Why we're using conf? If we drop it, we could directly supply params to ``timestamp`` report. 
So instead of 
```
--configure timestamp:conf:timezone="Europe/Paris":format="%Y-%m-%d %H:%M:%S %Z %z"
```
we could use
```
--configure timestamp:timezone="Europe/Paris":format="%Y-%m-%d %H:%M:%S %Z %z"
```
Is is somehow related to how Robocop parses ``:`` inside values? (just guessing, if there is any problem with defining the format)

robocop/reports.py
rikerfi: Ooops, I need to refactor that. Thanks for pointing out the issue.

robocop/reports.py
rikerfi: I was wondering that as well. The reason was that I thought it requires (name, value) pair. I'll study that a bit more.

bhirsz: Yeah it does. But in this case either ``format`` or ``timezone`` will be passed as name. If you supply
```
timestamp:timezone="Europe/Paris":format="%Y-%m-%d %H:%M:%S %Z %z"
```
configure function will be invoked twice for format and timezone separetly. So it does some preparing for you (it uses the same logic when passing configurables for rules).

Here is the code that does that: https://github.com/MarketSquare/robotframework-robocop/blob/24f51118cc8dd776f6af9b9d924e49d0aeb594e4/robocop/run.py#L249

robocop/reports.py
bhirsz: Actually scratch my last advice - I forgot that it's already prepared so name variable will be already either "format" or "timezone" and value will contain the value for format/timezone. It should be easier to parse it

rikerfi:
I am really not happy with the implementation. Back to the drawing board.

Issue #628, opened 2022-07-04
Create at least one report that logs date and time of checking RF scripts


UliSei:
When I run robocop with 
`-o myfolder/robocop.log`
saving the logs at the same place, I just need to refresh the log file.
But I want to be sure that the most recent robocop result is displayed.

Currently I run Robocop with all reports and I get a log like this:

![grafik](https://user-images.githubusercontent.com/89480399/177187234-0c65c627-5c06-4bb9-8a6e-da4ded984bc2.png)

Wouldn't it be good to have at least one report that logs date and time when robocop was ended?

Either as an own report or
as extension of existing reports, e.g.:

- Report generated on 2022-07-04 17:50 by Robocop version: 2.1.0, 
- Scan finished in 5.973s on 2022-07-04 17:50

bhirsz:
Good idea! Not sure though if we should create new report or reuse existing one (since it isn't a lot to add). @rikerfi does report generation time sounds good for version report or it should be printed somewhere else?

mnojek:
I think I would go with a separate report, since it's not obvious that version report will also report the datetime, but I don't have a strong opinion and would like to hear from others.

UliSei:
Maybe the datetime info is more related to robocop.reports.TimeTakenReport?
You get the time that was needed for the scan and additionally the timestamp when the scan had ended.
So there is even implicit the info when the scan was started.

UliSei:
A separate report has some advantages: 

- it can be switched on or off just when needed independently from existing reports
- there is a low risk of messing up with consumers of the log which expect the status quo of existing reports output.

bhirsz:
Timezones and time conversions, here we go again.. ;) OK, I don't have strong feelings in the subject so we can create separate report and make it configurable like @rikerfi suggested.

rikerfi:
Time, what a pain in computer science :smile:
I could draft a report proposal during the weekend. Probably, I need to study configuration options a bit.

bhirsz:
Implemented with timestamp report in #637. It will be released today as part of 2.3.0 release.

UliSei:
Great - I like it!

PR #637, opened 2022-07-14
Timestamp report

rikerfi:
Timestamp report implementation, tests, and documentation.

Report that returns Robocop execution timestamp.
Timestamp follows local time in format of `Year-Month-Day Hours(24-hour clock):Minutes:Seconds ±hh:mm UTC offset` as default.

Example:
`Reported: 2022-07-10 21:25:00 +0300`

Supports `timezone` configuration and various `format` options for the timestamp.

Resolves:  Create at least one report that logs date and time of checking RF scripts #628


robocop/run.py
```diff
@@ -248,22 +248,46 @@ def any_rule_enabled(self, checker) -> bool:
...
+        report_name, param_and_value = config.split(":", maxsplit=1)
+        if report_name == "return_status":
+            if param_and_value.count(":") < 1:
+                raise robocop.exceptions.ConfigGeneralError(
+                    f"Provided invalid config: '{config}' (general pattern: <report>:<param>:<value>)"
+                )
+            param, value = param_and_value.split(":", maxsplit=1)
+            self.reports[report_name].configure(param, value)
+        elif report_name == "timestamp":
+            self.reports[report_name].configure(None, param_and_value)
```
bhirsz, line 289: Why the difference between timestamp and other reports? For me it could be handled the same any other reports is handled. But I could be missing something

```
robotidy --configure timestamp:format:<format>
robotidy --configure timestamp:timezone:<format>
```

robocop/run.py
```diff
@@ -248,22 +248,46 @@ def any_rule_enabled(self, checker) -> bool:
...
+            )
+        rule_name, param, value = config.split(":", maxsplit=2)
+        if rule_name in self.rules:
+            self.rules[rule_name].configure(param, value)
+
+    def _configure_reports(self, config):
+        if config.count(":") < 1:
+            raise robocop.exceptions.ConfigGeneralError(
+                f"Provided invalid config: '{config}' (general pattern: <report>:<param>:<value>)"
+            )
```
bhirsz, line 279: We could define specific exception to avoid duplicating lines. ie:

```
class InvalidConfigPatternError(ConfigGeneralError):
    def __init__(self, config):
        msg = f"Provided invalid config: '{config}' (general pattern: <report>:<param>:<value>)"
        super().__init__(msg)
        
raise InvalidConfigPatternError(config)
```

robocop/run.py
rikerfi: Aaah, I was stuck in the thought that both `timezone` and `format` should be configured together... If we drop that, then configure could be like mentioned. That makes implementation much more straightforward.

bhirsz: Yeah, they can be configured separetly - the reports are singletons and they preserve all configs. So:

```
robotidy --configure timestamp:format:<format> --configure timestamp:timezone:<timezone>
```
will configure timestamp with format and timezone.

rikerfi: Now configurations are handled separately:

Both of default values, ``timezone`` and ``format`` can be configured by
    ``-c/--configure`` and ``timestamp:timezone:"<timezone name>"`` and/or ``timestamp:format:"<format string>"``::

        robocop --configure timestamp:timezone:"Europe/Paris" --configure timestamp:format:"%Y-%m-%d %H:%M:%S %Z %z"

This yields following timestamp report::

         Reported: 2022-07-10 20:38:10 CEST +0200

robocop/run.py
rikerfi: I think, simplifying the implementation solved also the issue topic...

robocop/reports.py
```diff
@@ -228,3 +232,96 @@ def add_message(self, *args):
...
+                self.format = conf_value
+            else:
+                raise ValueError()
+        except ValueError:
+            raise robocop.exceptions.ConfigGeneralError(
+                f"""Provided configuration '{value}' for report '{getattr(self, 'name')}' is not valid.
+        General pattern: timestamp:timezone="<timezone name>":format="<format string>"""
+            )  # noqa
+
+    def add_message(self, *args):
```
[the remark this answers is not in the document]
rikerfi: Good point! `Report` class has now `add_message` method.

robocop/reports.py
```diff
@@ -223,8 +226,79 @@ def __init__(self):
...
         self.description = "Returns Robocop version"
 
-    def add_message(self, *args):
-        pass
-
     def get_report(self) -> str:
         return f"\nReport generated by Robocop version: {__version__}"
+
+
+class TimestampReport(Report):
```
bhirsz, line 233: I have noticed few (really) small issues in docs here (I will point them out in comments) so I have generated docs locally so it's easier to see them:
image

You can also generate docs by running ``tox -e docs`` in the root of the repository. They will be generated in docs/_build/index.html .

robocop/reports.py
```diff
@@ -223,8 +226,79 @@ def __init__(self):
...
+
+        Reported: 2022-07-10 21:25:00 +0300
+
+    Both of default values, ``timezone`` and ``format`` can be configured by
+    ``-c/--configure`` and ``timestamp:timezone:"<timezone name>"`` and/or ``timestamp:format:"<format string>"``::
+
+        robocop --configure timestamp:timezone:"Europe/Paris" --configure timestamp:format:"%Y-%m-%d %H:%M:%S %Z %z"
+
+    This yields following timestamp report::
+         Reported: 2022-07-10 20:38:10 CEST +0200
```
bhirsz, line 251: ``::`` blocks require one empty line before/after wrapped code
```suggestion
    This yields following timestamp report::

         Reported: 2022-07-10 20:38:10 CEST +0200
```

robocop/reports.py
```diff
@@ -223,8 +226,79 @@ def __init__(self):
...
+        self.name = "timestamp"
+        self.description = "Returns Robocop execution timestamp."
+        self.timezone = "local"
+        self.format = "%Y-%m-%d %H:%M:%S %z"
+
+    def configure(self, name, value):
+        if name == "timezone":
+            self.timezone = value
+        elif name == "format":
+            self.format = value if value else f'''{self.format}
```
bhirsz, line 286: Is "Empty format string for `timestamp` report does not make sense. Default format used" string going to be saved to self.format? It looks that way from the code. I guess it should be something akin:

```
if value:
    self.format = value
else:
    print("Empty format string bla bla") <- since we're using defaunt, no need to set self.format
```

robocop/reports.py
```diff
@@ -223,8 +226,79 @@ def __init__(self):
...
+            super().configure(name, value)
+
+    def get_report(self) -> str:
+        return f'Reported: {self._get_timestamp()}'
+
+    def _get_timestamp(self) -> str:
+        try:
+            if self.timezone == 'local':
+                return datetime.now(tz.tzlocal()).strftime(self.format)
+            return datetime.now(pytz.timezone(self.timezone)).strftime(self.format)
```
bhirsz, line 298: Moved code around a bit
```suggestion
            timezone = tz.tzlocal() if self.timezone == "local" else pytz.timezone(self.timezone)
            return datetime.now(timezone).strftime(self.format)
```

bhirsz:
Thanks for the PR - it looks good - just left some comments for small things (for code refactors it's up to you if you want to leave it like it is).

robocop/reports.py
rikerfi: Thanks. That helped a lot in making docs :smile:

robocop/reports.py
rikerfi: Fixed.

robocop/reports.py
```diff
@@ -223,8 +226,79 @@ def __init__(self):
...
+         Reported: 2022-07-10 20:38:10 CEST +0200
+
+    For timezone names,
+    see: https://en.wikipedia.org/wiki/List_of_tz_database_time_zone
+
+    For timestamp formats,
+    see: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
+
+    Useful configurations::
+        Local time to ISO 8601 format:
```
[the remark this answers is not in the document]
rikerfi: Fixed

robocop/reports.py
rikerfi: Nice catch! Actually, changed it to use `UserWarning` without the hassle report itself.

robocop/reports.py
rikerfi: Love it! Much better with only one exit point. Fixed.

bhirsz (approved):
Good work! It's mergable, I will do it when I'll be on PC (tommorow probably).

PR #660, opened 2022-07-26
New reserved tags in RF 5.1

### LINKED ISSUE #640: Support robot:private tag for private keywords in RF 5.1
Support new ``robot:private`` tag introduced in https://github.com/robotframework/robotframework/issues/430 . 
Most likely it only requires updating the rule with the reserved tags names.

We can additionaly check if the new feature - private keywords - may need any extra rule from Robocop side.

### LINKED ISSUE #645: Update reserved tags in RF 5.1
New reserved tag ``robot:stop-on-failure``. Also check if there are any other reserved tag (except for #640).

bhirsz:
Also small fix for checking the private tags - the tag name is now case insensitive.

Closes #640 
Closes #645


robocop/checkers/tags.py
```diff
@@ -172,7 +174,9 @@ class TagNameChecker(VisitorChecker):
         "robot:recursive-continue-on-failure",
         "robot:skip",
         "robot:skip-on-failure",
+        "robot:stop-on-failure",
         "robot:exclude",
+        "robot:private",
```
mnojek, line 179: Shouldn't we test against all of those reserved tags?

bhirsz: I've decided to go with equivalence class theory here - we're testing for most of them and we're using this set in our code, so we don't need to cover all possible options.

mnojek (approved):

PR #662, opened 2022-07-26
Do not include all reports when using --reports all

### LINKED ISSUE #554: Reports json_report not working from command line execution[Bug]
### What happened?

Hi, 
Reading the doc seems i can get a report with the list of issues found in JSON format with the command `robocop -r json_report -o report.json file/to/parse.robot`.
But seems this doesn't work and instead I get a file with the classic text line `"{source}:{line}:{col} [{severity}] {rule_id} {desc} ({name})"`

Is it a bug or am i missing something? thank you very much

### Operating System

10.0.19042

### Robocop version

1.12.0

bhirsz:
When user used ``robocop --reports all`` all reports were included - and some of them may not be desired (like JsonReport which for now is internal-only report or Sarif report). 

Added ``DEFAULT`` class attribute that allows to decide if the report should be included in `all` reports.

Removed JsonReport from reports documentation - since it's internal only report. Thanks to ``DEFAULT`` attribute it will also not shown when running ``robocop --list-reports``. Fixes #554 (we need separate issue for refactoring JSON report so it does work externally as well).


robocop/reports.py
```diff
@@ -30,6 +30,16 @@
 
 
 class Report:
+    """
+    Base class for report class.
+    Override `configure` method if you want to allow report configuration.
+    Override `add_message`` if your report process the Robocop issues.
```
mnojek, line 36:
```suggestion
    Override `add_message`` if your report processes the Robocop issues.
```

mnojek (approved):

mnojek (approved):

Issue #614, opened 2022-05-31
Rule severity depending on thresholds


adrszad:
I'd like to configure e. g. line-too-long to raise warning if length>140, but error when length>200. The same for other parameterized rules.

mnojek:
Nice idea! I need to take a look if it's a low-hanging fruit or it requires significant changes to the code.

bhirsz:
I have an idea how it could be implemented - I will try it. 

The basic design for now: introduce new class ``SeverityThreshold`` which can be added to Rule definition. It should contain name of the RuleParam that 'controls' severity of the rule:
![image](https://user-images.githubusercontent.com/8532066/181704561-fe86fc7b-fa07-434a-8bb3-576641c07eee.png)

It could be then configured the same way return status is configured:

```
robocop -c line-too-long:severity_threshold:warning=140:error=200
```

By default it would support int but if there is need we could add support for other types (like boolean).

bhirsz:
I've manage to implement this. I'm currently working on updating our engine for generating the docs and tests. Docs example:

![image](https://user-images.githubusercontent.com/8532066/181724722-2df235c7-e705-40c7-9860-cc550dbf4ac8.png)

PR #669, opened 2022-07-29
Add too-few-calls-in-test-case rule

### LINKED ISSUE #610: [Rule] Test case without keyword
Test case without keyword yields FAIL in RF. This might not be desired. So users should be guided to use a keyword or set the desired result by PASS, FAIL or SKIP keyword.

```
*** Test Cases ***
I Just Have TC Idea
    [Documentation]  Test design on high level ongoing. No Keywords.

```
This fails in RF, but maybe Robocop could warn/info something like `Test contains no keywords will FAIL, use a keyword or explicitly set result using PASS, SKIP or FAIL keywords.`  

```
*** Test Cases ***
I Just Have TC Idea
    [Documentation]  Test design on high level ongoing.
    SKIP    Just an idea.
```

bhirsz:
Closes #610

### DIFF
--- a/robocop/checkers/lengths.py
+++ b/robocop/checkers/lengths.py
@@ -189,6 +189,22 @@ rules = {
         msg="Too many test cases ({{ test_count }}/{{ max_allowed_count }})",
         severity=RuleSeverity.WARNING,
     ),
+    "0528": Rule(
+        RuleParam(name="min_calls", default=1, converter=int, desc="number of keyword calls required in a test case"),
+        rule_id="0528",
+        name="too-few-calls-in-test-case",
+        msg="Test case '{{ test_name }}' has too few keywords inside ({{ keyword_count }}/{{ min_allowed_count }})",
+        docs="""
+        Test without keywords will fail. Add more keywords or set results using Fail, Pass, Skip keywords::
+        
+            *** Test Cases ***
+            Test case
+                [Tags]    smoke
+                Skip    Test case draft
+
+        """,
+        severity=RuleSeverity.ERROR,
+    ),
 }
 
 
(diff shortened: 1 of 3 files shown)


robocop/checkers/lengths.py
```diff
@@ -179,6 +179,22 @@
...
+        docs="""
+        Test without keywords will fail. Add more keywords or set results using Fail, Pass, Skip keywords::
+        
+            *** Test Cases ***
+            Test case
+                [Tags]    smoke
+                Skip    Test case draft
+
+        """,
+        severity=RuleSeverity.ERROR,
```
mnojek, line 196: The only concern I have is that if it's configurable, it shouldn't be an Error when someone configured the rule to 2 or more min_calls. Maybe it shouldn't be configurable and it should be named `no-keywords-in-test-case`?
We can also add one more rule that does something similar to this one you implemented but make it with Warning severity? I don't know. What are your thoughts, @bhirsz ?

bhirsz: Yeaaah, if someone configure it, it would stop being an error. I think we can do something else - set the severity to Warning and incoming #614 would allow for user to set their own severity depending on value. 

Or I can later add this feature. where severity would be dynamic by default, for 0 it would throw error and 1+ warning. Basically configure severity_threshold for this rule in our code.

mnojek: I still think that maybe there should be a separate rule that throws errors when there is no keyword at all. But 2 rules may overlap if configured with the same number which would be strange...

bhirsz: Both option have its pros and cons, I need to think about this. Right we can either:
 - create two separator rules, one with Error severity for 0 keyword calls
 - leave this rule, but allow to configure the severity
  - as suboption, we can by default configure error for 0 calls, and warning for 1+ calls (if someone configure it to trigger on <x )

One note that other rules like too-few-calls-in-keyword will have configurable severity depending on calls amount - so it would be more consistent to go with second option.

mnojek: Let's do the second option, like you proposed.

bhirsz: Sure, it will be done after #614 is merged.

mnojek (approved):

PR #672, opened 2022-07-29
Allow to configure rule severity depending on parameter value

bhirsz:
Closes #614

PR #674, opened 2022-08-04
configurable terms for todo-in-comment

rikerfi:
It would be good if `todo-in-comment` rule has configurable terms.

E.g. `robocop --configure todo-in-comment:todos:OMG,bug` also this work `robocop --configure "todo-in-comment:todos:remove me,fix this ugly bug"` This should enable also terms in local language.

~~Work in progress: documentation, tests~~

What would be the best location to place rule configuration tests?


bhirsz:
Like you probably noticed our 'test framework' auto generates test for all rules with default configs. To test rule with custom config we can use following file:

https://github.com/MarketSquare/robotframework-robocop/blob/master/tests/atest/custom_tests.yaml

Either add entry with name of the rule / or update if the rule is already there with your desired config and path to test data files.

For example:
```
  inconsistent-assignment:
    - config: inconsistent-assignment:assignment_sign_type:space_and_equal_sign
      src_dir: misc/inconsistent-assignment-const
    - config: inconsistent-assignment:assignment_sign_type:autodetect
      src_dir: misc/inconsistent-assignment-autodetect
```
Runs two separate tests for rule ``inconsistent-assignment`` with different config and test data directories. 

This:
```
  not-allowed-char-in-filename:
    - config: not-allowed-char-in-filename:pattern:\.(?!bar)
      src_dir: naming/not-allowed-char-in-filename-configured
```
is equivalent of running (in the tests/atest directory):
```
robocop --include not-allowed-char-in-filename --configure  not-allowed-char-in-filename:pattern:\.(?!bar) naming/not-allowed-char-in-filename-configured
```

Config can be single string (if you're only configuring one param) or list of strings (after #672 will be merged). But I think you don't need to configure more than one param per test so you don't need to wait for 672 and just use string for configuration.

bhirsz:
I actually put this information in our CONTRIBUTING docs but it's less detailed here, maybe I should update it soon ;) This can get confusing - our test framework is useful for our needs but even I (author) tend to forgot how it works with autogeneration/custom tests.

robocop/checkers/comments.py
```diff
@@ -150,19 +151,14 @@ def check_invalid_comments(self, name, node):
 
     def check_comment_content(self, token, content):
         content = content.lower()
-        if "todo" in content:
+        todos = self.param("todo-in-comment", "todos").split(",")
```
rikerfi, line 154: Probably there is a better place to construct `todos` list than here for every comment.

bhirsz: This is gonna get complicated soon, but in such case you can create custom converter. Like you noticed you cannot construct todos list in checker ``__init__`` because init will be called before rule is configured (so we will not get updated todos list). 

I was in similar situation with ``inconsistent-assignment`` rule - I wanted user to configure it using user friendly names (like 'equal_sign') but for checker purposes I wanted to save different value internally. For that reason I created custom converter:
https://github.com/MarketSquare/robotframework-robocop/blob/111b1f7610e4fd7dfef718840a0639db2dbdd71e/robocop/checkers/misc.py#L138
Anytime the rule is configred (also with default value) the converter is called and converter return value is saved internally instead.

You could do:

def csv_string_to_set(csv_string):
    return set(csv_string.split(",") 
    
 and use it as converter

rikerfi: I guess we could stand with following solution:
```python
        if self.todos:
            pass  # avoid resolving todos for each and every comment separately.
        else:
            self.todos = self.param("todo-in-comment", "todos").split(",")
```

robocop/checkers/comments.py
```diff
@@ -7,11 +7,17 @@
...
     "0701": Rule(
+        RuleParam(
+            name="todos",
+            default="todo,fixme",
+            converter=str,
+            desc="List of terms that violate the rule in comments.",
+        ),
         rule_id="0701",
         name="todo-in-comment",
         msg="Found {{ todo_or_fixme }} in comment",
```
rikerfi, line 23: Should we refactor `todo_or_fixme` to something else? Like `todo_term` or so.

robocop/checkers/comments.py
bhirsz: What about property then? or method:
```
def __init__(self):
    self._todos = None

def todos(self):
    if not self._todos:
        self._todos = self.param("todo-in-comment", "todos").split(",")
    return self._todos
```

robocop/checkers/comments.py
rikerfi: `marker` would be great.

rikerfi:
Something hassle with WIN environment. I try to fix that by guessing the problem is with Unicode characters. Unfortunately, I don't have WIN device available anytime soon. I need to postpone the PR a few days to find one :sweat:

bhirsz:
> Something hassle with WIN environment. I try to fix that by guessing the problem is with Unicode characters. Unfortunately, I don't have WIN device available anytime soon. I need to postpone the PR a few days to find one 😓

I can try to run in - I will let you know. From the results I also see it's running some of the tests that should be disabled by default (like benchmark test). It's most likely because of the 'tox -> nox' migration that happened recently. I will fix that one too

bhirsz:
> > Something hassle with WIN environment. I try to fix that by guessing the problem is with Unicode characters. Unfortunately, I don't have WIN device available anytime soon. I need to postpone the PR a few days to find one 😓
> 
> I can try to run in - I will let you know. From the results I also see it's running some of the tests that should be disabled by default (like benchmark test). It's most likely because of the 'tox -> nox' migration that happened recently. I will fix that one too

There is problem with encoding, but it's not connected to your PR. All values passed to arguments/options are affected by it. I need to think about solution - it's not easy mix of argparse & encoding for specific platform

bhirsz:
> > > Something hassle with WIN environment. I try to fix that by guessing the problem is with Unicode characters. Unfortunately, I don't have WIN device available anytime soon. I need to postpone the PR a few days to find one 😓
> > 
> > 
> > I can try to run in - I will let you know. From the results I also see it's running some of the tests that should be disabled by default (like benchmark test). It's most likely because of the 'tox -> nox' migration that happened recently. I will fix that one too
> 
> There is problem with encoding, but it's not connected to your PR. All values passed to arguments/options are affected by it. I need to think about solution - it's not easy mix of argparse & encoding for specific platform

Another update - it might be also issue in how I read the test data. There is hope!

rikerfi:
> > > Something hassle with WIN environment. I try to fix that by guessing the problem is with Unicode characters. Unfortunately, I don't have WIN device available anytime soon. I need to postpone the PR a few days to find one sweat
> > 
> > 
> > I can try to run in - I will let you know. From the results I also see it's running some of the tests that should be disabled by default (like benchmark test). It's most likely because of the 'tox -> nox' migration that happened recently. I will fix that one too
> 
> There is problem with encoding, but it's not connected to your PR. All values passed to arguments/options are affected by it. I need to think about solution - it's not easy mix of argparse & encoding for specific platform

Ok, thanks for checking. I'll rollback the latest commit to keep PR simple.

bhirsz:
Ups, for some reason the PR auto closed itself - working on the fix.

rikerfi:
> Ups, for some reason the PR auto closed itself - working on the fix.

No worries, I could create new one.

bhirsz:
> Ups, for some reason the PR auto closed itself - working on the fix.

I can't do it easily due to missing permissions (forked repo). Can you revert this PR somehow>? If not I can investigate more. 

To get it working on windows you need to add ``encoding="utf-8"`` here:
https://github.com/MarketSquare/robotframework-robocop/blob/111b1f7610e4fd7dfef718840a0639db2dbdd71e/tests/atest/conftest.py#L49
and here:
https://github.com/MarketSquare/robotframework-robocop/blob/111b1f7610e4fd7dfef718840a0639db2dbdd71e/tests/atest/test_rule.py#L45

bhirsz:
Looks good now! I will leave few comments regarding lower/upper case, it's not strictly connected to your changed but my though in general for this rule

robocop/checkers/comments.py
```diff
@@ -7,21 +7,31 @@
...
+            converter=str,
+            desc="List of terms that violate the rule in comments.",
+        ),
         rule_id="0701",
         name="todo-in-comment",
-        msg="Found {{ todo_or_fixme }} in comment",
+        msg="Found {{ marker }} in comment",
         severity=RuleSeverity.WARNING,
         docs="""
         Example::
```
bhirsz, line 26: Maybe prepend it with sentence like: "Report occurences of the configured, case insensitive marker in the comments. By default it reports TODO and FIXME markers."

robocop/checkers/comments.py
```diff
@@ -150,19 +170,12 @@ def check_invalid_comments(self, name, node):
...
-                todo_or_fixme="TODO",
-                lineno=token.lineno,
-                col=token.col_offset + 1 + content.find("todo"),
-            )
-        if "fixme" in content:
+        for marker in [term for term in self.todos if term.lower() in content]:
             self.report(
                 "todo-in-comment",
-                todo_or_fixme="FIXME",
+                marker=marker.upper(),
```
bhirsz, line 176: Should we always display it with upper()? Maybe we could define markers as we want them to display, ie:

```
todo-in-comment:todos:TODO,FIXME,Remove me,Fix this!
```
or display it like it's displayed in the source code ('Found 'fixme' or Found 'TODO"). Just thinking aloud what would be best approach here.

robocop/checkers/comments.py
```diff
@@ -150,19 +170,12 @@ def check_invalid_comments(self, name, node):
...
         content = content.lower()
-        if "todo" in content:
-            self.report(
-                "todo-in-comment",
-                todo_or_fixme="TODO",
-                lineno=token.lineno,
-                col=token.col_offset + 1 + content.find("todo"),
-            )
-        if "fixme" in content:
+        for marker in [term for term in self.todos if term.lower() in content]:
```
bhirsz, line 173: To avoid doing term.lower() for every comment, we can do it when defining todo list:
```
self._todos = [term.lower() for term in self.param("todo-in-comment", "todos").split(",")]
```
and then
```
for marker in [term for term in self.todos if term in content]:
```

rikerfi: Good point.

robocop/checkers/comments.py
rikerfi: Agree, that guidance should be accurate as possible... We need to preserve the original content and use that when fetching the violation.

bhirsz (approved):

robocop/checkers/comments.py
```diff
@@ -7,21 +7,31 @@
...
+        RuleParam(
+            name="todos",
+            default="todo,fixme",
+            converter=str,
+            desc="List of case insensitive terms or phrases that violate the rule in comments.",
+        ),
         rule_id="0701",
         name="todo-in-comment",
-        msg="Found {{ todo_or_fixme }} in comment",
+        msg="Found a term or phrase {{ marker }} in the comments",
```
mnojek, line 23: Why not to use just:
```suggestion
        msg="Found '{{ marker }}' in the comments",
```
?
Also, it's better to add quotes here, not where the rule violation is reported.

robocop/checkers/comments.py
```diff
@@ -7,21 +7,31 @@
...
 
 from robocop.checkers import RawFileChecker, VisitorChecker
-from robocop.rules import Rule, RuleSeverity
+from robocop.rules import Rule, RuleSeverity, RuleParam
 from robocop.utils import ROBOT_VERSION
 
 rules = {
     "0701": Rule(
+        RuleParam(
+            name="todos",
```
mnojek, line 16: I would rename `todos` to `terms` or `markers` to stay consistent. Right now we have named it in many ways:
- todos
- marker
- phrases
- terms
It provides confusion.

mnojek (changes requested):

robocop/checkers/comments.py
rikerfi: Good point. Makes code more clearer.

robocop/checkers/comments.py
rikerfi: Hah... Children of many ideas and fixes... Definitely needs consolidation. `marker` or `markers` it will be :smile:

robocop/checkers/comments.py
mnojek: I can't see this text in the code. @rikerfi Can you add it?

robocop/checkers/comments.py
```diff
@@ -7,21 +7,31 @@
...
+from robocop.rules import Rule, RuleSeverity, RuleParam
 from robocop.utils import ROBOT_VERSION
 
 rules = {
     "0701": Rule(
+        RuleParam(
+            name="markers",
+            default="todo,fixme",
+            converter=str,
+            desc="List of case insensitive markers that violate the rule in comments.",
```
mnojek, line 19:
```suggestion
            desc="List of case-insensitive markers that violate the rule in comments.",
```

robocop/checkers/comments.py
```diff
@@ -7,21 +7,31 @@
...
         docs="""
         Example::
         
             # TODO: Refactor this code
-            # fixme 
+            # fixme
         
+        Configuration example::
+
+            robocop --configure "todo-in-comment:markers:todo,fixme,Remove me,Fix this!"
```
mnojek, line 33: I think the quotes are not needed here for rule configuration. Can you please verify and remove if redundant?

mnojek (changes requested):

robocop/checkers/comments.py
rikerfi: Using a phrase with spaces on it needs quotes for the configuration.
What do you think? Should we add two config examples with and without phrases? 

`robocop --configure todo-in-comment:markers:todo,fixme,TBD`
and
`robocop --configure "todo-in-comment:markers:Remove me,Fix this!"`

robocop/checkers/comments.py
rikerfi: > Maybe prepend it with sentence like: "Report occurences of the configured, case insensitive marker in the comments. By default it reports TODO and FIXME markers."

Ah, this is something we do not want to be the rule message, but the `RuleParam` description. Do you agree?

robocop/checkers/comments.py
rikerfi: See comment below...

robocop/checkers/comments.py
mnojek: Actually I was thinking to add it to `docs`, not to the `desc`. Description was fine, IMO.
@bhirsz Please advise.
@rikerfi Sorry for confusion.

robocop/checkers/comments.py
mnojek: Oh, sorry. I missed there are spaces inside. Good catch!
@bhirsz I think we should add this info to our documentation.
@rikerfi I think you can keep it like it is, no need to update it. Leave it with quotes and spaces.

robocop/checkers/comments.py
bhirsz: Yeah, it should be in the docs of the Rule (since it's general information). RuleParam desc is mostly for short information on given parameter, and it don't have to contain defalt value (it's automatically loaded to external docs anyway). It can be short like just "Marker"

mnojek: Yes, so I would revert the last change and add the mentioned sentence to the docs instead of adding it to the description.

robocop/checkers/comments.py
bhirsz: @mnojek Sure, I will also separably put this information to our docs in some common page (how to provide values with spaces).

robocop/checkers/comments.py
rikerfi: Oukei, to summarize. Desired `desc`, `msg` and `docs` are:

```python
desc="List of case insensitive markers that violate the rule in comments."
msg="Found a marker '{{ marker }}' in the comments"
docs="""
Report occurrences of the configured, case-insensitive marker in the comments. By default, it reports TODO and FIXME markers.

Example::
        
       # TODO: Refactor this code
       # fixme
        
Configuration example::

       robocop --configure "todo-in-comment:markers:todo,Remove me,Fix this!"

"""
```
I got it right? :smile:

bhirsz: Yes :)

mnojek (approved):
Great addition! Thanks, @rikerfi ! And sorry for being so nit-picky! 😉

rikerfi:
> Great addition! Thanks, @rikerfi ! And sorry for being so nit-picky! wink

Hah, no worries. Reviews are the right place for that :smile:

PR #689, opened 2022-08-30
configurable block comment for `missing-space-after-comment`

rikerfi:
Configurable block comment using regex allows more flexible commenting.

E.g. comments like
```python
    #*****
    #
    # Important topics here!
    #
    #*****
```
```python
    #***** Headers *****#
```


robocop/checkers/comments.py
```diff
@@ -129,6 +160,13 @@ def markers(self):
             self._markers = self.param("todo-in-comment", "markers").lower().split(",")
         return self._markers
 
+    @property
+    def block(self):
+        if not self._block:
+            self._block = self.param("missing-space-after-comment", "block")
```
bhirsz, line 166: Since this parameter is simple and we don't need to calculate anything (like compile regex patterns), we dont need lazy loading for it. Instead of block we can use self param directly

bhirsz:
Thanks for the PR, it looks nice! Yes, it is intentional -some people have more hashes (ie whole line of hashes like line separator). Do you think block should be regex pattern to allow more flexibility, or current solution will be enough?

rikerfi:
> Do you think block should be regex pattern to allow more flexibility, or current solution will be enough?

Regex is a good idea! I'll study the topic like could we support both block-string and regex or just one. Get back to you.

rikerfi:
The first experimental for block comment using regex. It seems. that mixing block-string and regex is not a good idea because of regex validation. I end up using regex only.

robocop/checkers/comments.py
```diff
@@ -9,6 +11,15 @@
 from robocop.checkers import RawFileChecker, VisitorChecker
 from robocop.rules import Rule, RuleSeverity, RuleParam
 from robocop.utils import ROBOT_VERSION
+from robocop.exceptions import ConfigGeneralError
+
+def regex(value):
+    converted = rf'{value}'
+    try:
+        re.compile(converted)
```
bhirsz, line 19: You're not returning compiled pattern, but converted string instead. I guess it should be:
```
try:
    return re.compile(converted)
    except re.error as regex_err:
        raise ValueError(f'Regex error: {regex_err}')
```

robocop/checkers/comments.py
```diff
@@ -38,17 +49,40 @@
         """,
     ),
     "0702": Rule(
+        RuleParam(
+            name="block",
+            default="^###",
+            converter=regex,
+            desc="Define regex for block comment.",
```
bhirsz, line 56: Rather than using verb, we should describe the parameter, ie: ``Block comment regex pattern``

robocop/checkers/comments.py
rikerfi: fixed.

robocop/checkers/comments.py
rikerfi: fixed,

mnojek (approved):
Looking good :)

Issue #688, opened 2022-08-30
Feature request: add a check ensuring the minimum spacing between keyword & arguments


Lucas-C:
Example of error currently detected by Visual Studio Intellisense:
![MicrosoftTeams-image (3)](https://user-images.githubusercontent.com/925560/187410120-27dc824b-2cf4-4704-90da-1d823c434899.png)

Could `robotframework-robocop` be able to detect this issue?

Lucas-C:
**Note:** `robot --dryrun .` does not detect this problem either

mnojek:
Hey @Lucas-C, thanks for your interest in Robocop.
Similar request as in #687 - we do not plan to implement such features. They go beyond what a "linter tool" is.
Closing as not planned.

Lucas-C:
😔😢
OK, your call
Just FYI, a linter like Pylint does exactly this kind of static code analysis

Gadzy:
I second Lucas here.
White-space consistency is one of the most basic functionalities of a linter tool. How is it "beyond" scope ?

bhirsz:
We're checking for whitespace in the same way pylint is doing and even more (I will explain why I think that way later). If user provides not enough whitespace in such case we don't have any way of recognizing if we're dealing with argument or argument name - especially given how Robot Framework is flexible and allows you to input whole sentences as keywords. 

Using pylint argument as example, this python code:
```
defmy_method
```
No linter will recognize that you actually meant:
```
def my_method
```
Something similiar happens in your code.

What can you do is to detect unrecognized keyword with other tools like sherlock - but sherlock is still in growing phase and it don't have that many useful features like Robocop Or use ``robot --dryrun`` - it's actually interesting that it didn't detect this problem because normally it should). 

And coming back to my first sentences why I think Robocop is actually going bit beyong - there are cases when we detect unrecognized syntax:
```
[Arguments] arg  # only one space here
```
If some lines starts with something that looks like Robot Framework setting but it's syntax error/ unrecognized as such, we're reporting it as not-enough-whitespace-after-* errors.

bhirsz:
If you mostly use the same libraries and you want to prevent such cases you can hovewer create custom rules. Search for keyword calls and see if their begining match keywords from your libraries. If it does but have something extra at the end - it will mean not enough whitespace. In your case keyword starts with "Click Element" which is probably from popular SeleniumLibrary but it have extra not recognized text after.

Lucas-C:
[...]

PR #690, opened 2022-08-31
New rule: missing-4-spaces-before-argument - implement #688

Lucas-C:
Given that #688 was initially rejected,
I'm going to wait for maintainers to give me their approval
before adding unit tests to this PR.

Given the following keyword definition:
```
Display dialog on click
    Click Element xpath://*[text()="Submit me"]
    Wait Until Page Contains Element css:[role="dialog"]
```

This check would raise this error on calling `robocop`:
```
test.resource:123:13 [E] 0321 There should be 4 spaces before argument(s) of keyword "Click Element" (missing-4-spaces-before-argument)
test.resource:124:24 [E] 0321 There should be 4 spaces before argument(s) of keyword "Wait Until Page Contains" (missing-4-spaces-before-argument)
```

Note that this new rule is able to detect the existence of `SeleniumLibrary` without any configuration provided, just based on its presence in the `$PYTHONPATH` / `site-packages/`

### DIFF
--- a/robocop/checkers/naming.py
+++ b/robocop/checkers/naming.py
@@ -2,12 +2,19 @@
 Naming checkers
 """
 import re
+import os.path
 from collections import defaultdict
 from pathlib import Path
 
 from robot.api import Token
+from robot.conf import RobotSettings
+from robot.output import LOGGER as OUTPUT_LOGGER, Output
 from robot.parsing.model.blocks import Keyword
 from robot.parsing.model.statements import Arguments, KeywordCall
+from robot.running.builder import ResourceFileBuilder, TestSuiteBuilder
+from robot.running.context import EXECUTION_CONTEXTS
+from robot.running.namespace import Namespace
+from robot.variables import VariableScopes
 
 from robocop.checkers import VisitorChecker
 from robocop.rules import Rule, RuleParam, RuleSeverity
@@ -289,6 +296,12 @@ rules = {
         `regex_pattern` should define regex pattern for characters not allowed in names. For example `[@\[]` pattern
         reports any occurrence of `@[` characters.
         """,
+    ),
+    "0321": Rule(
+        rule_id="0321",
+        name="missing-4-spaces-before-argument",
+        msg='There should be 4 spaces before argument(s) of keyword "{{ keyword_name }}"',
+        severity=RuleSeverity.ERROR
     ),
 }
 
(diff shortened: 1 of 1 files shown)


mnojek:
Why do you want to have 4 spaces in the first place?
RF allows for at least 2 spaces and it's up to the programmer to choose how many are going to be used across the files.

rikerfi:
> Why do you want to have 4 spaces in the first place? RF allows for at least 2 spaces and it's up to the programmer to choose how many are going to be used across the files.

It is kind of forcing coding convention... But the possible rule should be configurable where a number of spaces are defined.

Gadzy:
4 spaces is a convention and the expected number of spaces could (should ?) be configurable.

But as pointed out in #688, a single space is not currently detected as a wrong keyword/argument separation.

bhirsz:
Thanks for the contribution! Clever idea with the TestSuiteBuilder. Hovewer, coming from my experience in various projects, it will not always work (there are many edge cases, dynamic importing based on variables from other file etc). That's why it will probably fit more into external rule. However we recently mentioned in the other issue the possibility of having public repository for such rules - where users can easily add their own external rules. This could be good time to create it and put it there. Users could install such "extra", not fully tested rules with ``pip install robocop-extra-rules`` (example) and then load them with --ext-rules RobocopExtra.NameOfTheRule . 

If we would decide on creating such place, I could bootstrap everything necessary (test framework - one bit easier than the one in the Robocop, release process, Github Actions etc).

robocop/checkers/naming.py
```diff
@@ -290,6 +297,12 @@
         reports any occurrence of `@[` characters.
         """,
     ),
+    "0321": Rule(
+        rule_id="0321",
+        name="missing-4-spaces-before-argument",
```
bhirsz, line 302: If we decide on configurable number of spaces, name of the rule will be invalid. RF also supports tabs and pipes. It should be ``missing-whitespace-before-argument`` or similar.

robocop/checkers/naming.py
```diff
@@ -743,3 +756,43 @@ def check_if_keyword_is_deprecated(self, keyword_name, node):
...
+
+
+class SpacesBeforeArgumentChecker(VisitorChecker):
+    reports = ("missing-4-spaces-before-argument",)
+
+    def visit_KeywordCall(self, node):
+        if node.keyword in KNOWN_KEYWORDS:
+            return
+        try:
+            known_keyword = next(keyword for keyword in KNOWN_KEYWORDS if node.keyword.startswith(f"{keyword} "))
```
bhirsz, line 768: We need to normalize name to also catch:
```
click_element
ClickElement
```
Most likely KNOWN_KEYWORDS should contain already normalized names, then on visit_KeywordCall we normalize the keyword name and compare.

bhirsz, line 768: What about ``Click Element${not_even_single_space}``?

robocop/checkers/naming.py
```diff
@@ -743,3 +756,43 @@ def check_if_keyword_is_deprecated(self, keyword_name, node):
...
+            self.report("missing-4-spaces-before-argument", node=node, col=len(known_keyword), keyword_name=known_keyword)
+        except StopIteration:
+            pass
+
+
+def all_builtin_keywords(suite_dir='.', src_filepath='./settings.robot'):
```
bhirsz, line 774: What's the ``settings.robot``?

robocop/checkers/naming.py
```diff
@@ -743,3 +756,43 @@ def check_if_keyword_is_deprecated(self, keyword_name, node):
...
+    ns = Namespace(variables, suite, resource)
+    EXECUTION_CONTEXTS.start_suite(suite, ns, output, dry_run=True)
+    ns.handle_imports()
+    for lib in ns.libraries:
+        for keyword in lib.handlers:
+            yield keyword
+    EXECUTION_CONTEXTS.end_suite()
+
+
+KNOWN_KEYWORDS = [keyword.name for keyword in all_builtin_keywords()]
```
bhirsz, line 798: KNOWN_KEYWORDS should be initialized in checker init class, or even better it could be lazy loaded in ``visit_File`` (if the rule is disabled, it would never create this lookup saving as some time). By lazy loading I mean that it could create some kind of property:

```
def __init__(self):
    self._known_keywords = None

@property
def known_keywords(self):
    if self._known_keyword is None:
          self._known_keywords = get_known_keywords()
    return self._known_keywords
```

Also the scope changes for every file - one file could import for example SeleniumLibrary while the other not.. but it would be hard to track since the scope could be global or local.

mnojek:
@Lucas-C
Hey :) I am doing some clean-up of the issues and PRs we have in the project.

Are you still interested in finishing this PR and adjust it according to the comments?
Maybe I can be of any help here?

robocop/checkers/naming.py
mnojek: I think this is the reason why we should not include this rule into Robocop. There are many ways to dynamically import new keywords and libraries, and I am just worried that it will not be bullet-proof. Also, I think the performance of Robocop may drop, since this has quite extensive operations, in compare to the rest of Robocop rules.

What do you think @bhirsz ?

Although, this is a great candidate for a community rules, which is one of the ideas I would love to drive soon - to have a main repository of user-created rules that are helpful, but maybe not the best candidates for including them as built-in rules.

bhirsz:
Closing because of inactivity. The idea behind PR is good but it's unfortunately virtually impossible to determine if 'Keyword Call arg' is proper keyword call or it should be 'Keyword Call    arg' due to nature of the Robot Framework without loading whole project, dependencies, dynamic variables etc. Hovewer in some specific cases it could be created as external, company rule (if you know your codebase well enough you can determine if you can workaround some limitiations etc).

Such cases usually are catch by robot --dryrun with the exception of the keyword in Run Keywords

PR #708, opened 2022-09-05
Refactor handling empty lines for consecutive and in between rules

### LINKED ISSUE #611: [Bug] Rule #1005 ignores comment lines
### What happened?

An example code

```
Keyword1
     Keyword2

# Comment

Keyword3
     Keyword4
```
won't pass the rule 1005 (empty-lines-between-keywords). The received output will be as follows:
> Invalid number of empty lines between keywords (2/1)

It's down to preferences, but as a RF user I see use cases for having e.g. keyword library section separators (and multiple other use cases) that could be done with single-line comments between keywords.

### Operating System

Windows

### Robocop version

2.0.2

### LINKED ISSUE #707: [Bug] consecutive-empty-lines should ignore comments in the end of keyword/test
Following code:
```
*** Keywords ***
Keyword 1
    Step

# comment 1

    # comment 2
    # comment 3

# comment 4

Keyword 2
    Step
```

reports consecutive-empty-lines twice on line 6, and then on line 10 (counting 4 consecutive empty lines even if there are comments in between).

bhirsz:
Fixes #611 
Fixes #707


bhirsz:
Thanks to refactored I believe we finally will have it done right. The only downside is that this code will report not enough empty lines between keywords/test:
```
*** Keywords ***
Keyword
    Step

# comment
Keyword 2
    Step
```

But I think it's good enough since we're losing all weird quirks around comments and empty lines. For reminder, following report 3 empty lines between nodes before the refactor:
```
*** Keywords ***
Keyword
    Step

    # comment

# comment
Keyword 2
    Step
```

Issue #290, opened 2021-03-23
Add end line and end col information to issues


bhirsz:
Right now we're reporting issues only with line and col. To improve integration with tools like LSP we need also to report end line and col (some rules can report range where the problem appears).

bhirsz:
Keyword naming rules needs this update - for example wrong case in keyword name is reported on first column of the line only.

PR #729, opened 2022-09-15
Provide end line/col information to invalid-argument rule

bhirsz:
Relates #290

The goal is to slowly add end col information to all rules - whenever it is necessary. For that purpose I'm overwriting default issue format (with ``end_col`` argument). In the end it's likely that I will replace default issue format in our test framework with the one that has end col information and then remove all ``issue_format="end_col"`` arguments.

### DIFF
--- a/docs/releasenotes/2.6.0.rst
+++ b/docs/releasenotes/2.6.0.rst
@@ -20,3 +20,10 @@ Such headers will raise deprecation warning and eventually will not be supported
 
 - unrecognized header is now only reported by ``parsing-error`` and not by ``duplicated-setting`` (#683)
 - empty lines and standalone comments are now not counted towards keyword/test length in ``too-long-test-case`` and ``too-long-keyword`` rules (#671)
+
+## Issue end location (#290)
+
+Our test framework now supports custom format of the issue - which should help in enhancing the precision of the reported
+issues. Following rules has improved precision (reported beginning and end of the rule) in this release:
+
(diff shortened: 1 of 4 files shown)


docs/releasenotes/2.6.0.rst
```diff
@@ -20,3 +20,10 @@ Such headers will raise deprecation warning and eventually will not be supported
 
 - unrecognized header is now only reported by ``parsing-error`` and not by ``duplicated-setting`` (#683)
 - empty lines and standalone comments are now not counted towards keyword/test length in ``too-long-test-case`` and ``too-long-keyword`` rules (#671)
+
+## Issue end location (#290)
+
+Our test framework now supports custom format of the issue - which should help in enhancing the precision of the reported
```
mnojek, line 26:
```suggestion
Our test framework now supports custom format of the issue - which should help with enhancing the precision of the reported
```

docs/releasenotes/2.6.0.rst
```diff
@@ -20,3 +20,10 @@ Such headers will raise deprecation warning and eventually will not be supported
...
+issues. Following rules has improved precision (reported beginning and end of the rule) in this release:
```
mnojek, line 27:
```suggestion
issues. The following rules have improved precision (reported beginning and end of the rule) in this release:
```

mnojek (approved):

PR #747, opened 2022-10-14
update sarif config example to use colon delimiters

bollwyvl:
Thanks for robocop!

This docs-only change helps avoid unexpected copy-pasta errors like:

```
Error: Provided invalid config: 'sarif:output_dir=build/reports/sarif' (general pattern: <rule>:<param>:<value>)
```


bhirsz:
Thanks, the docs updates are always most welcome!

bhirsz (approved):

PR #749, opened 2022-10-19
Fix deprecated-statement in RF 6

bhirsz:
Also other minor fixes (mostly test data updates for RF6).

### DIFF
--- a/tests/atest/rules/errors/invalid_if/test_rule.py
+++ b/tests/atest/rules/errors/invalid_if/test_rule.py
@@ -3,7 +3,10 @@ from tests.atest.utils import RuleAcceptance
 
 class TestRuleAcceptance(RuleAcceptance):
     def test_rule(self):
-        self.check_rule(src_files=["test.robot"], expected_file="expected_output_rf5.txt", target_version=">=5")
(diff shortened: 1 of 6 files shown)


tests/test_data/override_severity/test.robot
```diff
@@ -1,6 +1,6 @@
 *** Settings ***
 Documentation    Testing Robocop
-Force Tags       comeANDgo    # Evoke rule 0602 with severity I
+Default Tags       comeANDgo    # Evoke rule 0602 with severity I
```
bhirsz, line 3: It invokes different rule now (Force Tags is deprecated in 6+) that's why I used Default Tags to trigger 0602

tests/atest/rules/errors/invalid_if/test_rule.py
```diff
@@ -3,7 +3,10 @@
 
 class TestRuleAcceptance(RuleAcceptance):
     def test_rule(self):
-        self.check_rule(src_files=["test.robot"], expected_file="expected_output_rf5.txt", target_version=">=5")
+        self.check_rule(src_files=["test.robot"], expected_file="expected_output_rf6.txt", target_version=">=6")
```
bhirsz, line 6: The error message for invalid IF is different in RF 6

PR #774, opened 2023-02-14
Argument file handling improvements

### LINKED ISSUE #693: [Bug] The --ext-rules option does properly support a relative path
### What happened?

When a relative path is used with --ext-rules it works only if robocop is invoked from a directory for which the relative path is valid. This is an issue when specifying --ext-rules in a .robocop file since in that case it does not work if, for example, you invoke robocop from a subdirectory of the directory containing the .robcop file.

As an example, the following entry in .robocop works only if robocop is invoked from the same directory that contains the .robocop file.

--ext-rules .robocop-custom.py

### Operating System

_No response_

### Robocop version

\>= 2.0.1

### LINKED ISSUE #770: Add support for nested argument files
I would like Robocop to support nested argument files similar to Robot Framework itself.

This would allow separating common settings from custom settings and, thus, minimizing the need of changing multiple files when something changes in the common settings

bhirsz:
The general consesus was to favor ``pyproject.toml`` over argument files but a lot of the users prefer argument files and some things are not possible easily in the toml (like nesting). That's why I have update our argument file handling a bit.

Allow to import argument files inside argument files: closes #770 
Support relative paths in the argument file: closes #693


mnojek (commented):
Overall, a great addition, but please add some documentation for it under [docs/user_guide.rst:25](https://github.com/MarketSquare/robotframework-robocop/blob/master/docs/user_guide.rst#L25-L25) before merging.

mnojek (approved):

PR #775, opened 2023-02-14
Filter --list/--list-configurables by status

### LINKED ISSUE #764: List only enabled rules
As rule list is quite long already and growing I'd like a way of listing only the rules I have enabled.

It is quite impossible to spot only the enabled rules from all the all rules.

bhirsz:
Closes #764 

PR based on #774


docs/releasenotes/2.8.0.rst
```diff
@@ -0,0 +1,53 @@
...
+    --ext-rules robocop/rules.py # it will now refer to config/robocop/rules.py
+
+List only enabled or disabled rules #764
+------------------------------------------
+
+Robocop can list all rules with ``--list`` option (or ``--list-configurables`` to only list rules that can be
+configured). Our every growing list of rules made the output of this command hard to read. That's why it is now
+possible to filter rules with ``enabled`` or ``disabled`` special patterns::
+
+    robocop --list enabled
```
mnojek, line 43: Having the option accepting PATTERN and also some custom words like `enabled`/`disabled` may be confusing, because one may think that we're using these words as pattern, not as special controls.

I like the idea of being able to list only enabled/disabled rules, so I thought that maybe a good idea would be to make them uppercase, to indicate them as special specifiers? For example, we can allow for:
```
robocop --list ENABLED
```
which will print all enabled rules. But running:
```
robocop --list enabled
```
would treat it as a pattern. What do you think, @bhirsz?

I think it will make it more clear how to use this option properly and will not confuse the users.

docs/releasenotes/2.8.0.rst
```diff
@@ -0,0 +1,53 @@
...
+    robocop -i *doc* -i line-too-long --list enabled
+    robocop --exclude empty-lines-between-test-cases --list disabled
+    robocop --list-configurables enabled
+
+The new feature still coexist with existing feature that allows to filter list output using pattern::
```
mnojek, line 48:
```suggestion
The new feature still coexists with existing feature that allows to filter the output using pattern::
```

docs/releasenotes/2.8.0.rst
```diff
@@ -0,0 +1,53 @@
...
+
+    robocop --list *line*
+
+However ``enabled``/``disabled`` patterns are more useful when you want to list only the rules that are enabled using
```
mnojek, line 52:
```suggestion
However, ``enabled``/``disabled`` patterns are more useful when you want to list only the rules that are enabled using
```

tests/utest/test_default_config.py
```diff
@@ -232,3 +232,15 @@ def test_override_default_config(self, path_to_test_data):
                 config = Config(from_cli=True)
             assert str(config.config_from) == str(other_config)
             assert config.include == {"0810"}
+
+
+def test_nested_argument_files(path_to_test_data):
```
mnojek, line 237: I think this is the part of another PR #774 and should be removed from here.

tests/test_data/argument_file/args_nested.txt
```diff
@@ -1,3 +1,3 @@
 --include some_rule
--A other_file.txt
+-A args_nested.txt
```
mnojek, line 2: These changes are in the scope of another PR #774 and should be removed from here.

tests/e2e/test_e2e.py
```diff
@@ -178,12 +178,12 @@ def test_argument_file_without_path(self):
             config.parse_args(["--argumentfile"])
         assert 'Argument file "" does not exist' in str(err)
 
-    def test_use_nested_argument_file(self, test_data_dir):
+    def test_use_circular_argument_file(self, test_data_dir):
```
mnojek, line 181: These changes are in the scope of another PR #774 and should be removed from here.

robocop/exceptions.py
```diff
@@ -27,9 +27,9 @@ def __init__(self, source):
         super().__init__(msg)
 
 
-class NestedArgumentFileError(RobocopFatalError):
+class CircularArgumentFileError(RobocopFatalError):
```
mnojek, line 30: These changes are in the scope of another PR #774 and should be removed from here.

robocop/config.py
```diff
@@ -82,6 +82,90 @@ def validate_regex(pattern: str) -> Pattern:
...
+    path = Path(orig_path)
+    if path.is_absolute():
+        return orig_path
+    resolved_path = config_dir / path
+    if not ensure_exists or resolved_path.exists():
+        return str(resolved_path)
+    return orig_path
+
+
+class ArgumentFileParser:
```
mnojek, line 95: These changes are in the scope of another PR #774 and should be removed from here.

docs/releasenotes/2.8.0.rst
```diff
@@ -0,0 +1,53 @@
+:orphan:
```
mnojek, line 1: We should also update `docs/cli_options.rst:52-55` file accordingly and the description of this CLI command in `robocop/config.py:229`.

docs/releasenotes/2.8.0.rst
bhirsz: I agree, it will make better distinction. I'm not too fond of using the same way of filtering by pattern and enabled/disabled but that was the best thing I could think of without introducing too much complexity. In some way "enabled/disabled" is also pattern of sort so we can use this field as "filter by anything that matches criteria". I will make it uppercase

tests/utest/test_default_config.py
bhirsz: Rebased after merging #774

tests/e2e/test_e2e.py
bhirsz: Rebased after merging #774

robocop/config.py
bhirsz: Rebased after merging #774

tests/utest/test_listing_rules.py
```diff
@@ -135,6 +135,39 @@ def test_list_disabled_rule(self, robocop_pre_load, msg_0101, msg_disabled_for_4
...
+        init_empty_checker(robocop_pre_load, msg_0102_0204, exclude=True)
+        with pytest.raises(SystemExit):
+            robocop_pre_load.list_checkers()
+        out, _ = capsys.readouterr()
+        assert (
+            out == "Rule - 0102 [E]: other-message: this is description (disabled)\n",
+            "Rule - 0204 [I]: another-message: Message with meaning 4 (disabled)\n",
+            "Altogether 1 rule with following severity:\n"
+            "    0 error rules,\n"
+            "    2 warning rule,\n"
```
mnojek, line 166: I think this will have plural form:
```suggestion
            "    2 warning rules,\n"
```

bhirsz: You're right and it should be catched by test itself - due to string formatting issues it only compare with line. I fixed the string formatting and string itself so now the assertion is correct.

mnojek (approved):

PR #794, opened 2023-03-21
Adjust rule message location

### LINKED ISSUE #290: Add end line and end col information to issues

mnojek:
This pull request completely reviews all locations of the rule violations.
Some beginnings and most of the endings of the rule reports were updated.
Logic in some rules needed to be adjusted to improve locating the correct place of the issues or to support different versions of RF.

It was tested with VS Code and Robot Framework Language Server plugin to see if the Robocop integration works properly.
It was verified with different versions of Robot Framework and Python.

Robot Framework version 6 is added to the unit tests execution.

I think it would be a good idea to add another class to store rule message location (`col`, `end_col`, `lineno`, `end_lineno`), instead of providing all of these attributes to the `BaseChecker.report()` method. Something to improve in the next releases.

Wow, right before 2nd anniversary of #290!
Closes #290


robocop/checkers/naming.py
```diff
@@ -574,8 +580,15 @@ def visit_Setup(self, node):  # noqa
 
     def visit_LibraryImport(self, node):  # noqa
         self.check_setting_name(node.data_tokens[0].value, node)
-        with_name = node.get_token(*self.ALIAS_TOKENS)
-        if with_name is None:
+        if ROBOT_VERSION.major < 6:
+            arg_nodes = node.get_tokens(Token.ARGUMENT)
+            # ignore cases where 'AS' is used to provide library alias for RF < 5
```
bhirsz, line 585: I think AS was introduced in RF 6.0, so the code for < RF 6 shouldn't care if there is AS or not?

tests/atest/rules/lengths/empty_test_setup/expected_output.txt
```diff
@@ -1 +1 @@
-test.robot:3:11 [E] 0520 Test Setup does not have any keywords
\ No newline at end of file
+test.robot:3:1 [E] 0520 Test Setup does not have any keywords
```
bhirsz, line 1: So the old one marked the whole Test Setup, and new one just the beginning? ie:

```
*** Settings  ***
Test Setup    Keyword
```

Old one marked ``Test Setup`` and new one just the beginning of the line.

tests/atest/rules/lengths/too_long_keyword/expected_output_severity_threshold.txt
```diff
@@ -1,2 +1,2 @@
-severity_threshold.robot:59:1 [W] 0501 Keyword 'Keyword' is too long (44/40)
-severity_threshold.robot:112:1 [E] 0501 Keyword 'Keyword2' is too long (51/40)
\ No newline at end of file
+severity_threshold.robot:15:1 [W] 0501 Keyword 'Keyword' is too long (44/40)
```
bhirsz, line 1: So now we're pointing at the start of the keyword, not the end? (I'm fine with the change, just checking)

tests/atest/rules/misc/empty_variable/expected_output.txt
```diff
@@ -1,14 +1,14 @@
 test.robot:3:1 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
 test.robot:4:1 [I] 0912 Use built-in variable @{EMPTY} instead of leaving variable without value or using backslash
 test.robot:5:1 [I] 0912 Use built-in variable &{EMPTY} instead of leaving variable without value or using backslash
-test.robot:7:3 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
-test.robot:20:3 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
-test.robot:22:3 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
-test.robot:25:3 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
-test.robot:30:3 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
-test.robot:33:3 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
-test.robot:35:25 [I] 0912 Use built-in variable ${EMPTY} instead of leaving variable without value or using backslash
```
bhirsz, line 10: So now we're always pointing at the start of the line rather than place where the 'empty' variable is? 

Ie
```
${EMPTY_WITH_BACKSLASH}  \
```
previous showed at "\" and now shows at "$"

tests/atest/rules/misc/inconsistent_assignment/autodetect/expected_output.txt
```diff
@@ -1 +1 @@
-autodetect${/}test.robot:30:23 [W] 0909 The assignment sign is not consistent within the file. Expected '' but got '=' instead
+autodetect${/}test.robot:30:15 [W] 0909 The assignment sign is not consistent within the file. Expected '' but got '=' instead
```
bhirsz, line 1: Where the new message shows?

The code:
```
    ${var}    ${var2}=    Keyword
```
Previous one showed at ``=`` and new one shows at the beginning of the second var?

tests/atest/rules/misc/inconsistent_assignment_in_variables/autodetect/expected_output.txt
```diff
@@ -1 +1 @@
-autodetect${/}test.robot:7:9 [W] 0910 The assignment sign is not consistent inside the variables section. Expected ' =' but got '=' instead
\ No newline at end of file
+autodetect${/}test.robot:7:1 [W] 0910 The assignment sign is not consistent inside the variables section. Expected ' =' but got '=' instead
```
bhirsz, line 1: The start of the line over the place where the assignment sign is?

tests/atest/rules/misc/keyword_after_return/expected_output.txt
```diff
@@ -1,2 +1,10 @@
 test.robot:20:5 [W] 0901 [Return] is not defined at the end of keyword. Note that [Return] does not quit from keyword but only set variables to be returned
-keyword_after_return_from.robot:21:5 [W] 0901 Keyword call after 'Return From Keyword'
\ No newline at end of file
+keyword_after_return_from.robot:21:5 [W] 0901 Keyword call after 'Return From Keyword'
+test.robot:57:9 [W] 0901 RETURN is not defined at the end of keyword
```
bhirsz, line 3: Note that RETURN can be returned whenever, it's like ``return`` in Python:
```
IF    $condition
    RETURN
Other Keyword If Not Condition
```

The reason behind this rule is that previously RF only had ``[Return]`` and ``Return From Keyword`` (which worked as current ``RETURN``). ``[Return]`` didn't return from the keyword which was common misconception - it only set the variables which will be returned at the end of keyword. That's why I added this rule so the people will see that the whole code will always execute even if they put ``[Return]`` in the middle.

tests/atest/rules/spacing/bad_indent/expected_output.txt
```diff
@@ -1,40 +1,40 @@
-bug375.robot:7:5 [W] 1008 Line is under-indented
-bug375.robot:8:5 [W] 1008 Line is under-indented
-bug375.robot:23:7 [W] 1008 Line is over-indented
-bug375.robot:33:4 [W] 1008 Line is under-indented
-comments.robot:19:3 [W] 1008 Line is over-indented
-comments.robot:23:4 [W] 1008 Line is under-indented
-comments.robot:26:5 [W] 1008 Line is over-indented
+bug375.robot:7:1 [W] 1008 Line is under-indented
```
bhirsz, line 1: Now we're showing the beginning of the line instead of the end of bad indent?

tests/atest/rules/spacing/misaligned_continuation_row/expected_output.txt
```diff
@@ -1,7 +1,7 @@
-test.robot:5:19 [W] 1015 Each next continuation line should be aligned with the previous one
-test.robot:10:8 [W] 1015 Each next continuation line should be aligned with the previous one
-test.robot:19:19 [W] 1015 Each next continuation line should be aligned with the previous one
-test.robot:38:14 [W] 1015 Each next continuation line should be aligned with the previous one
-test.robot:41:10 [W] 1015 Each next continuation line should be aligned with the previous one
-test.robot:44:13 [W] 1015 Each next continuation line should be aligned with the previous one
-test.robot:58:14 [W] 1015 Each next continuation line should be aligned with the previous one
\ No newline at end of file
+test.robot:5:4 [W] 1015 Each next continuation line should be aligned with the previous one
```
bhirsz, line 1: where it points now?

bhirsz:
Good work! I've read all and I put some comments - it doesn't mean I don't agree, I just want to clarify new rules positions sometimes.

tests/atest/rules/lengths/empty_test_setup/expected_output.txt
mnojek: It still underlines the whole `Test Setup`:

image

See: https://github.com/MarketSquare/robotframework-robocop/blob/adjust-rule-message-location/robocop/checkers/lengths.py#L583-L583

tests/atest/rules/lengths/too_long_keyword/expected_output_severity_threshold.txt
mnojek: Yes, because the `col` indicates where the underlining should start, and we want to underline the whole keyword, to make it visible for the user, not just the last character. 

In this case, the `end_col` is now what used to be the value of the `col`.

I think this:

image

looks much better than this:

image

tests/atest/rules/misc/inconsistent_assignment/autodetect/expected_output.txt
mnojek: Yes. I found that such small one-character-long underlines are easy to miss and hard to hover.
In this case, the equal sign is connected to the second var, so I decided that it would be better to underline the whole `${var2}=`. Looks nicer from my perspective and I think should be more friendly for the user. I'd be happy to see another perspective.

tests/atest/rules/misc/empty_variable/expected_output.txt
mnojek: Yes, because from the user's perspective, there is a problem with this whole variable, so pointing to a place where there is no value (it basically underlined empty space before) is hard to notice in the IDE.

robocop/checkers/naming.py
mnojek: If you open `tests/atest/rules/naming/empty_library_alias/test.robot` file in your IDE with RF 5 installed, you'll see that it reports this problem on the lines where `AS` is used, while IMO it should ignore such cases, instead of reporting an issue there.

It detects `AS` and following tokens as arguments.

See example here (RF 5 installed):

image

tests/atest/rules/misc/inconsistent_assignment_in_variables/autodetect/expected_output.txt
mnojek: Yes, because the variables in Variables section start at the beginning of the line.

image

tests/atest/rules/misc/keyword_after_return/expected_output.txt
mnojek: Well, I think I haven't changed the logic of this rule, so not sure why it happened. I will double-check.

tests/atest/rules/spacing/bad_indent/expected_output.txt
mnojek: Yes, because the indentation starts at the beginning of the line.
Note, that these changes are for the LSP in IDE, so that the issues are properly underlined in the code editor.

image

tests/atest/rules/spacing/bad_indent/strict/tmp.txt
```diff
@@ -1,85 +1,85 @@
-bug375.robot:9:24 [W] 1008 Line is over-indented
```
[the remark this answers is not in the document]
mnojek: I haven't created that file. Maybe you can help me here why it's like this.

tests/atest/rules/misc/inconsistent_assignment/autodetect/expected_output.txt
bhirsz: Ok I just noticed it underlines more than one character - this changes things. Yes it is fine. Should we update our expected files to include both start/end col and line? Because now we check only for start line and col. Or at least easier way to configure it when needed in our test framework - although we may have something in place already

tests/atest/rules/spacing/misaligned_continuation_row/expected_output.txt
mnojek: Exactly where the issue occurs: 

image

tests/atest/rules/misc/inconsistent_assignment/autodetect/expected_output.txt
bhirsz: Yeah, see for example https://github.com/MarketSquare/robotframework-robocop/blob/2720f2ed06bc6bb20a9b7e5948d74f6b34ef89f9/tests/atest/rules/errors/invalid_argument/test_rule.py . We can use "end_col" to set different format (see TestRule class for format details)

mnojek: I think it's possible to mark the end of the issue (see `tests/atest/rules/errors/invalid_argument/expected_output.txt`), but I don't know if it's needed in the output.
Start of the issue is enough for the user to locate it. End seems a bit confusing, either to me.

Does it give any additional value to the user? We also need to make sure, that it does not break the underlining in IDEs.

bhirsz: I dont mean to provide it to user, but include it on our tests - if the rule reports more than just single char (ie underline word) it could use extended format in the test so we can check if it is reporting end col/line correctly

mnojek:
> Good work! I've read all and I put some comments - it doesn't mean I don't agree, I just want to clarify new rules positions sometimes.

Thanks. Please take a look at my responses and resolve any conversations that you agree with, or add more comments if you need more info.

tests/atest/rules/misc/inconsistent_assignment/autodetect/expected_output.txt
bhirsz: Too much effort for this PR, let's now just update the locations and then we can slowly update the tests if needed in separate PRs

tests/atest/rules/misc/keyword_after_return/expected_output.txt
mnojek: Ok, I know what happened. I changed the file names.
Old `expected_output.txt` became new `expected_output_pre_rf5.txt` (because it is set to `target_version="<5"`) and the old `expected_output_rf5.txt` is now new `expected_output.txt` (because it should run with any other version of RF - `target_version=">=5"`, so previously the name was confusing).

Everything is fine here.

tests/atest/rules/spacing/bad_indent/strict/tmp.txt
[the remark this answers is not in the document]
mnojek: This file is never used. I'll remove it.

bhirsz (approved):

Issue #758, opened 2022-11-16
[Bug] bad-indent when using data-driven style tests with `[Tags]`


Zeckie:
### What happened?

When using [data-driven style](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#data-driven-style) tests, and including `[Tags]`, robocop gives errors like `foobar.robot:15:37 [W] 1008 Line is over-indented (bad-indent)`

For the robot file:
```robotframework
*** Settings ***
Test Template       Login with invalid credentials should fail

*** Test Cases ***                  USERNAME            PASSWORD
Invalid User Name                   [Tags]              foo
                                    invalid             ${VALID PASSWORD}
Invalid Password                    [Documentation]     foo
                                    [Tags]              bar
                                    ${VALID USER}       invalid
Invalid User Name and Password      [Tags]              baz
                                    invalid             invalid
Empty User Name
                                    ${EMPTY}            ${VALID PASSWORD}
Empty Password                      ${VALID USER}       ${EMPTY}
                                    [Tags]              spam  eggs
Empty User Name and Password        ${EMPTY}            ${EMPTY}


*** Keywords ***
Login With Invalid Credentials Should Fail
    [Arguments]    ${username}    ${password}
    No Operation
```

Editing the lines with errors (removing spaces until it no longer gives any indentation related errors) gives the following. It looks like it is subtracting the length of the test case name from the number of required spaces.

```robotframework
*** Settings ***
Test Template       Login with invalid credentials should fail

*** Test Cases ***                  USERNAME            PASSWORD
Invalid User Name                   [Tags]              foo
                   invalid             ${VALID PASSWORD}
Invalid Password                    [Documentation]     foo
                                    [Tags]              bar
                                    ${VALID USER}       invalid
Invalid User Name and Password      [Tags]              baz
      invalid             invalid
Empty User Name
                                    ${EMPTY}            ${VALID PASSWORD}
Empty Password                      ${VALID USER}       ${EMPTY}
                      [Tags]              spam  eggs
Empty User Name and Password        ${EMPTY}            ${EMPTY}


*** Keywords ***
Login With Invalid Credentials Should Fail
    [Arguments]    ${username}    ${password}
    No Operation
```

Related robotidy issue: https://github.com/MarketSquare/robotframework-tidy/issues/443

### Robocop version

2.6.0

PR #731, opened 2022-09-21
Merge two indent rules into bad-indent

bhirsz:
Closes #504 

Completely redesigned our rule for indentation. It lead to merging ``uneven-indent`` into ``bad-indent`` rule.

``bad-indent`` rule now covers uneven indent (which will result in the warning message) or invalid indent (resulting in
the error message). It is now also possible to use ``strict`` (default ``False``) mode to enforce indentation to be
multiply of ``indent`` (default 4) spaces.

As side effect of the changes I also fixed minor bugs related to indentation (it failed to detect some problems).

PR #797, opened 2023-03-22
Remove predefined severity for 1008 bad-indent rule

### LINKED ISSUE #793: [Bug] W1008 rule cannot be configured
Rule W1008 cannot be configured to the Error nor Info severity. Both command line and argument file configurations are failing.

Robocop version: 2.8.1

mnojek:
For some reason, we preconfigured this rule with specific severity, based on the different cases. This made it impossible to configure the rule, because the severity specified in the code always overwrote it.

Fixes #793

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -618,7 +618,6 @@ class UnevenIndentChecker(VisitorChecker):
                     self.report(
                         "bad-indent",
                         bad_indent_msg="Line is over-indented",
-                        severity=RuleSeverity.WARNING,
                         node=child,
                         col=token_col(child, Token.COMMENT),
                     )
@@ -710,7 +709,6 @@ class UnevenIndentChecker(VisitorChecker):
             self.report(
                 "bad-indent",
                 bad_indent_msg="Indent expected. Provide 2 or more spaces of indentation for statements inside block",
-                severity=RuleSeverity.ERROR,
                 node=statement,
                 col=indent + 1,
             )
@@ -724,7 +722,6 @@ class UnevenIndentChecker(VisitorChecker):
         self.report(
             "bad-indent",
             bad_indent_msg=f"Line is {over_or_under}-indented",
-            severity=RuleSeverity.WARNING,
             node=statement,
             col=indent + 1,
         )


bhirsz:
We're preconfiguring rule with severity because current ``bad-indent`` is merge of two old rules - one was warning and the second error:

image

If we remove it completely it will report over/under indentation as error (or invalid indent as warning). We would need to think how it should be done to allow reporting as two different severity depending on value without overwriting the configured values. Maybe we can reuse what we already have in place - like setting artificial SeverityThreshold for 0/1 or False/True and then provide the value:

```
SeverityThreshold("is_error", compare_method="equal")  # we dont have compare method equal
...

self.report(
...
   is_error=True
)

mnojek:
Honestly, I do not remember why we merged these 2 rules. Can you remind me?
Maybe we should keep them separate? Since the message varies, maybe it does not make sense to keep it all under one rule?

bhirsz:
I don't have idea why it was done either, and it was me who did that : https://github.com/MarketSquare/robotframework-robocop/pull/731 :D Probably it was just to not have rules that were checking for similar stuff - but in the end it is something different (invalid indent which is error or just not pretty indent which is warning)

mnojek:
@bhirsz What do you think?

bhirsz:
I agree on splitting but not sure if we should have two rules for under/over indent. Maybe just "uneven" indent rule?

mnojek:
> I agree on splitting but not sure if we should have two rules for under/over indent. Maybe just "uneven" indent rule?

Then we are getting back exactly to what we had before 🤣 

I don't have a strong opinion here, but I think that having them separately gives some additional value for the user - it is immediately known what action can fix the rule violation: adding or removing whitespaces. "uneven" just gives information that something is wrong with this line, but for different users "even" may mean different output, so it may not be clear what to do with such a line. That's why I would opt to have 2 rules for that.

bhirsz:
Ah right :D Ok, I also don't have any preference here - we can go with 3 rules then

mnojek:
Ok, then I am closing this PR and will prepare another one to solve it the way we agreed.

Also, there needs to be done additional work for these rules to also support templated test cases, as for the bug reported in #758

mnojek:
I started rewriting the rules and I realized that it doesn't make much sense.
Both `over-indent` and `under-indent` rules would need to have the `indent` parameter, which makes it redundant, because the user would need to specify the indent level separately for both of them. Also, there are these `strict` and `ignore_uneven` parameters, which also need to be duplicated.

Now I think, it makes sense to split the `bad-indent` rule into 2:
- `bad-block-indent` which will always report an error for bad indentation in block segments
- `bad-indent` (or `uneven-indent` like we had before 😅) which will check over and under indentation issues and here we can actually use parametrized message for these 2 cases and reuse the parameters in 1 rule.

@bhirsz WDYT?

mnojek:
So basically I can leave the `bad-indent` rule like it is, but add 1 new for checking the `bad-block-indent` and adjust the conditions for the `bad-indent` rule to exclude this case and handle it by the new rule.

bhirsz:
I'm fine with the changes - still better than now where we have one rule with different severities which is quite confusing

mnojek:
The work is continued in #803

PR #802, opened 2023-03-26
Remove packaging

### LINKED ISSUE #742: Could the `packaging` dependency be dropped?
From what I can see, packaging is used just for its `Version` and it checks mostly the major/minor versions, where getting the major/minor version through a split with `.` and using it as a tuple only with major/minor should suffice.

This dep brings or `pyparsing` too, so, it brings up a lot of code just to check the version...

As for the reason for me, sometimes I do have issues because a version of `robocop` is vendored in the language server, which means it's not pip-installed, but if the user has an older version of `packaging` installed in its env `robocop` doesn't really work...

bhirsz:
This PR removes packaging dependency from our repo. It's because we have a lot of problems with this dependency, and it introduces a lot of interdependencies we do need. 

The code was copied from packaging.version and them trimmed and simplified until it matched our needs. I will review it once again soon to see if we can afford to cut it more (hence draft state).

Closes #742

### DIFF
--- a/robocop/utils/version_matching.py
+++ b/robocop/utils/version_matching.py
@@ -1,0 +1,395 @@
+import collections
+import itertools
+import re
+from functools import total_ordering
+from typing import List, Optional, SupportsInt, Tuple, Union
+
+VERSION_PATTERN = r"""
+    v?
+    (?:
+        (?P<release>[0-9]+(?:\.[0-9]+)*)                  # release segment
+        (?P<pre>                                          # pre-release
+            [-_\.]?
+            (?P<pre_l>(a|b|c|rc|alpha|beta|pre|preview))
+            [-_\.]?
+            (?P<pre_n>[0-9]+)?
+        )?
+        (?P<dev>                                          # dev release
+            [-_\.]?
+            (?P<dev_l>dev)
+            [-_\.]?
+            (?P<dev_n>[0-9]+)?
+        )?
+    )
+    (?:\+(?P<local>[a-z0-9]+(?:[-_\.][a-z0-9]+)*))?       # local version
+"""
+
+
+class InfinityType:
+    def __lt__(self, other: object) -> bool:
+        return False
+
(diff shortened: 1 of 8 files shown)


tests/utest/test_utils.py
```diff
@@ -190,3 +191,68 @@ def test_invalid_pattern_type(self):
...
         assert r"Invalid regex pattern: bad escape \\9 at position 1" in str(error)
+
+
+@pytest.mark.parametrize(
+    "robot_version, rule_version, should_pass",
+    [
+        ("6.0.2", ">=6", True),
+        ("6.1a2.dev1", ">=6", True),
+        ("6.0rc2.dev1", ">=6", False),
+        ("6.0pre2.dev1", ">=6", False),
```
mnojek, line 202: Shouldn't these be `True`?

robocop/utils/version_matching.py
mnojek, line 1: Well, it looks like there is definitely a big niche related to versioning systems on the market.

mnojek (commented):
Nice work!

tests/utest/test_utils.py
bhirsz: On first look I also thought that, but its actually realase candidate of the 6.0 version (so its still not 6 oficially). We dont want to enable/disable rules for version that is yet to be oficially released - after all Robot Framework for example automatically switch to dev version right after the release

mnojek: It's good to better understand how version specifiers work. Thanks for clarifying.

mnojek:
Do you want to include this in the 3.0.0 release?
I think I am done with any further changes.

bhirsz:
> Do you want to include this in the 3.0.0 release?
> I think I am done with any further changes.

Not sure. My changes are actually ready, I just need to make release notes for it. Maybe including it in 3.0 would be best since removing dependency is always risky. I will update PR today

PR #803, opened 2023-03-26
Refactor bad indent rules

### LINKED ISSUE #793: [Bug] W1008 rule cannot be configured

mnojek:
- Added new E1017 `bad-block-indent` rule to check for errors in indentation inside code block (FOR loops, TRY-EXCEPT etc.)
- Removed `ignore_uneven` parameter for W1008 `bad-indent` rule, since it was related to the case for indentation inside block, now moved to another rule
- Remove `strict` parameter, since it was only used together with `indent` parameter
- `indent` parameter now has a default value of `-1` and is ignored, until it's explicitly configured
- Added new handling for deprecated rules
  - fixed a case when the rule id was provided in the CLI instead of the rule name (the former never actually triggered the deprecation warning)
  - it is now able to create different deprecation messages: renamed rules and deprecated rules
    - Deprecated rules only show a deprecation warning
    - Renamed rules trigger deprecation warning and also are dynamically exchanged for new rules
  - We can think of some E2E test to check if deprecation warning appears, I don't have idea now how to implement it. 
  - Added a function that removes deprecation warnings from the output of the tests

The PR is in a draft because 3 tests fail when run with `nox`, but if I run the same Robocop command over the same files with the same configuration, the output is fine. I spent some time debugging the problem and I can't see it. I need to take some rest from the code, because I'm going circles.

@bhirsz If you will have some time, I'd appreciate help here. 
I will take one more try with it and if I will fail, I think I'm going to release what's already been done and leave this for later.

Fixes #793


robocop/checkers/spacing.py
```diff
@@ -75,24 +75,12 @@
         severity=RuleSeverity.WARNING,
     ),
     "1008": Rule(
-        RuleParam(
```
bhirsz, line 78: If we're going to remove parameters we will need to make next release major release (because people configuration will break). We could also introduce deprecation period but it would require extra work, may not be worth it

bhirsz:
I also don't know why those 3 tests fail - investigating. But it only happens if I run all tests together (running single test is fine). I suspect something is up with how pytest is clearing out the data between tests. Maybe we're reusing some fixture etc from other tests?

robocop/checkers/spacing.py
bhirsz: After some debugging, for example if I run tests on all_passing.robot file in suite with all tests, bad-indent will be configured to 3 spaces. Somehow pytest remebers the last rule configuration and uses it in different test (even if we're using totally new Robocop instance)

bhirsz: Updating last test in bad-indent fix 2 failing tests:

```
    @pytest.mark.parametrize("file_suffix, target_version", [("", ">=5"), ("_rf4", "==4.*"), ("_rf3", "==3.*")])
    def test_rule(self, file_suffix, target_version):
        self.check_rule(config="-c bad-indent:indent:4", expected_file=f"expected_output{file_suffix}.txt", target_version=target_version)

```
(we override previous -c with indent of 3 spaces)

bhirsz:
After some debugging, for example if I run tests on all_passing.robot file in suite with all tests, bad-indent will be configured to 3 spaces. Somehow pytest remebers the last rule configuration and uses it in different test (even if we're using totally new Robocop instance)

bhirsz:
Updating last test in bad-indent fix 2 failing tests:

```
    @pytest.mark.parametrize("file_suffix, target_version", [("", ">=5"), ("_rf4", "==4.*"), ("_rf3", "==3.*")])
    def test_rule(self, file_suffix, target_version):
        self.check_rule(config="-c bad-indent:indent:4", expected_file=f"expected_output{file_suffix}.txt", target_version=target_version)

```
(we override previous -c with indent of 3 spaces)

robocop/checkers/spacing.py
mnojek: So we're talking about Robocop 3.0 here?

mnojek: Because I did not remove this rule, just scoped down its functionalities and moved to another rule. Previously, in such cases, we were bumping up the second number (Y) in X.Y.Z. versioning.

mnojek:
> Updating last test in bad-indent fix 2 failing tests:
> 
> ```
>     @pytest.mark.parametrize("file_suffix, target_version", [("", ">=5"), ("_rf4", "==4.*"), ("_rf3", "==3.*")])
>     def test_rule(self, file_suffix, target_version):
>         self.check_rule(config="-c bad-indent:indent:4", expected_file=f"expected_output{file_suffix}.txt", target_version=target_version)
> ```
> 
> (we override previous -c with indent of 3 spaces)

Thanks for this clue! That helped me fix it. I run this test with the default value for indent, which is `-1`, but I don't know why I even need to configure it. Anyway, doing this:

```
    @pytest.mark.parametrize("file_suffix, target_version", [("", ">=5"), ("_rf4", "==4.*"), ("_rf3", "==3.*")])
    def test_rule(self, file_suffix, target_version):
        self.check_rule(
            config="-c bad-indent:indent:-1",
            expected_file=f"expected_output{file_suffix}.txt",
            target_version=target_version,
        )
```
Make the tests pass. Now I want to understand why Pytest runs Robocop instance with previous configuration...

And I think I got it. Different tests are in the same class, so the `check_rule`, even if called several times, already have some values defined. If I do not provide some default parameter, which was overwritten by the previous tests, it reuses it. Not sure exactly why, but seems that this is how it works.

Interesting thing is that, this problem didn't hit us before...

bhirsz:
> > Updating last test in bad-indent fix 2 failing tests:
> > 
> > ```
> >     @pytest.mark.parametrize("file_suffix, target_version", [("", ">=5"), ("_rf4", "==4.*"), ("_rf3", "==3.*")])
> >     def test_rule(self, file_suffix, target_version):
> >         self.check_rule(config="-c bad-indent:indent:4", expected_file=f"expected_output{file_suffix}.txt", target_version=target_version)
> > ```
> > 
> > (we override previous -c with indent of 3 spaces)
> 
> Thanks for this clue! That helped me fix it. I run this test with the default value for indent, which is `-1`, but I don't know why I even need to configure it. Anyway, doing this:
> 
> ```
>     @pytest.mark.parametrize("file_suffix, target_version", [("", ">=5"), ("_rf4", "==4.*"), ("_rf3", "==3.*")])
>     def test_rule(self, file_suffix, target_version):
>         self.check_rule(
>             config="-c bad-indent:indent:-1",
>             expected_file=f"expected_output{file_suffix}.txt",
>             target_version=target_version,
>         )
> ```
> Make the tests pass. Now I want to understand why Pytest runs Robocop instance with previous configuration...
> 
> And I think I got it. Different tests are in the same class, so the `check_rule`, even if called several times, already have some values defined. If I do not provide some default parameter, which was overwritten by the previous tests, it reuses it. Not sure exactly why, but seems that this is how it works.
> 
> Interesting thing is that, this problem didn't hit us before...

Good investigation too, I was at loss why it happens. Previously we dynamically generated tests so it worked differently (hovewer I kind of remember similar issue long time ago, which I think we also workarounded by using the default configuration ). For me it is pretty much the bug in the pytest. Never every, even with the same test name, we should use the same context as the other test (if we dont use the same fixture with the same scope)

robocop/checkers/spacing.py
bhirsz: Yes, Robocop 3.0. It is because if user have configuration that works for 2.8.0, it will result in exception (missing parameter) in 2.9.0 - we don't want that. Ie:

robocop -c bad-indent:strict=True . 

Will not work anymore

bhirsz: Optionally you can leave the parameter, but don't use them and after deprecation period remove them

bhirsz (approved):
I'm fine with the change - the only open topic is the next version number.

robocop/checkers/spacing.py
mnojek: Anyway, I think that even if I leave the parameter, we still should release 3.0.0, because of the changes to the columns for many rules. If someone used disablers, they may stop working, being misplaced now due to new locations, right?

mnojek:
> For me it is pretty much the bug in the pytest. Never every, even with the same test name, we should use the same context as the other test (if we dont use the same fixture with the same scope)

I agree. Maybe I will find some time to dig into it more and report it. For now, we at least fixed it on our side.

robocop/checkers/spacing.py
bhirsz: Yes

mnojek: Then we need to go with 3.0.0.

Should we add the parameters back for backward compatibility but just deprecate them and maybe remove in the next version? I think better not to break the day of the users 😉 As you said, we can add them and just don't call anywhere.

bhirsz: I agree, some deprecation period would be graceful for users

mnojek:
@bhirsz Please resubmit your review and I will merge it when it's done.

bhirsz (approved):

PR #805, opened 2023-03-27
Support DDT for bad indent

### LINKED ISSUE #758: [Bug] bad-indent when using data-driven style tests with `[Tags]`

mnojek:
- Added support for templated suite for `bad-indent` rule
- Added tests for the `bad-indent` rule
- Removed mentions of the old rules from docs
- Added docstrings to mysterious functions that I debugged
- Reformatted release notes and added more info

Please take a look at line 547 in `robocop/checkers/spacing.py` file below. This is what handles templated suites properly, but I'm not entirely sure if this is a desired implementation. I added a comment here, so maybe it can stay like this.

Fixes #758

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -517,6 +517,11 @@ class InconsistentUseOfTabsAndSpacesChecker(VisitorChecker, ModelVisitor):
 
 
 def get_indent(node):
+    """Calculate the indentation length for given node
+
+    Returns:
+        int: Indentation length
+    """
     tokens = node.tokens if hasattr(node, "tokens") else node.header.tokens
     indent_len = 0
     for token in tokens:
@@ -527,17 +532,34 @@ def get_indent(node):
 
 
 def count_indents(node):
+    """Counts number of occurrences for unique indent values
+
+    Returns:
+        Counter: A counter of unique indent values with associated number of occurrences in given node
+    """
     indents = Counter()
     if node is None:
         return indents
     for line in node.body:
         if isinstance(line, (EmptyLine, Comment)):
             continue
-        indents[(get_indent(line))] += 1
+        # for templated suite, there can be data on the same line where the test case name is
+        if node.lineno == line.lineno and isinstance(node, TestCase):
+            indents[len(node.name) + (get_indent(line))] += 1
+        else:
+            indents[(get_indent(line))] += 1
     return indents
 
 
 def most_common_indent(indents):
+    """Returns most commonly occurred indent
+
+    Args:
+        indents (Counter): A counter of unique indent values with associated number of occurrences in given node
+
+    Returns:
+        indent (int): Most common indent or the first one
+    """
     common_indents = indents.most_common(1)
     if not common_indents:
         return 0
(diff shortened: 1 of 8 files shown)


robocop/checkers/spacing.py
```diff
@@ -517,6 +517,11 @@ def visit_Statement(self, node):  # noqa
 
 
 def get_indent(node):
+    """Calculate the indentation length for given node
+
+    Returns:
+        int: Indentation length
```
bhirsz, line 523: Maybe we should think about increasing the min required python version of Robocop? That way we could start using typing

bhirsz (approved):

robocop/checkers/spacing.py
mnojek: I was thinking about the same thing exactly when adding these docstrings :D

I would love to have typing, but that would mean that we need to update Python to 3.9 and drop older versions.
I am wondering how this will affect our users.

bhirsz: Robotidy uses 3.7 and it already supports typing: https://docs.python.org/3.7/library/typing.html . 3.8 does not introduce anything revolutionary to typing, and 3.9 replaces ``Dict`` or ``List`` types by direct type (``dict``, ``list`` etc). So it would be enough to bump one major version - 3.6 is already at the end of life:

image

mnojek: It's good to actually follow the versions of the dependencies along with their support, and Python version is one of them. I will create another PR for that

Since 3.6 is already dead for over a year, I think we should definitely bump the Python up to 3.7, even for security reasons. 

3.7 is also at the end of its life, so we need to keep an eye on that and update Python versions accordingly. 
I think we can require 3.8 in the second half of the year.

bhirsz (dismissed):

bhirsz (approved):

PR #813, opened 2023-04-02
Ignore nested variables in rules 0309 and 0310

### LINKED ISSUE #678: [enhancement] configure `non-local-variables-should-be-uppercase` to allow lower-case variables
It should be possible to configure `non-local-variables-should-be-uppercase` to allow lower-case variables in it.

```python
*** Variables ***
@{PROTOCOLS}  TCP  UDP


*** Test Cases ***
Test
    [Documentation]  Doc.
    KW


*** Keywords ***
KW
    [Documentation]  non-local-variables-should-be-uppercase
    FOR  ${proto}  IN  @{PROTOCOLS}
        ${stuff}  Get Stuff  ${proto}
        # This is good, but confusing:
        Set Test Variable  ${EXAMPLE_${PROTO}}  ${stuff}
        # This could be allowed via configure:
        Set Test Variable  ${EXAMPLE_${proto}}  ${stuff}
    END
```

mnojek:
Rules W0310 `non-local-variables-should-be-uppercase` and W0309 `section-variable-not-uppercase`
were previously reporting when the variable had another nested variable with lowercase name,
e.g. `${EXAMPLE_${lowercase}}`.
Now, the nested variables are ignored and if the rest of the name is uppercase, the rules
will not report the issue anymore.

Closes #678

### DIFF
--- a/docs/releasenotes/3.1.0.rst
+++ b/docs/releasenotes/3.1.0.rst
@@ -1,0 +1,56 @@
+:orphan:
+
+=============
+Robocop 3.1.0
+=============
+
+<Release description>
+
+You can install the latest available version by running
+
+::
+
+    pip install --upgrade robotframework-robocop
+
+or to install exactly this version
+
+::
+
+    pip install robotframework-robocop==3.1.0
+
+.. contents::
+   :depth: 2
+   :local:
+
+Most important changes
+======================
+
+Change title (#0)
+-----------------------------------------------
+
+Description
+
+Rule changes
+============
+
+Rules W0310 and W0309 now ignore nested variables (#678)
+------------------------------------------------------------------------------------------
+
+Rules W0310 ``non-local-variables-should-be-uppercase`` and W0309 ``section-variable-not-uppercase``
+were previously reporting when the variable had another nested variable with lowercase name,
+e.g. `${EXAMPLE_${lowercase}}`.
+Now, the nested variables are ignored and if the rest of the name is uppercase, the rules
+will not report the issue anymore.
+
+Other features
+==============
+
+Feature title (#0)
+--------------------------------
+
+Description
+
+Acknowledgements
+================
+
+Thanks to...
--- a/docs/releasenotes/3.1.0.rst
+++ b/docs/releasenotes/3.1.0.rst
@@ -33,14 +33,18 @@ Description
 Rule changes
 ============
 
-Rules W0310 and W0309 now ignore nested variables (#678)
+Rules W0310 and W0309 now handle nested variables (#678)
 ------------------------------------------------------------------------------------------
 
 Rules W0310 ``non-local-variables-should-be-uppercase`` and W0309 ``section-variable-not-uppercase``
 were previously reporting when the variable had another nested variable with lowercase name,
 e.g. `${EXAMPLE_${lowercase}}`.
-Now, the nested variables are ignored and if the rest of the name is uppercase, the rules
+Now, the nested variable names passed as an argument to one of the keywords `Set Test Variable`,
+`Set Suite Variable` or `Set Global Variable` are ignored and if the rest of the name is uppercase, the rules
 will not report the issue anymore.
+For variables in Variables section, the name still needs to be all uppercase (including
+nested variable), because all nested variables in this section need to be global anyway.
+
 
 Other features
 ==============
--- a/robocop/checkers/naming.py
+++ b/robocop/checkers/naming.py
@@ -666,25 +666,24 @@ class VariableNamingChecker(VisitorChecker):
         }
         super().__init__()
 
-    def visit_VariableSection(self, node):  # noqa
-        for child in node.body:
-            if not child.data_tokens:
-                continue
-            token = child.data_tokens[0]
-            if token.type != Token.VARIABLE or not token.value:
-                continue
-            var_name = search_variable(token.value).base
-            if var_name is None:
-                continue  # in RF<=5, a continuation mark ` ...` is wrongly considered a variable
-            normalized_var_name = remove_nested_variables(var_name)
-            if not normalized_var_name.isupper():
-                self.report(
-                    "section-variable-not-uppercase",
-                    variable_name=token.value.strip(),
-                    lineno=token.lineno,
-                    col=token.col_offset + 1,
-                    end_col=token.col_offset + len(token.value) + 1,
-                )
+    def visit_Variable(self, node):  # noqa
+        if not node.data_tokens:
+            return
+        token = node.data_tokens[0]
+        if not token.value:
+            return
+        var_name = search_variable(token.value).base
+        if var_name is None:
+            return  # in RF<=5, a continuation mark ` ...` is wrongly considered a variable
+        normalized_var_name = remove_nested_variables(var_name)
+        if not normalized_var_name.isupper():
+            self.report(
+                "section-variable-not-uppercase",
+                variable_name=token.value.strip(),
+                lineno=token.lineno,
+                col=token.col_offset + 1,
+                end_col=token.col_offset + len(token.value) + 1,
+            )
 
     def visit_KeywordCall(self, node):  # noqa
         for token in node.get_tokens(Token.ASSIGN):
@@ -703,7 +702,7 @@ class VariableNamingChecker(VisitorChecker):
             if len(node.data_tokens) < 2:
                 return
             token = node.data_tokens[1]
-            if token.type != Token.ARGUMENT or not token.value:
+            if not token.value:
                 return
             var_name = search_variable(token.value).base
             normalized_var_name = remove_nested_variables(var_name)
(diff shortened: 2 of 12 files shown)


robocop/checkers/naming.py
```diff
@@ -653,7 +671,9 @@ def visit_VariableSection(self, node):  # noqa
             if not child.data_tokens:
                 continue
             token = child.data_tokens[0]
-            if token.type == Token.VARIABLE and token.value and not token.value.isupper():
+            var_name = search_variable(token.value).base
+            normalized_var_name = remove_nested_variables(var_name)
+            if token.type == Token.VARIABLE and token.value and not normalized_var_name.isupper():
```
bhirsz, line 676: Do we need to assign var_name at all if token.type != Token.VARIABLE?

robocop/checkers/naming.py
```diff
@@ -679,7 +699,9 @@ def visit_KeywordCall(self, node):  # noqa
             if len(node.data_tokens) < 2:
                 return
             token = node.data_tokens[1]
-            if token.type == Token.ARGUMENT and not token.value.isupper():
+            var_name = search_variable(token.value).base
+            normalized_var_name = remove_nested_variables(var_name)
+            if token.type == Token.ARGUMENT and not normalized_var_name.isupper():
```
bhirsz, line 704: Same as previous comment - if token.type != Token.ARGUMENT, we will not need var_name and normalized_var_name . In that case we could return early

bhirsz:
Create note in https://github.com/MarketSquare/robotframework-robocop/tree/master/docs/releasenotes to not lose the changes.

robocop/checkers/naming.py
mnojek: You're right, I fixed that. Is it better now?

mnojek: Wait a sec, I broke something.

mnojek: Well, I found a bug in Robot Framework, so this PR needs to wait.

mnojek:
So in RF v3.2.2, all these variables written with red text are not recognized as variables in Variables section. Is it expected?
image

bhirsz:
> So in RF v3.2.2, all these variables written with red text are not recognized as variables in Variables section. Is it expected?
> image
> 

Those are not proper variables - see if they have some errors in node? Variable needs to  be left aligned and do not contain aby nested variables. Variable value can contain nested variables. So:

${VAR} ${nested${var}} is allowed

But

${nested${var}} is not

robocop/checkers/naming.py
```diff
@@ -653,10 +671,16 @@ def visit_VariableSection(self, node):  # noqa
             if not child.data_tokens:
                 continue
             token = child.data_tokens[0]
-            if token.type == Token.VARIABLE and token.value and not token.value.isupper():
+            if token.type != Token.VARIABLE or not token.value:
+                continue
```
bhirsz, line 675: It says it is not covered by tests, because youre iterating over every child in variables section and then taking first data token - the first data token for statement in variables section is always Token.Variable so this if condition is always false.

Also, you can check variable directly by using visit_Variable and then getting first data token

robocop/checkers/naming.py
```diff
@@ -679,7 +703,11 @@ def visit_KeywordCall(self, node):  # noqa
             if len(node.data_tokens) < 2:
                 return
             token = node.data_tokens[1]
-            if token.type == Token.ARGUMENT and not token.value.isupper():
+            if token.type != Token.ARGUMENT or not token.value:
```
bhirsz, line 706: Second data token of set variable keyword will be always token.Argument so this condition is always false

mnojek:
> Those are not proper variables - see if they have some errors in node? Variable needs to be left aligned and do not contain aby nested variables. Variable value can contain nested variables. So:
> 
> ${VAR} ${nested${var}} is allowed
> 
> But
> 
> ${nested${var}} value is not

The nodes have the error `Invalid variable name (...)`, but when I run this using `robot`, it works fine.

Can you point me to the place in User Guide where it says that such nested variables are not allowed? I couldn't find it...

robocop/checkers/naming.py
mnojek: I refactored it to `visit_Variable`. Is it fine now?

robocop/checkers/naming.py
mnojek: Please take a look now.

bhirsz:
> > Those are not proper variables - see if they have some errors in node? Variable needs to be left aligned and do not contain aby nested variables. Variable value can contain nested variables. So:
> > ${VAR} ${nested${var}} is allowed
> > But
> > ${nested${var}} value is not
> 
> The nodes have the error `Invalid variable name (...)`, but when I run this using `robot`, it works fine.
> 
> Can you point me to the place in User Guide where it says that such nested variables are not allowed? I couldn't find it...

I had the same problem when working on ``RenameVariables`` in Robotidy and I have decided to just ignore nodes with errors - errors in variables are already handled by Robocop (just remember to use get_errors defined somewhere in Robocop repo, because RF3 have different logic for storing the errors). 

The only mention in the user guide is:

(under [Variable section](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#toc-entry-311))
```
Their main disadvantages are that values are always strings and they cannot be created dynamically
```

mnojek:
@bhirsz Take a look at my example here:
https://github.com/robotframework/robotframework/issues/4716

bhirsz:
> @bhirsz Take a look at my example here: [robotframework/robotframework#4716](https://github.com/robotframework/robotframework/issues/4716)
> 
> EDIT: If you run your code with `robot --variable value:1 test.robot`, it will work.

Hm, isn't it because "variable needs to exist to be used as part of the other variable name in ``*** Variables ***``"? And ``--variable`` set variable globally so it does exist before parsing this file. Similar effect would be running the code with some set global variable / import variables. Tough subject but in our case I would say that everything in the variable name in  ``*** Variables ***``" needs to be upper case because we are 100% sure all variables needs to be global. So we can't ignore nested variables for such case:

```
*** Variables ***
${NAME}    ${VALUE}
${NAME_${SUBNAME}  value  # all those variables needs to be global, so they need to be upper case anyway

*** Keywords ***
    Keyword    ${UPPER_${lower}}  # here we cannot say, so it's safer to ignore nested
```

mnojek:
> > @bhirsz Take a look at my example here: [robotframework/robotframework#4716](https://github.com/robotframework/robotframework/issues/4716)
> > EDIT: If you run your code with `robot --variable value:1 test.robot`, it will work.
> 
> Hm, isn't it because "variable needs to exist to be used as part of the other variable name in `*** Variables ***`"? And `--variable` set variable globally so it does exist before parsing this file. Similar effect would be running the code with some set global variable / import variables. Tough subject but in our case I would say that everything in the variable name in `*** Variables ***`" needs to be upper case because we are 100% sure all variables needs to be global. So we can't ignore nested variables for such case:
> 
> ```
> *** Variables ***
> ${NAME}    ${VALUE}
> ${NAME_${SUBNAME}  value  # all those variables needs to be global, so they need to be upper case anyway
> 
> *** Keywords ***
>     Keyword    ${UPPER_${lower}}  # here we cannot say, so it's safer to ignore nested
> ```

Well, that makes sense! I will adjust the code.

I am still curious how RF should handle it. Let's see what Pekka responds.

mnojek:
@bhirsz The PR is ready, please review when you have time.

docs/releasenotes/3.1.0.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Description
+
+Rule changes
+============
+
+Rules W0310 and W0309 now ignore nested variables (#678)
+------------------------------------------------------------------------------------------
+
+Rules W0310 ``non-local-variables-should-be-uppercase`` and W0309 ``section-variable-not-uppercase``
```
bhirsz, line 39: Does this statement is true after recent changes to  ``section-variable-not-uppercase`` rule? Ie we don't ignore nested variables in variable name, just in the variable value:

```
*** Variables ***
${VAR_${VAR2}}    ${VAR_${var2}}  # should pass
```

robocop/checkers/naming.py
```diff
@@ -679,7 +703,13 @@ def visit_KeywordCall(self, node):  # noqa
             if len(node.data_tokens) < 2:
                 return
             token = node.data_tokens[1]
-            if token.type == Token.ARGUMENT and not token.value.isupper():
+            if not token.value:
```
bhirsz, line 706: It says the code is not covered by tests. It's possible if you add something like below code to tests:

```
Set Global Variable
...
...    previous arg is empty now
```

docs/releasenotes/3.1.0.rst
mnojek: Thanks for noticing, updated.

robocop/checkers/naming.py
```diff
@@ -648,19 +666,25 @@ def __init__(self):
...
-            if token.type == Token.VARIABLE and token.value and not token.value.isupper():
-                self.report(
-                    "section-variable-not-uppercase",
-                    variable_name=token.value,
-                    lineno=token.lineno,
-                    col=token.col_offset + 1,
-                    end_col=token.col_offset + len(token.value) + 1,
-                )
+    def visit_Variable(self, node):  # noqa
+        if not node.data_tokens:
```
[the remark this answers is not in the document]
mnojek: I skipped and I think it should work, thanks.

bhirsz (approved):

PR #815, opened 2023-04-03
Add new multiline-inline-if rule

### LINKED ISSUE #804: New rule to detect multiline inline IFs
I will create an issue for the new rule to detect multiline inline IFs.
I know they are syntactically correct, but they should be avoided at all cost. Not everything what's allowed, is good :)

And I agree that we need to suppress the `misaligned-continuation-row` rule for such cases.

_Originally posted by @mnojek in https://github.com/MarketSquare/robotframework-robocop/issues/798#issuecomment-1484683651_

mnojek:
Closes #804 

I'll be honest - I don't know how to suppress the `misaligned-continuation` rule, and I don't know if it's needed here.


bhirsz (approved):
You can suppress rule in its source - so the checker with ``misaligned-continuation`` needs to have check for inline-if and return early in such case:

```
def visit_If(self, node):
    if ROBOT_VERSION.major >= 5 and self.is_if_inline(node):
        return
```

bhirsz:
And I also think it should be WARNING severity. This looks like a abuse of the syntax after all

mnojek:
> You can suppress rule in its source - so the checker with `misaligned-continuation` needs to have check for inline-if and return early in such case:
> 
> ```
> def visit_If(self, node):
>     if ROBOT_VERSION.major >= 5 and self.is_if_inline(node):
>         return
> ```

Oh, it's that easy. Thanks :)

PR #816, opened 2023-04-06
Fix starting position for multiline-inline-if rule

mnojek:


bhirsz (approved):
Small but still it's worth to add to release notes (ie. title of this PR will suffice)

mnojek:
> Small but still it's worth to add to release notes (ie. title of this PR will suffice)

I was expecting exactly this comment from you :D But on the other hand, it's fixed and it was not yet released, so is it worth adding it to the release notes?

bhirsz:
:) Ah, I didnt realize it wasn't released yet. In that case we dont need to update release notes.

PR #820, opened 2023-04-15
Add empty-template and empty-test-template rules

### LINKED ISSUE #443: [Rule] Empty Test Template
Empty test template can be used to overwrite suite test template. But if a) there is no test template in suite settigns b) we're using [Template] without more explicit NONE value we should report it.

bhirsz:
Closes #443 

For ``empty-template`` I have decided to not detect if there is ``Test Template`` set in ``*** Settings ***``. Any empty ``[Template]`` will report the same message (that there is empty setting and if you want to overwrite Test Template then use NONE).

### DIFF
--- a/docs/releasenotes/3.1.0.rst
+++ b/docs/releasenotes/3.1.0.rst
@@ -81,6 +81,48 @@ New rules that report empty Template and Test Template (#443)
 Added new W0530 ``empty-template`` rule that reports ``[Template]`` setting without value.
 There is also new E0529 ``empty-test-template`` that reports empty ``Test Template``.
 
+Fixes
+=====
+
+Fixes for rule W1015 `misaligned-continuation-row` (#821, #818)
+---------------------------------------------------------------
+
+Fixed a case (#821) where rule W1015 `misaligned-continuation-row` was failing due to the
+continuation of the keyword starting with a comment at the beginning of the line::
+
+    *** Test Cases ***
+    Robocop Crash Example
+        Run Keywords
+        ...    Log Message     1   AND
+    # ...    Log Message     2   AND
+        ...    Log Message     3
+
+Another bug (#818) was reported for the case when there was a properly indented comment
+in between the statements, like this::
+
+    *** Test Cases ***
+    My Keyword
+        ${arf}  Create List   1  2  3
+        # ...  4
+        ...  5
+
+Which cause the rule to be incorrectly reported in the last line of the example above
+even though it's aligned properly. It is now fixed and the rule does not report in such case.
+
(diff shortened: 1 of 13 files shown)


docs/releasenotes/3.1.0.rst
```diff
@@ -75,6 +75,12 @@ New W0918 ``multiline-inline-if`` rule that detects when inline IF is split into
 multiple lines. It's allowed, but it should be avoided, since it decreases readability.
 It recommends to use normal IF/ELSE instead.
 
+New rules that reports empty Template and Test Template (#443)
```
mnojek, line 78:
```suggestion
New rules that report empty Template and Test Template (#443)
```

docs/releasenotes/3.1.0.rst
```diff
@@ -75,6 +75,12 @@ New W0918 ``multiline-inline-if`` rule that detects when inline IF is split into
...
+--------------------------------------------------------------
+
+Added new W0530 ``empty-template`` rule that reports ``[Template]`` setting without value.
+There is also new E0539 ``empty-test-template`` that reports empty ``Test Template``.
```
mnojek, line 82:
```suggestion
There is also new E0529 ``empty-test-template`` that reports empty ``Test Template``.
```

mnojek: Maybe let's also explain here why one has a warning level, and another one has error level of severity?

mnojek (approved):

docs/releasenotes/3.1.0.rst
bhirsz: Yeah, I will do it

bhirsz:
@mnojek I have added docs to ``empty-test-template`` to explain why it is error severity

robocop/checkers/lengths.py
```diff
@@ -227,11 +227,22 @@
...
     ),
-    "0529": Rule(rule_id="0529", name="empty-test-template", msg="Test Template is empty", severity=RuleSeverity.ERROR),
+    "0529": Rule(
+        rule_id="0529",
+        name="empty-test-template",
+        msg="Test Template is empty",
+        docs="""
+        The ``Test Template`` set template to all tests in the suite. Empty value is considered an error
+        because it leads users to wrong impression on how the suite operates.
+        Without value, the setting is ignored and tests are not templated.
```
mnojek, line 237:
```suggestion
        ``Test Template`` sets the template to all tests in a suite. Empty value is considered an error
        because it leads the users to wrong impression on how the suite operates.
        Without value, the setting is ignored and the tests are not templated.
```

mnojek (approved):

PR #824, opened 2023-04-18
Add rule for Test Case/Task in resource file

### LINKED ISSUE #525: Defining *** Test Case / Task *** section in resource file is reported as parsing-error
Related #524 

If Robocop auto detect file as resource type, but it will contain `*** Test Case / Task ***` section it will be reported as parsing error. It should be parsed and reported as other type of issue:
test.robot:
```
*** Settings ***
Resource    resource.robot

```
resource.robot:
```
*** Test Cases ***
Test
    No Operation

```
Command:
```
robocop test.robot resource.robot
```
Output:
```
resource.robot:1:1 [E] 0401 Robot Framework syntax error: Resource file with 'Test Cases' section is invalid. (parsing-error)
```

bhirsz:
Closes #525

### DIFF
--- a/robocop/checkers/errors.py
+++ b/robocop/checkers/errors.py
@@ -220,9 +220,10 @@ rules = {
     "0415": Rule(
         rule_id="0415",
         name="invalid-section-in-resource",
-        msg="Resource file with '{{ section_name }}' section is invalid",
+        msg="Resource file can't contain '{{ section_name }}' section",
         docs="""
-        Resource file cannot contain ``Test Cases`` or ``Tasks`` sections.
+        The higher-level structure of resource files is the same as that of test case files,
+        but they can't contain Test Cases or Tasks sections.
         """,
         severity=RuleSeverity.ERROR,
     ),
--- a/robocop/checkers/errors.py
+++ b/robocop/checkers/errors.py
@@ -12,7 +12,7 @@ except ImportError:
 
 from robocop.checkers import VisitorChecker
 from robocop.rules import Rule, RuleSeverity
-from robocop.utils import ROBOT_VERSION, find_robot_vars, token_col
+from robocop.utils import ROBOT_VERSION, find_robot_vars
 
 rules = {
     "0401": Rule(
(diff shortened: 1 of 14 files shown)


robocop/checkers/errors.py
```diff
@@ -217,6 +217,15 @@
         severity=RuleSeverity.ERROR,
         version=">=5.0",
     ),
+    "0415": Rule(
+        rule_id="0415",
+        name="invalid-section-in-resource",
+        msg="Resource file with '{{ section_name }}' section is invalid",
```
mnojek, line 223: I think I would go with something similar what is in the docs below, like:
`Resource file can't contain '{{ section_name }}'`. It's so short and easier to understand that more general "Resource file with (section) is invalid", because it does not provide the reason.

What do you think?

robocop/checkers/errors.py
```diff
@@ -217,6 +217,15 @@
...
+        docs="""
+        Resource file cannot contain ``Test Cases`` or ``Tasks`` sections.
```
mnojek, line 225: For the docs we can then use a bit more extended sentence like:
"The higher-level structure of resource files is the same as that of test case files, but they can't contain ``Test Cases``  or ``Tasks`` sections."

robocop/checkers/errors.py
```diff
@@ -217,6 +217,15 @@
         severity=RuleSeverity.ERROR,
         version=">=5.0",
     ),
+    "0415": Rule(
```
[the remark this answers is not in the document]
bhirsz: Oh, right. I think it should be separate rule (or we should rename the rule to be more generic, like invalid-syntax-in-resource)

robocop/checkers/errors.py
bhirsz: This error is literally what Robot Framework already reports (and we're catching this as parsing-error). But I like ``Resource file can't contain '{{ section_name }}'`` better so I will use it

robocop/checkers/errors.py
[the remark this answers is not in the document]
mnojek: I think we can add a separate rule for that.

mnojek (approved):

PR #835, opened 2023-04-21
Add unused_argument and argument/variable-overwritten-before-usage  rules

### LINKED ISSUE #501: Detect unused arguments
I suggest creating a new rule which would report an argument of a keyword if it is not used at all in the body of keyword (be it declared in [Arguments], or in the keyword name in case of embedded arguments). E. g.:
```
My Keyword
    [Arguments}  ${arg}  # <-- unused!
    Log  Hello!
```
Note: No dynamic analysis...

Note 2: I do not suggest this new rule to check when an argument is used, but a value is assigned to it only without retieving it again --- this situation should be covered by https://github.com/MarketSquare/robotframework-robocop/issues/209 once it will be implemented. :)

### LINKED ISSUE #836: Rule for variable overwritten before it was used
Example:

```
*** Keywords ***
Defined In Keyword Call
    ${var}    ${var2}    Assign Variables
    Log    ${var2}
    ${var}    Overwrite Variable

Defined In Args
    [Arguments]    ${arg}
    ${arg}    Set Variable    value

Defined In FOR
    FOR    ${var}    IN  @{LIST}
          ${var}    Keyword Call  
    END
```

bhirsz:
Closes #501 
Closes #836


robocop/checkers/misc.py
```diff
@@ -299,17 +306,18 @@
 
         For example::
 
-        Example Keyword
```
bhirsz, line 302: I didn't change the docs here, only added indent - if the code example has the same indent as ``line::``, it will not render correctly:

image

robocop/checkers/misc.py
```diff
@@ -339,6 +348,27 @@
             END
         """,
     ),
+    "0919": Rule(
+        rule_id="0919",
+        name="unused-argument",
+        msg="Keyword '{{ arg }}' argument is not used",
```
mnojek, line 354: I think I would change it to:
```suggestion
        msg="Keyword argument '{{ arg }}' is not used",
```
because the previous one makes me think that it is about an argument from keyword named {{ arg }}, not an argument names  {{ arg }}.

robocop/checkers/misc.py
```diff
@@ -339,6 +348,27 @@
...
+        severity=RuleSeverity.WARNING,
+        docs="""
+        Keyword argument was defined but not used::
+        
+            *** Keywords ***
+            Keyword
+                [Arguments]    ${used}    ${overwritten_before_used}    ${not_used}
+                Log    ${used}
+                ${overwritten_before_used}    Process Stuff    value
+                Log    ${overwritten_before_used}  # used, but original argument was overwritten before using
```
mnojek, line 364: I'm wondering if it's good we detect such cases in the `unused-argument` rule, since this is a bit different case and the users might actually not expect it to work that way. Shouldn't we split this into 2 separate rules: `unused-argument` and maybe `overwritten-before-usage`?

bhirsz: I was actually thinking that this rule would be good base for other rules, such as ``unused-local-variable`` rule. It's also similar and it could be also handled by this checker (since we're already doing most of the heavy lifting with parsing the variables etc). And it would be more more obvious to the users.

tl;dr; I agree, let's split it into ``overwritten-before-usage`` rule. In that case I should probably do this in the same PR. I will create issue for it (let's discuss naming of the rule and misc stuff here), I will link it

bhirsz: #836

mnojek: Maybe, to make it more explicit, we should name it `overwritten-arg-before-usage`? Or even `overwritten-argument-before-usage`? It makes it a bit longer, but it better points out what is this rule about. WDYT? Or maybe just `overwritten-argument`?

mnojek: Hmm... Yeah, right. Well, does it have any negative impact if we create another separate rule for overwritten local variable? I might be missing something, but I think higher granularity is fine.

bhirsz: Yeah, Im also for more granulity as long it makes sense. Makes it easier to understand issue /disable rule if needed. I will do it then

docs/releasenotes/3.2.0.rst
```diff
@@ -41,17 +41,30 @@ For example::
...
+New rules for overwriting arguments and variables before first use (#836)
+--------------------------------------------------------------------------
+
+Added new rules W0920 ``argument-overwritten-before-usage`` and W0921 ``variable-overwritten-before-usage`` for
+arguments or variables overwritten before first use::
+
+    *** Keywords ***
+    Overwritten Argument
+        [Arguments]    ${overwritten}  # we do not use ${overwritten} value at all
+        ${overwritten}    Set Variable    value
```
mnojek, line 62: I would suggest to rename to:
```suggestion
        [Arguments]    ${overwritten}  # we do not call ${overwritten} value at all
        ${overwritten}    Set Variable    value  # we only overwrite it
```

because we use it (for overwriting), but we never call it.

docs/releasenotes/3.2.0.rst
```diff
@@ -41,17 +41,30 @@ For example::
...
+
+    Overwritten Variable
+    ${value}    Keyword
+    ${value}    Keyword
```
mnojek, line 66: Adding missing indentation and maybe a bit more convincing example from issue #836:
```suggestion
        ${var1}    ${var2}    Assign Variables
        Log        ${var2}  # this variable is used
        ${var1}    Overwrite Variable  # this one is only overwritten
```

robocop/checkers/misc.py
```diff
@@ -351,22 +351,52 @@
...
+        name="argument-overwritten-before-usage",
+        msg="Keyword argument '{{ name }}' is overwritten before usage",
+        severity=RuleSeverity.WARNING,
+        docs="""
+        Keyword argument was overwritten before it is used::
+        
+            *** Keywords ***
+            Overwritten Argument
+                [Arguments]    ${overwritten}  # we do not use ${overwritten} value at all
+                ${overwritten}    Set Variable    value
```
mnojek, line 383:
```suggestion
                ${overwritten}    Set Variable    value  # we only overwrite it
```

robocop/checkers/misc.py
```diff
@@ -351,22 +351,52 @@
...
+
+        """,
+    ),
+    "0921": Rule(
+        rule_id="0921",
+        name="variable-overwritten-before-usage",
+        msg="Local variable '{{ name }}' is overwritten before usage",
+        severity=RuleSeverity.WARNING,
+        docs="""
+        Keyword, Test Case or Task local variable is overwritten before it is used::
```
mnojek, line 393:
```suggestion
        Local variable in Keyword, Test Case or Task is overwritten before it is used::
```

robocop/checkers/misc.py
```diff
@@ -797,43 +827,49 @@ def check_statement_in_loop(self, node, token_type):
...
 
 
 class UnusedVariablesChecker(VisitorChecker):
-    reports = ("unused-argument",)
+    reports = ("unused-argument", "argument-overwritten-before-usage", "variable-overwritten-before-usage")
 
     def __init__(self):
         self.arguments = {}
+        self.variables = [{}]  # variables are list of scope-dictionaries, to support IF branches
+        self.ignore_overwriting = False  # temporarily ignore overwriting, ie in FOR loops
```
mnojek, line 839:
```suggestion
        self.ignore_overwriting = False  # temporarily ignore overwriting, e.g. in FOR loops
```

robocop/checkers/misc.py
mnojek, line 1: If possible, please add short docstrings to not documented methods.

mnojek (changes requested):

mnojek (approved):
Nice!

PR #839, opened 2023-04-23
List reports configured

### LINKED ISSUE #833: [Bug]  --list-reports does not list most of the reports
```
robocop --list-reports
```
returns
```
Available reports:
return_status        - Checks if number of specific issues exceed quality gate limits
all                  - Turns on all default reports
```

bhirsz:
Closes #833 

I still need to update the release notes - hence draft status.

I'm not sure how to note that report is non-default (it will not be enabled using ``all``, currently only ``sarif`` works that way). Current output:

```
 robocop --list-reports
Available reports:
file_stats           - Prints overall statistics about number of processed files (disabled)
rules_by_error_type  - Prints total number of issues grouped by severity (disabled)
rules_by_id          - Groups detected issues by rule id and prints it ordered by most common (disabled)
sarif                - Generate SARIF output file (disabled - non-default)
scan_timer           - Returns Robocop execution time (disabled)
timestamp            - Returns Robocop execution timestamp. (disabled)
version              - Returns Robocop version (disabled)

Enable report by passing report name using --reports option. Use `all` to enable all default reports. Non-default reports can be only enabled using report name.

```


mnojek:
> I'm not sure how to note that report is non-default (it will not be enabled using `all`, currently only `sarif` works that way).

Maybe, instead of the report, we should make SARIF report a separate CLI option (`--sarif`)?

robocop/reports.py
```diff
@@ -9,13 +9,21 @@
...
+To enable all default reports use ``--reports all``.  Non-default reports can be only enabled using report name.
 
 The order of the reports is preserved. For example, if you want ``timestamp`` report to be printed before any
 other reports, you can use following configuration::
 
     robocop --reports timestamp,all src.robot
 
+Print list of all reports with their configured status by using ``--list-reports``::
+
+    robocop --reports all --list-reports
```
mnojek, line 21: Is it required to also call `--reports all` when the user just wants to list all the reports? Would be great to be able to just run `robocop --list-reports` and see all of them.

bhirsz: It is not required. If you just run list reports it will run them all but they will have status (disabled). --reports all for only to show how configuration affects the output

docs/releasenotes/3.2.0.rst
```diff
@@ -83,12 +83,36 @@ existed in the root directory. Now, it is possible to load it from any path usin
 List reports now list all reports (#833)
 ----------------------------------------
 
-<insert description here>
+List reports command previously only listed enabled reports. Now the option list all reports together with its
+enabled/disabled status. Also, internal reports are now omitted::
```
mnojek, line 87:
```suggestion
List reports command (``--list-reports``) previously only listed enabled reports. Now, the option lists all reports
together with their enabled/disabled status. Also, internal reports are now omitted::
```

docs/releasenotes/3.2.0.rst
```diff
@@ -83,12 +83,36 @@ existed in the root directory. Now, it is possible to load it from any path usin
...
+    Available reports:
+    file_stats           - Prints overall statistics about number of processed files (disabled)
+    rules_by_error_type  - Prints total number of issues grouped by severity (disabled)
+    rules_by_id          - Groups detected issues by rule id and prints it ordered by most common (disabled)
+    sarif                - Generate SARIF output file (disabled - non-default)
+    scan_timer           - Returns Robocop execution time (disabled)
+    timestamp            - Returns Robocop execution timestamp. (disabled)
+    version              - Returns Robocop version (disabled)
+
+    Enable report by passing report name using --reports option. Use `all` to enable all default reports.
```
mnojek, line 99:
```suggestion
    Enable report by passing its name using ``--reports`` option. Use ``all`` to enable all default reports.
```

docs/releasenotes/3.2.0.rst
```diff
@@ -83,12 +83,36 @@ existed in the root directory. Now, it is possible to load it from any path usin
...
+    Non-default reports can be only enabled using report name.
```
mnojek, line 100:
```suggestion
    Non-default reports (like ``sarif``) can be only enabled using their name explicitly.
```

docs/releasenotes/3.2.0.rst
```diff
@@ -83,12 +83,36 @@ existed in the root directory. Now, it is possible to load it from any path usin
...
+
+
+List reports output can be filtered using ENABLED/DISABLED (#833)
 ------------------------------------------------------------------
 
-<insert description here>
+Same as ``--list`` rules option, ``--list-reports`` now support optional ``ENABLED`` / ``DISABLED`` argument to filter
```
mnojek, line 106:
```suggestion
Same as ``--list`` rules option, ``--list-reports`` now supports optional ``ENABLED`` / ``DISABLED`` argument to filter
```

mnojek (approved):

PR #840, opened 2023-04-27
Add unused-variable rule

### LINKED ISSUE #209: Rule: variable assigned but not used

bhirsz:
Closes #209 

Thanks for previously added rules (such as ``unused-arguments``) it was easy to add this rule . And it was in our backlog for 2,5 years.. ;) 

I have also changed the ids of the rules. We didn't release them yet so it's fine. It made more sense to have ``unused-variable`` right after ``unused-argument``.

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -375,21 +375,23 @@ rules = {
         msg="Variable '{{ name }}' is assigned but not used",
         severity=RuleSeverity.INFO,
         docs="""
-    Variable was assigned but not used::
-
-        *** Test Cases ***
-        Test
-            ${variable}    Keyword Call  # not used variable
-            ${variable2}    Keyword Call
-            Log    ${variable2}
-
-    Use ``${_}`` variable name if you purposefully do not use variable::
-
-        *** Keywords ***
-        Keyword
-            FOR    ${_}   IN RANGE    10
-                Process Value
-            END
+        Variable was assigned but not used::
+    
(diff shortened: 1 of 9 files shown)


robocop/checkers/misc.py
```diff
@@ -371,6 +371,30 @@
...
         rule_id="0920",
+        name="unused-variable",
+        msg="Variable '{{ name }}' is assigned but not used",
+        severity=RuleSeverity.INFO,
+        docs="""
+    Variable was assigned but not used::
+
+        *** Test Cases ***
+        Test
+            ${variable}    Keyword Call  # not used variable
```
mnojek, line 382: I am thinking about making the examples like a real-world use cases. Something like:

```
*** Keywords ***
Get Triangle Base Points
    [Arguments]       ${triangle}
    ${p1}    ${p2}    ${p3}    Get Triangle Points    ${triangle}
    Log      Triangle base points are: ${p1} and ${p2}.
    RETURN   ${p1}    ${p2}
```

In the above example, variable `${p3}` is never used.

mnojek (approved):

robocop/checkers/misc.py
bhirsz: Definitely looks better - I have also started to use more real like examples somewhere else, it just reads better than ``Test, Keyword Call,,,`` dry stuff :)

PR #843, opened 2023-04-28
Add inconsistent-variable-name rule

### LINKED ISSUE #496: Create a rule checking whether the name of a variable is always written the same way
This could be a sort of extension of rule _possible-variable-overwriting_. 

It need not be obvious that variables e.g. `${abc}` and `${A_B C}` are the same.  Imagine the situation: someone sees variable `${MY VAR}` in a foreign code. They want to check occurences of the variable to discover where it is used and for what purpose etc. but they are not able to find all its occurences because the same variable is named e.g. `${MY_VAR}` somewhere. In this case it is almost impossible to find all occurences of the variable by standard means: AFAIK, Robot language server cannot do it; and standard search for a string can usualy ignore the case of letters, but cannot ignore mismatch in spaces and underscores. The only way I consider to be usable for search for all the possible forms of the variable is regular expression `[mM][_ ]?[yY][_ ]?[vV][_ ]?[aA][_ ]?[rR]`. Ugh.

My suggestion: allow the only form of the name of a variable. It would be probably the best to check whether the variable is named everywhere like _at its first occurence_ (since this will probably be its definition or first setting of its value).

### LINKED ISSUE #841: [Bug] possible-variable-overwriting does not detect IF assignment
Example
```
${var}   IF   $cond    Keyword
${VAR}   IF   $cond    Keyword
```

### LINKED ISSUE #842: [Bug] possible-variable-ovewriting does not handle default values
Example
```
Keyword
    [Arguments]    ${variable}='value'
    ${va_riable}    Keyword Call  # should report, but it does not
```

bhirsz:
Closes #496 

Fixes #841 
Fixes #842


docs/releasenotes/3.2.0.rst
```diff
@@ -66,9 +66,28 @@ arguments or variables overwritten before first use::
         Log        ${var2}  # this variable is used
         ${var1}    Overwrite Variable  # this one is only overwritten
 
+New rule for not consistent variable naming (#496)
```
mnojek, line 69:
```suggestion
New rule for inconsistent variable naming (#496)
```

docs/releasenotes/3.2.0.rst
```diff
@@ -66,9 +66,28 @@ arguments or variables overwritten before first use::
...
+--------------------------------------------------
+
+Added new rule W0323 ``inconsistent-variable-name`` which reports variables with inconsistent naming::
```
mnojek, line 72:
```suggestion
Added new rule W0323 ``inconsistent-variable-name`` which reports variables with inconsistent naming.
Since Robot Framework ignore underscores and spaces in variable names, and also normalizes them to
lowercase, this code should report new tule violations::
```

robocop/checkers/naming.py
```diff
@@ -242,6 +243,17 @@
...
+        docs="""
+        Following assignments overwrite the same variable::
+        
+            *** Keywords ***
+            Keyword
+                ${variable}    Keyword Call
+                ${VARIABLE}    Keyword Call
+                ${vari_ab le}    Keyword Call
+        
+        Use consistent variable naming guidelines to avoid not intended variable overwriting.
```
mnojek, line 255:
```suggestion
        Use consistent variable naming guidelines to avoid unintended variable overwriting.
        Remember that variable names in Robot Framework are case-insensitive and
        underscores and whitespaces are ignored.
```

robocop/checkers/naming.py
```diff
@@ -350,6 +362,36 @@
...
     ),
+    "0323": Rule(
+        rule_id="0323",
+        name="inconsistent-variable-name",
+        msg="Variable '{{ name }}' has inconsistent naming. First used as '{{ first_use }}'",
+        severity=RuleSeverity.WARNING,
+        docs="""
+        Variable names are case-insensitive and ignore underscores and spaces. It is possible to 
+        write the variable in multiple ways and it will be a valid Robot Framework code. However
+        it makes it harder to maintain the code that is not following consistent naming.
```
mnojek, line 373:
```suggestion
        Variable names are case-insensitive and ignore underscores and spaces. It is possible to 
        write the variable in multiple ways and it will be a valid Robot Framework code. However,
        it makes it harder to maintain the code that does not follow the consistent naming.
```

robocop/checkers/naming.py
```diff
@@ -726,61 +759,168 @@ def visit_KeywordCall(self, node):  # noqa
 class SimilarVariableChecker(VisitorChecker):
     """Checker for finding same variables with similar names."""
 
-    reports = ("possible-variable-overwriting",)
+    reports = ("possible-variable-overwriting", "inconsistent-variable-name")
 
     def __init__(self):
-        self.variables = defaultdict(set)
+        self.assign_variables = defaultdict(list)
```
mnojek, line 765: I'm not a fan of this name. It suggests an activity, because `assign` is used like a verb here. Maybe we should name it `variable_variants`?

mnojek (approved):

robocop/checkers/naming.py
bhirsz: I agree on name suggesting activity. It's strictly containing variables that were assigned somewhere (from keyword calls, for loop helper variable, if return). The ``variable_variants`` would be too generic. Maybe ``assigned_variables``? If not, I will go with ``variable_variants

mnojek: `assigned_variables` is fine, let's do it.

PR #844, opened 2023-04-29
Create e2e tests workflow

### LINKED ISSUE #832: Add pipeline for running Robocop on Robot Framework internal tests
We could avoid issues like #829 if we would run all our rules on all possible Robot Framework syntax. It can be achieved by using Robot Framework repository and their own internal tests. It can be in form of Github pipeline and run on merges to master/or on demand.

bhirsz:
Closes #832


mnojek (approved):
❤️

PR #851, opened 2023-05-31
Fix 0404 rule reporting on args for variable files

### LINKED ISSUE #845: [Bug] import variable files with module name and arguments gives a wrong variables-import-with-args message
### What happened?

starting with this simple suite

```robot
*** Settings ***
Variables       variables    ${ENVIRON}

*** Variables ***
${ENVIRON}    dev

*** Test Cases ***
first test case
    Log    ${ENVIRONMENT_NAME}
```

where I import a variable file with arguments, I get this error:

```
...\import_vars.robot:2:1 [E] 0404 Robot and YAML variable files do not take arguments (variables-import-with-args)
```

robocop .\tests\import_vars.robot

```
...\import_vars.robot:2:1 [E] 0404 Robot and YAML variable files do not take arguments (variables-import-with-args)
```

starting from robotframework 5 it is possible to import variables with its module name like libraries, see also [documentation](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#taking-variable-files-into-use) 

so this message is invalid with robotframework version >= 5

I think the rule 0404 should check that the file extension is `yaml` and not `py`.
And are you shure that you can import `robot` files as variables? I think not. What do you mean with robot files?

### Operating System

windows, linux

### Robocop version

3.1.1

mnojek:
Fixes #845

### DIFF
--- a/robocop/checkers/errors.py
+++ b/robocop/checkers/errors.py
@@ -60,14 +60,15 @@ rules = {
     "0404": Rule(
         rule_id="0404",
         name="variables-import-with-args",
-        msg="Robot and YAML variable files do not take arguments",
+        msg="YAML variable files do not take arguments",
         severity=RuleSeverity.ERROR,
+        version="<5.0",
         docs="""
         Example of rule violation::
         
             *** Settings ***
-            Variables    vars.yaml    arg1
-            Variables    variables.robot    arg
+            Variables    vars.yaml        arg1
+            Variables    variables.yml    arg2
         
         """,
     ),
@@ -611,5 +612,5 @@ class VariablesImportErrorChecker(VisitorChecker):
     reports = ("variables-import-with-args",)
 
     def visit_VariablesImport(self, node):  # noqa
-        if node.name and not node.name.endswith(".py") and node.get_token(Token.ARGUMENT):
+        if node.name and (node.name.endswith(".yaml") or node.name.endswith(".yml")) and node.get_token(Token.ARGUMENT):
             self.report("variables-import-with-args", node=node)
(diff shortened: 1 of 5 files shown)


robocop/checkers/errors.py
```diff
@@ -60,14 +60,15 @@
     "0404": Rule(
         rule_id="0404",
         name="variables-import-with-args",
-        msg="Robot and YAML variable files do not take arguments",
+        msg="YAML variable files do not take arguments",
         severity=RuleSeverity.ERROR,
+        version="<5.0",
```
bhirsz, line 65: By doing that, this rule will be disabled for 5 and 6+ RF versions. I think it should be still enabled, but only report on arguments supplied to yaml file.

mnojek: You're totally right. I was in a hurry and I should mark it as WIP, because I didn't even make the tests work.
Thanks for noting that. I will update the code.

mnojek: I think very slowly lately but I just revisited the issue and I think we should not even use the version condition here. The rule should always report if the YAML file has arguments, because it's not possible for any RF version. The 5+ versions only add a possibility to import the variable file using module name. Am I right?

bhirsz: I agree, there is no need to limit the version. We just need to raport only on yaml variablw file with arguments (and not on python variable file like it is now).

mnojek:
It works now. @bhirsz Please take a look and review when you have time, thanks :)

bhirsz (approved):

robocop/checkers/errors.py
```diff
@@ -611,5 +612,5 @@ class VariablesImportErrorChecker(VisitorChecker):
     reports = ("variables-import-with-args",)
 
     def visit_VariablesImport(self, node):  # noqa
-        if node.name and not node.name.endswith(".py") and node.get_token(Token.ARGUMENT):
+        if node.name and (node.name.endswith(".yaml") or node.name.endswith(".yml")) and node.get_token(Token.ARGUMENT):
```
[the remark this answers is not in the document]
mnojek: Thanks! Updated :)

PR #866, opened 2023-06-15
Fix variable rules to accept  as ignored variable syntax

### LINKED ISSUE #862: [Bug] variable-overwritten-before-usage should ignore ${_}
${_} may be used to denote that variable name is not important. variable-overwritten-before-usage should ignore those variables.

bhirsz:
Those rules are not yet released hence no need to update release docs with the fix.

Fixes #862

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -423,7 +423,7 @@ rules = {
                 ${value}    Keyword
                 ${value}    Keyword
 
-        In case the value of the variable is not important it is possible to use `${_}` name::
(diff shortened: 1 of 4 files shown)


robocop/checkers/misc.py
```diff
@@ -423,6 +423,15 @@
                 ${value}    Keyword
                 ${value}    Keyword
 
+        In case the value of the variable is not important it is possible to use `${_}` name::
```
mnojek, line 426:
```suggestion
        In case the value of the variable is not important, it is possible to use `${_}` name::
```

mnojek (approved):
LGTM

PR #871, opened 2023-06-16
Fix issues with detecting variables used in loop or with math operations

### LINKED ISSUE #868: [Bug] Variables in loop are not detected as used

```
WHILE    ${counter} < 10
    Log To Console    ${counter}
END
```

```
WHILE    ${counter} < 10
    ${not_used}    Keyword
END
```

### LINKED ISSUE #870: [Bug] Variable with simple operations are not detected as used variable
```
${sum}    Calculate    ${random}
${modulo}    Set Variable    ${sum%11}
```

sum is used with simple operation (modulo %). Same with + etc.

bhirsz:
Fixes #868 
Fixes #870


docs/releasenotes/3.2.1.rst
```diff
@@ -0,0 +1,61 @@
+:orphan:
+
+=============
+Robocop 3.2.1
+=============
+
+Fix issues with variable rules.
```
mnojek, line 7:
```suggestion
This release fixes the issues with new variable rules.
```

mnojek (approved):

PR #872, opened 2023-06-16
Fix handling comments in W1003 rule

### LINKED ISSUE #861: [Bug] empty-lines-between-sections ignores comments
Following code:

```
*** Variables ***
${VAR}    value

# TODO: Add more variables


*** Test Cases ***
```

reports empty-lines-between-sections 3/2 couting the lines before comment.

mnojek:
Fixes #861 

Another changes:

- Adds a `tests/conftest.py` file to enable testing specific rules with `pytest` command
- Changes the session names in `nox` to shorter ones: `robot_version` → `rf`
- Documents how to run specific test


tests/atest/rules/spacing/empty_lines_between_sections/bug861.robot
```diff
@@ -0,0 +1,23 @@
+*** Settings ***
+Library     String
+
+# Comment
```
bhirsz, line 4: bug861.robot:4:1 [W] 1003 Invalid number of empty lines between sections (1/2)

Shouldn't it be 0/2? If at all, it's problematic because some people want to have comments right above header


```
*** Settings ***


*** Keywords ***  # fine
```

```
*** Settings ***


# comment - should be fine?
*** Keywords ***
```

```
*** Settings ***


# comment

*** Keywords ***  1/2 issue
```

mnojek: In a `tests/atest/rules/spacing/empty_lines_between_sections/golden.robot` file we have a case:
```
*** Variables ***
${MY_VAR}    1


# some comment
*** Keywords ***
```

and the issue is not reported there. So I don't know 😁

mnojek: Maybe we should just simplify it and count from 2 empty lines after any kind of text (including comments)?

bhirsz: Yeah, I'm for simplification

bhirsz: But please check history of git - I have strangr feeling we discussed something similar not so long ago and thats why it is implemented this way

mnojek: Ok, found it. We discussed it in #611. And I think that if we align current implementation of 1003 rule to what was decided in the related issue, it should work exactly as I mentioned above - it should not apply fancy rules for comments and just ignore them and count right after the last comment. All other empty lines between comments will be handled by the 1005 rule.

mnojek: Changed, take a look.

bhirsz: Now it looks fine!

Do you want it to be part of current fix release (not yet released 3.2.1) or another feature release (3.3.0)? Please update release docs with this change depending on which version you want to use. If you don't want it to include it in 3.2.1 I will make release before you merge this PR

mnojek: Since this is a fix, we can push it with the 3.2.1. I will adjust the release notes.

mnojek: Done. I think we're ready to release.

bhirsz (approved):

PR #873, opened 2023-06-16
Add new rule for unsupported settings in resource files

### LINKED ISSUE #827: New rule for checking invalid settings in Resource file
What about [this sentence from User Guide](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#resource-file-structure):

"Additionally, the Setting section in resource files can contain only import settings (Library, Resource, Variables) and Documentation. The Variable section and Keyword section are used exactly the same way as in test case files."

Are we going to have a separate rule for that?

_Originally posted by @mnojek in https://github.com/MarketSquare/robotframework-robocop/pull/824#discussion_r1169604875_

mnojek:
Closes #827 

Added new rule E0416 `invalid-setting-in-resource` that detects unsupported settings in resource files.

Also, updated `.gitignore` with RF output files.

### DIFF
--- a/docs/releasenotes/3.3.0.rst
+++ b/docs/releasenotes/3.3.0.rst
@@ -1,0 +1,45 @@
+:orphan:
+
+=============
+Robocop 3.3.0
+=============
+
+Description.
+
+You can install the latest available version by running
+
+::
+
+    pip install --upgrade robotframework-robocop
+
+or to install exactly this version
+
+::
+
+    pip install robotframework-robocop==3.3.0
+
+.. contents::
+   :depth: 2
+   :local:
+
+Rule changes
+============
+
+New rule for detecting not supported settings in resource files (#827)
+----------------------------------------------------------------------
+
+According to the official RF User Guide:
+
+"Additionally, the Setting section in resource files can contain only
+import settings (Library, Resource, Variables) and Documentation.
+The Variable section and Keyword section are used exactly the same way
+as in test case files."
+
+New E0416 ``invalid-setting-in-resource`` rule detects unsupported
+settings in resource files.
+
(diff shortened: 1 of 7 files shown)


docs/releasenotes/3.3.0.rst
```diff
@@ -0,0 +1,45 @@
...
+
+New rule for detecting not supported settings in resource files (#827)
+----------------------------------------------------------------------
+
+According to the official RF User Guide:
+
+"Additionally, the Setting section in resource files can contain only
+import settings (Library, Resource, Variables) and Documentation.
+The Variable section and Keyword section are used exactly the same way
+as in test case files."
```
bhirsz, line 36: Additionally sounds bit weird here since it's the beginning. Let's start the quote from the Settings section

```suggestion
"Setting section in resource files can contain only import settings (Library, Resource, Variables)
and Documentation. The Variable section and Keyword section are used exactly the same way
as in test case files.
```

tests/atest/rules/errors/invalid_setting_in_resource/test_rule.py
```diff
@@ -0,0 +1,9 @@
+from tests.atest.utils import RuleAcceptance
+
+
+class TestRuleAcceptance(RuleAcceptance):
+    def test_rule(self):
+        self.check_rule(expected_file="expected_output_rf6.txt", target_version="==6.*")
```
bhirsz, line 6: what about RF 7? Maybe it should be:

```suggestion
        self.check_rule(expected_file="expected_output_rf6.txt", target_version=">=6")
```

bhirsz (approved):

PR #874, opened 2023-06-17
Improve RST formatting in docs and wording

mnojek:
- Replaced singular backticks with doubled, to properly apply RST code formatting
- Updated some sentences, fixed typos and improved wording
- Added new `robot:flatten` tags from RF 6.1

### DIFF
--- a/robocop/checkers/tags.py
+++ b/robocop/checkers/tags.py
@@ -29,18 +29,18 @@ rules = {
         " Hint: make sure to include this tag using lowercase name to avoid issues",
         severity=RuleSeverity.INFO,
         docs="""
-        OR and AND words are used to combine tags when selecting tests to be run in Robot Framework. Using following 
+        ``OR`` and ``AND`` words are used to combine tags when selecting tests to be run in Robot Framework. Using following 
         configuration::
         
             robot --include tagANDtag2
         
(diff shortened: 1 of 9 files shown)


robocop/checkers/tags.py
```diff
@@ -170,6 +171,7 @@ class TagNameChecker(VisitorChecker):
     is_keyword = False
     reserved_tags = {
         "robot:exit",
+        "robot:flatten",
```
bhirsz, line 174: At this point I wonder if Robot Framework store reserved tags in some constant we can use? Instead of updating our rule whenever there is new reserved tag.

mnojek: Good point, I will check.

mnojek: Haven't found it in the docs. In the API there's only this: https://robot-framework.readthedocs.io/en/stable/_modules/robot/model/tags.html

bhirsz: Right, I see they're checking for tags depending on the current context. Ie https://github.com/robotframework/robotframework/blob/6a266bf852ad0525c350ea7412f7b33f81455e96/src/robot/running/context.py#L181. We need to stick to our current implementation.

bhirsz (approved):

PR #875, opened 2023-06-17
Add rule for checking task-related settings

### LINKED ISSUE #522: It is possible to use Task X setting names (Task Teardown, Task Timeout..).
See https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#id605 . We need to verify if it doesn't break anything in Robocop and all issues are properly reported. 

Should we suggest using Task X variant if there is Task section defined in the file?

mnojek:
Closes #522 

Ia m now thinking if we should make another rule to check the opposite for files with `*** Test Cases ***` section, but when the task-related settings are used.
Or maybe this should be one rule that checks for `settings-consistency`?

And let me know if you agree that this should be an INFO rule, or maybe it should be a WARNING severity level?


bhirsz:
I'm conflicted on severity but we can go with info.

It would be good to also check it for test cases. And to not multiply rules unnecessary we can go with one common settings-consistency.

docs/releasenotes/3.3.0.rst
```diff
@@ -37,6 +37,14 @@ as in test case files.
 New E0416 ``invalid-setting-in-resource`` rule detects unsupported
 settings in resource files.
 
+New rule that suggests using task-related settings if Task section is used (#522)
+---------------------------------------------------------------------------------
+
+If the file contains ``*** Task ***`` section, it is recommended to use settings like ``Task Setup``,
+``Task Teardown``, ``Task Template`` and ``Task Timeout`` instead of their `Test` variants.
```
bhirsz, line 44: Test has single backticks, where other name have double backticks

bhirsz, line 44:
```suggestion
``Task Teardown``, ``Task Template`` and ``Task Timeout`` instead of their ``Test`` variants.
```

robocop/checkers/naming.py
```diff
@@ -420,6 +421,16 @@
             ```
             """,
     ),
+    "0326": Rule(
+        rule_id="0326",
+        name="task-related-setting",
+        msg="Use task-related setting '{{ setting }}' if Tasks section is used.",
+        docs="""
+        If ``*** Tasks ***`` section is present in the file, use task-related settings like ``Task Setup``,
```
bhirsz, line 430: Since there are Test Tags, are there Task Tags? Just curious

robocop/checkers/naming.py
```diff
@@ -668,8 +681,18 @@ def visit_SectionHeader(self, node):  # noqa
                 end_col=node.col_offset + len(name) + 1,
             )
 
+    def visit_File(self, node):  # noqa
+        for section in node.sections:
+            if isinstance(section, TestCaseSection) and "task" in section.header.name.lower():
+                self.task_section = True
```
bhirsz, line 687: What if it is translated section name? I believe that since RF 6 the Task header is actually TaskSectionHeader instead of Test Case like before. We can use it to support ltranslations since RF 6. Pre 6 your code should work fine

robocop/checkers/naming.py
```diff
@@ -720,6 +747,10 @@ def check_setting_name(self, name, node):
                 "setting-name-not-in-title-case", setting_name=name, node=node, col=col + 1, end_col=col + len(name) + 1
             )
 
+    def check_task_related_settings(self, name, node):
+        if "test" in name.lower() and not "task" in name.lower():
```
bhirsz, line 751: We also dont support translations here for RF 6+

bhirsz: Also looking at condition - can name contain both task and test in name? Do we need to check for both?

robocop/checkers/naming.py
mnojek: I actually haven't found `TaskSectionHeader`. In the latest docs there is:
image

robocop/checkers/naming.py
mnojek: Will check.

docs/releasenotes/3.3.0.rst
mnojek: Here I added singular backticks on purpose, since "Test" alone is not any part of the code, but it's a word from statements like `Test Teardown`. So here it does not stand as an RF code element, rather as a name for the variants of it.

robocop/checkers/naming.py
mnojek: Oh, you're right. I was thinking to first filter out all settings like `Documentation`, which don't have test- or task-related variants. But, yeah, no sense in doing that.

robocop/checkers/naming.py
bhirsz: The section is TestCaseSection for both, but now the header can be TASK_HEADER (see your screenshot, the last line).

To test this theory, I have created following file:

```
language: pl


*** Zadania ***
Test Timeout    1 min
```
and put breakpoint in visit_TestCaseSection. That's what I got:

image

It properly recognized that "Zadania" is Task header - so for RF 6 it's better to check for Task Header (note it doesn't exist pre RF 6)

bhirsz: I've also added *** Ustawienia *** with Test/Task Timeouts to see how they look like when translated:

Task Timeout:
image

Test Timeout:
image


Sadly without cross referencing it with translation mapping we don't know if "Test timeout" is actually test or task timeout.

bhirsz: To sum it up, you can add support to translated section names, but supporting translated setting names is too much work right now. You can add # TODO in the code to denote it could be done in the future.

mnojek: I think we should create an issue for that directly in RF repo. I'll do it to request this change.

robocop/checkers/naming.py
mnojek: Fixed.

robocop/checkers/naming.py
```diff
@@ -668,8 +681,21 @@ def visit_SectionHeader(self, node):  # noqa
                 end_col=node.col_offset + len(name) + 1,
             )
 
+    def visit_File(self, node):  # noqa
+        for section in node.sections:
+            if isinstance(section, TestCaseSection):
+                if ROBOT_VERSION.major < 6 and "task" in section.header.name.lower():
+                    self.task_section = True
+                elif ROBOT_VERSION.major == 6 and section.header.type == Token.TASK_HEADER:
```
bhirsz, line 689: ROBOT_VERSION.major >= 6

mnojek:
> What about creating one common `settings-consistency` rule instead of one dedicated to task/test?

Oh, yeah, I forgot about it. Thanks 😁 I'll do it.

robocop/checkers/naming.py
mnojek: Updated.

robocop/checkers/naming.py
```diff
@@ -668,8 +681,21 @@ def visit_SectionHeader(self, node):  # noqa
                 end_col=node.col_offset + len(name) + 1,
             )
 
+    def visit_File(self, node):  # noqa
+        for section in node.sections:
+            if isinstance(section, TestCaseSection):
+                if ROBOT_VERSION.major < 6 and "task" in section.header.name.lower():
```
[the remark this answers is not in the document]
mnojek: Good idea, updated.

mnojek:
@bhirsz I am not 100% satisfied with the rule name (`settings-consistency`) but I don't have a better candidate at this moment. We can change it later before the next release, if we come up with something.

Also, I am not entirely sure it INFO severity is good here. I am still hesitating to make it a WARNING.

bhirsz:
Let's go with WARNING then, since I'm also on fence with it. 

How about ``mixed-task-test-settings``?

tests/atest/rules/naming/settings-consistency/test_rule.py
```diff
@@ -0,0 +1,17 @@
+from tests.atest.utils import RuleAcceptance
+
+
```
bhirsz, line 3: We have golden all task, golden all test, all test but task section, all task but test section but we don't have mixed tests (mixes test/task settings with task/test section).

mnojek: Added. Please check if it's fine now and if that's what you meant.

mnojek:
> Let's go with WARNING then, since I'm also on fence with it.
> 
> How about `mixed-task-test-settings`?

Agreed. Updated the severity to WARNING and changed the name – much better!

Issue #809, opened 2023-04-01
Create a place where community rules can be stored


mnojek:
As the idea sparked in the comments of #680, here is a follow-up issue to discuss the topic and how it can be achieved.

### The idea

A centralized place for the rules that do not fit into the core of Robocop, from where the users can easily download them and use in their projects.

The reasons why a rule may not fit into the core of Robocop:
- it reports something that is possible/allowed in Robot Framework, but not very common
- it reports a very specific issue with the code, which is not a general problem
- it requires reading the context (more than one file) or importing the libraries/resources and so on
- ?

I imagine this in 3 ways (but there may be more, and even better than these):
1. A separate repository serving as a storage for custom rules, where users can suggest adding new ones by PR. Each rule may need to have basic tests to ensure that it's working properly for the scope it affects, and a meaningful documentation.
2. Custom rules located in a separate directory in this project and available for the users by running `pip install robotframework-robocop[extra-rules]` which will install them to their local Robocop instance (proposed by @bhirsz [here](https://github.com/MarketSquare/robotframework-robocop/issues/680#issuecomment-1248481659)).
3. Maybe just a documentation page where the working examples will be presented and available to download/copy, with examples and documentation?

### Things for further investigation
1. Who would decide which rule should be added to the storage? Authors? Community (by voting)? Or anything that meets minimum requirements (tests, docs, etc.) would be accepted?
2. The custom rules should be specifically marked, to not misguide people this is a built-in rule (some extra field or maybe just the ID convention - like 5-digit 1XXYY or 99YY)?
3. Should all the custom rules be downloaded and turned on (off?) locally by default, or should they be picked individually by the users before downloading? I imagine an issue when a user downloads all custom rules and some may actually be failing in his project, or maybe report too many issues, or just didn't apply to their specific case.

### Rule candidates

I'll just mention the issues where the users asked for a specific rule to be added, which we rejected, due to various reasons.

- #680 - inform the user that `Test Teardown`/`Setup` and `Teardown`/`Setup` are used within the same test suite
- #766 - detect unusual symbols in keyword name / test case name
- #688 - detect case where built-in keyword was not properly separated with the arguments (not enough spaces)
- #778 - detect Sleep keyword
- #666 - detect if each word is capitalized in the test case name
- #566 - detect when the keyword is called with an acceptable but a non-standard way of writing its name
- #850 - suggest using `robot:continue-on-failure` tag
[...]

PR #876, opened 2023-06-18
Create rules disabled by default

bhirsz:
Relates #809

I have added option to pass ``enabled`` parameter when creating the Rule. It allows us to create rules disabled by default.
The enabled parameter reuses the same concept as ``severity`` and it makes it possible to configure it from CLI too (robocop -c rule_name:enabled:True). 

I have also detected issue that plagued us for quite a long time - the tests that were failing not related to our changes, and only if we run all tests. Turns out when we're loading the rules we're importing them (robocop.checkers.comments.. ) and each test configuration can overwrite the rules attributes stored in sys.modules. I have added a workaround that forcefully removes the already imported rules module if we attempt to load it again (for example when running multiple tests in the same process).


docs/external_rules.rst
```diff
@@ -148,6 +148,29 @@ Dotted syntax is also supported::
 :code:`rules` dictionary should be available at the same level as checker that is using it. That's why if you are defining your
 external rules using modules and ``__init__.py`` it should be also imported (or defined directly in ``__init__.py``).
 
+Disabled by default rules
+-------------------------
+
+All rules are enabled by default and included after importing them. It is possible to define rule that is disabled
+by default by using ``enabled`` parameter::
```
mnojek, line 155:
```suggestion
All rules are enabled by default and included after importing them. It is possible to define a rule that is disabled
by using ``enabled`` parameter with ``False`` value::
```

docs/external_rules.rst
```diff
@@ -148,6 +148,29 @@ Dotted syntax is also supported::
...
+            msg="Custom rule message",
+            severity=RuleSeverity.INFO,
+            enabled=False,
+            docs="""
+            Custom rule description.
+            """,
+        )
+    }
+
+Such rules can be enabled with ``--include`` option or by configuring ``enabled`` parameter directly::
```
mnojek, line 170:
```suggestion
Such rules can be enabled when called explicitly with ``--include`` option::

    robocop --include custom-rule .

 or by configuring ``enabled`` parameter directly::
```

docs/releasenotes/3.3.0.rst
```diff
@@ -37,6 +37,32 @@ as in test case files.
 New E0416 ``invalid-setting-in-resource`` rule detects unsupported
 settings in resource files.
 
+Other features
+==============
+
+New ``enabled`` Rule parameter used to create disabled by default rules (#809)
```
mnojek, line 43:
```suggestion
New ``enabled`` rule parameter used to create rules disabled by default (#809)
```

docs/external_rules.rst
```diff
@@ -148,6 +148,29 @@ Dotted syntax is also supported::
 :code:`rules` dictionary should be available at the same level as checker that is using it. That's why if you are defining your
 external rules using modules and ``__init__.py`` it should be also imported (or defined directly in ``__init__.py``).
 
+Disabled by default rules
```
mnojek, line 151:
```suggestion
Rules disabled by default
```

docs/releasenotes/3.3.0.rst
```diff
@@ -37,6 +37,32 @@ as in test case files.
...
+------------------------------------------------------------------------------
+
+Previously, all defined rules were enabled by default and could be disabled using ``--exclude`` option (or by
+using ``--include`` if not listed with the option). It is now possible to create rule disabled by default using
```
mnojek, line 47:
```suggestion
using ``--include`` if not listed with the option). It is now possible to create a rule disabled by default using
```

docs/releasenotes/3.3.0.rst
```diff
@@ -37,6 +37,32 @@ as in test case files.
...
+            msg="Custom rule message",
+            severity=RuleSeverity.INFO,
+            enabled=False,
+            docs="""
+            Custom rule description.
+            """,
+        )
+    }
+
+Such rules can be enabled using ``--include`` option or by configuring ``enabled`` parameter directly::
```
mnojek, line 63:
```suggestion
Such rules can be enabled when called explicitly with ``--include`` option::

    robocop --include custom-rule .

 or by configuring ``enabled`` parameter directly::
```

robocop/utils/misc.py
mnojek, line 1: I think it makes sense to split these 2 changes into separate PRs - one for ``enabled`` parameter, and one for fixing the issue with Pytest. Is it fine for you to do it?

mnojek (changes requested):

robocop/utils/misc.py
bhirsz: The problem is that this PR will not work without this fix. I could split and merge the fix first, then this PR, is it fine?

mnojek: It would be better, but if you don't see the value in doing this, I am fine with merging these 2 things in one PR.

mnojek (approved):

PR #877, opened 2023-06-18
Add added_in_version to mark rule release version

### LINKED ISSUE #788: Documentation: Marking the release for new rules
Hi,

you deliver new rules from release to release (thanks for that), but there is no information in the documentation what are the rules that have been introduced in the given release.
Could you please somehow mark the given rule that came with which release?

For example in the rule documentation table:
After "Rule severity", "Rule id", and "Robot Framework version" as a fourth column.

And/or highlight it as a new section in the Release notes.

Thank you in advance!

bhirsz:
Closes #788

If the added_in_version is added to rule, it looks like this:

image

Otherwise, it looks like this:

image

I have also added test (currently expected to fail) that checks if all our internal rules have added_in_version parameter.


docs/rules.rst
```diff
@@ -29,9 +29,11 @@ Rules
   * - Severity
     - Rule id
     - Robot Framework version
+    - Robocop version
```
mnojek, line 32: This may not be very clear for the users what is means. It may suggest that the rule is only available for this specific Robocop version.
I would rephrase it to `New from Robocop` or `Added in Robocop version` or `Introduced in Robocop (version)`. I am not 100% sure if the word `version` is needed here or not (or maybe added in parenthesis?).

mnojek (commented):
Thanks for this change!
One small comment regarding the name of the table header.

docs/rules.rst
bhirsz: Maybe let's go with ``New from Robocop``, without version

mnojek (approved):

mnojek:
Wow, actually I am impressed how fast you did it. I thought it will be a long task to complete 😁 

Regarding the `bad-indent` – I think we can add the latest date when it was changed, since the previous implementation of it was a bit different. WDYT?

bhirsz:
I'm also suprised :D But we're actually doing good job of documenting our code changes (our each release is tagged by robocop version, So you can git checkout x.y.z tag).

Agreed

mnojek (approved):
Amazing!

PR #878, opened 2023-06-19
Split reports into separate files

### LINKED ISSUE #834: Refactor reports code
All Robocop reports code is stored in one file `robocop/reports.py` and it's hard to maintain, since all classes and util functions and the parent class are in one place. We should refactor them to locate each report into a separate file, which will make it easier to maintain the code and easier to grasp the logic behind it.

bhirsz:
Closes #834


mnojek (commented):
All good, but since we are already touching it, I decided to read the docs and improved them a bit with a better wording.

PR #880, opened 2023-06-19
Use threshold value in rule message instead of minimum level

### LINKED ISSUE #791: Print issues for rules that accept "Rule severity thresholds" using threshold level not the minimum level
Reading log like bellow is quite confusing:

```
backup_alarm.robot:17:1 [W] 0505 Test case 'ATC U-TS SpO2 Check Probe Backup Invalid Measurement Status' has too many keywords inside (13/10) (too-many-calls-in-test-case)
backup_alarm.robot:33:1 [W] 0505 Test case 'ATC U-TS SpO2 Check Probe Backup Frequent Status Changes' has too many keywords inside (24/10) (too-many-calls-in-test-case)
backup_alarm.robot:62:1 [E] 0505 Test case 'ATC U-TS SpO2 Check Probe Backup Alarm Precondition' has too many keywords inside (30/10) (too-many-calls-in-test-case)
backup_alarm.robot:95:1 [W] 0505 Test case 'ATC U-TS SpO2 Check Probe Backup Alarm Filtering' has too many keywords inside (19/10) (too-many-calls-in-test-case)
```
If you don't know the configuration you might be wondering why I'm getting different severity levels when all are higher than 10?

So the reference number should be the value which triggered the corresponding severity level.

Even I have started to wonder if my config is not read at all because error should only happen if the count is higher than (or equal) to 30 (in this case).

bhirsz:
Closes #791 

If the rule severity changed due to configured severity threshold, the rule message should now display severity threshold that caused it instead of default rule min/max value. Not configured severity should be omitted when reporting the rule.

I still need to update docs in this PR.


docs/releasenotes/3.3.0.rst
```diff
@@ -56,6 +56,25 @@ Our internal reports structure was refactored to make it easier to maintain and
 each report is stored in a separate file in Robocop project. From the user's perspective, reports should work
 the same as before.
 
+Severity threshold print value (#791)
+-------------------------------------
+
+Rules with severity threshold should now print the message with the
```
bhirsz, line 62: I will rewrite release notes later, because I see few typos or sentences that could be worded better.

robocop/rules.py
```diff
@@ -220,6 +221,18 @@ def get_severity(self, value):
                 return severity
         return None
 
+    def get_matching_threshold(self, value):
+        """
+        Find first threshold that pass the condition with passed value.
```
mnojek, line 226:
```suggestion
        Find first threshold that passes the condition with passed value.
```

docs/releasenotes/3.3.0.rst
mnojek: Sure. I will review this part later then.

mnojek (commented):
Code looks fine. Waiting for the release notes to be updated. Also, I think the documentation needs to be updated too.

docs/releasenotes/3.3.0.rst
bhirsz: Done

docs/releasenotes/3.3.0.rst
```diff
@@ -84,6 +84,35 @@ Our internal reports structure was refactored to make it easier to maintain and
 each report is stored in a separate file in Robocop project. From the user's perspective, reports should work
 the same as before.
 
+Rule severity thresholds printed value (#791)
+-----------------------------------------------
+
+Rule severity threshold allows to change rule severity depending on the threshold value and configured severities.
+However it was confusing to see reports for the same rule with different severities but the same limit value.
```
mnojek, line 91:
```suggestion
In fact, it was confusing to see the reports for the same rule with different severities but the same limit value.
```

docs/releasenotes/3.3.0.rst
```diff
@@ -84,6 +84,35 @@ Our internal reports structure was refactored to make it easier to maintain and
...
+
+Previously, if the rule severity threshold was configured but the rule value didn't met any conditions it was reported
+with default severity. Now if the value don't pass any configured severity threshold, the issue is ignored.
+
+For example, if the ``line-too-long`` (with default severity W) is configured to have severity E if number of lines
+is higher than 150::
+
+    robocop -c line-too-long:severity_threshold:error=150
+
+With above configuration, ``line-too-long`` will only be reported if number of lines is higher than 150.
```
mnojek, line 111:
```suggestion
With above configuration, ``line-too-long`` will only be reported if number of the lines is higher than 150.
```

mnojek (commented):
Small suggestions

mnojek (approved):

docs/releasenotes/3.3.0.rst
```diff
@@ -84,6 +84,37 @@ Our internal reports structure was refactored to make it easier to maintain and
...
+In fact, it was confusing to see the reports for the same rule with different severities but the same limit value.
+We changed this behaviour to log the actual value that met the severity threshold.
+
+As for example, you can configure ``too-many-calls-in-test-case`` to report warning on 10 keywords and error on 30
+keywords. With such configuration you will get following reports::
+
+    test.robot:33:1 [W] 0505 Test case 'Test case 1' has too many keywords inside (24/10) (too-many-calls-in-test-case)
+    test.robot:62:1 [E] 0505 Test case 'Test case 2' has too many keywords inside (30/30) (too-many-calls-in-test-case)
+
+Not configured severity threshold should be ignored (#791)
```
mnojek, line 100: I just realized that this change is backward-incompatible. Am I right?

This is not a big change that affects how Robocop works in general, but it affects the configuration of severity for the rules. I am thinking if that forces us to make a major release 4.0.0, or can we just add it to release notes as backward-incompatible and go with 3.3.0?

@bhirsz What do you think?

bhirsz: It's not entirely backward incompatible - depends on how we define it :) In the past we released major versions for changes that **breaks** the code, make configuration invalid or big changes to the tool etc. This change requires users to update the severity thresholds if they want them to catch all issues (in case they omitted default severity in severity threshold config). But it doesn't break anything, we simply changes how it behaves. I will go with 3.3.0 but add a note that this change requires updating current configuration

mnojek: Updated in #885

PR #881, opened 2023-06-20
Add new rules for Keyword Tags

### LINKED ISSUE #641: Support new setting Keywords Tags in RF 5.1
From https://github.com/robotframework/robotframework/issues/4373

This require:
1) updating our visitors (so tags from Keywords Tags are visited and also setting related rules like naming ones does work on new setting)
2) refactoring some of our tag rules

mnojek:
Closes #641 

This PR adds 3 new rules to support `Keyword Tags`:

``empty-keyword-tags`` rule (W0531) checks for the situation where ``Keyword Tags`` in `*** Settings ***` section
does not have any value provided.

``could-be-keyword-tags`` rule (I0610) checks if all keywords in a file share the same tag. In that case, it can be
simply added to ``Keyword Tags`` setting, which the rule suggests.

``tag-already-set-in-keyword-tags`` rule (I0611) detects a situation where all keywords share the same tag
which is already added to the ``Keyword Tags`` setting. In such case, the tags in keywords may be removed.

It also adds few more tests to already existing rules, since I found some corner cases where there were false positives, due to the fact that `visit_Tags` also visits keywords' tags, which I fixed.


robocop/checkers/tags.py
```diff
@@ -356,3 +399,69 @@ def visit_Tags(self, node):  # noqa
...
+        self.keyword_tags_node = node
+
+    def visit_KeywordSection(self, node):  # noqa
+        self.in_keywords = True
+        self.generic_visit(node)
+        self.in_keywords = False
+
+    def visit_Tags(self, node):  # noqa
+        if self.in_keywords:
+            self.keywords_tags.append([tag.value for tag in node.data_tokens[1:]])
```
bhirsz, line 455: Keywords_tags is sometimes set and sometimes list, what type it should be?

robocop/checkers/tags.py
```diff
@@ -166,6 +166,48 @@
...
+        added_in_version="3.3.0",
+    ),
+    "0611": Rule(
+        rule_id="0611",
+        name="tag-already-set-in-keyword-tags",
+        msg="Tag '{{ tag }}' is already set by {{ keyword_tags }} in suite settings",
+        severity=RuleSeverity.INFO,
+        version=">=6",
+        docs="""
+        Avoid repeating same tags in keywords when the tag is already declared in ``Keyword Tags``.
```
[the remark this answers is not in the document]
mnojek: Yup, you're right. It was copied from another rule. I fixed that one too.

robocop/checkers/tags.py
mnojek: I think you might be confused because there are `self.keyword_tags` (for storing all tags set by `Keyword Tags` setting) and `self.keywords_tags` (for storing all tags set in keywords' `[Tags]` setting). I renamed the latter to `self.tags_in_keywords` to avoid confusion.

Also, the code is heavily copied from the corresponding test-related rules for tags.

bhirsz: Ah, you're right - didn't notice the extra s. Thanks for the rename

bhirsz (approved):

PR #883, opened 2023-06-21
Unsupported settings in init file

### LINKED ISSUE #498: [Rule] Setting 'X' not allowed in Y file should be separate rule
parsing-error rule can report:
```
Setting 'Force Tags' is not allowed in resource file. (parsing-error)
```
Same for other suite only settings and for ``__init__.robot`` file. It could be separate rule ie not-supported-setting-for-file

mnojek:
Closes #498


docs/releasenotes/3.3.0.rst
```diff
@@ -63,6 +64,14 @@ simply added to ``Keyword Tags`` setting, which the rule suggests.
 ``tag-already-set-in-keyword-tags`` rule (I0611) detects a situation where all keywords share the same tag
 which is already added to the ``Keyword Tags`` setting. In such case, the tags in keywords may be removed.
 
+New rule for discovering unsupported setting in initialization file (#498)
+--------------------------------------------------------------------------
+
+According to RF User Guide, ``Default Tags`` and ``Test Template`` settings are not allowed
```
bhirsz, line 70: What about Test Timeout, Test Teardown, Test Setup, Task equivalents etc?

robocop/checkers/naming.py
```diff
@@ -770,6 +787,24 @@ def visit_LibraryImport(self, node):  # noqa
...
 
+    def visit_SettingSection(self, node):  # noqa
+        self.generic_visit(node)
+        if not self.init_file:
+            return
+        for setting in node.body:
+            if not setting.data_tokens:
+                continue
+            setting_name = setting.data_tokens[0].value
+            if setting_name.lower() in ["test template", "default tags"]:
```
bhirsz, line 798: If we would parse error instead, would we also support translated settings? I'm thinking about pros and cons of both implementations

mnojek: Good point. I will check it. The translations introduced a lot of complications to our tool :D

mnojek: Yes, it works for translations if this is just a properly served error. It does not work for the current implementation.
I will update the PR.

mnojek:
Interesting part is that [there is no translation](https://github.com/robotframework/robotframework/blob/6a266bf852ad0525c350ea7412f7b33f81455e96/src/robot/conf/languages.py#L732) for `Default Tags` and `Force Tags` in any language.
I know they are deprecated, but they are still supported.

robocop/checkers/naming.py
mnojek: Updated. Please review.

bhirsz (approved):

PR #886, opened 2023-06-25
Import external rules using ``import x.y`` syntax

### LINKED ISSUE #790: Load all checker modules (and rules) under a specific package (or folder).
Instead of listing every checker package individually in robocop configs it should be possible to define a top package under with all the valid checker modules (and rules) are loaded.

bhirsz:
Closes #790

I did rewrite all our core methods for importing rules & checkers. Our previous solution used 3 different ways of importing the rules (depending if it's internal, external from file or external from module). The implementation is now more homogeneous while also allowing to use different kind of imports in external rules. The change is fully backward compatible (all our old tests are still working) but I have decided to bump to 4.0.0 due to size of the changes in our core methods.


docs/external_rules.rst
```diff
@@ -18,70 +18,83 @@ Every custom checker needs to complete following requirements:
...
 ..  code-block:: python
-    :caption: dummy.py
+    :caption: example.py
 
     from robocop.checkers import VisitorChecker
     from robocop.rules import Rule, RuleSeverity
 
     rules = {
-        "9999": Rule(rule_id="9999", name="dummy-in-name", msg="There is 'Dummy' in test case name", severity=RuleSeverity.WARNING)
+        "9999": Rule(rule_id="9999", name="example-in-name", msg="There is 'Example' in test case name", severity=RuleSeverity.WARNING)
```
bhirsz, line 30: Renamed ``dummy``, it now sounds a bit more inclusive :)

docs/external_rules.rst
```diff
@@ -18,70 +18,83 @@ Every custom checker needs to complete following requirements:
...
+    robocop --ext-rules my/own/rule.py --configure example-in-name:param_name:AnotherExample
 
-Value of the configurable parameter can be retrieved using :code:`param` method::
+Value of the configurable parameter can be retrieved using :code:`param` method:
+
+..  code-block:: python
 
     self.param("name-of-the-rule", "name-of-param")
 
+Parameter value is passed as string. Use ``converter`` argument to define method that will be use to convert the value:
```
bhirsz, line 85: I have added missing documentation for Param converters

tests/test_data/ext_rules/ext_rule_module_simple_import/RobocopRules/single_rule.py
```diff
@@ -0,0 +1,21 @@
...
+}
+
+
+class CustomRuleChecker(VisitorChecker):
+    """Checker for missing keyword name."""
+
+    reports = ("external-rule",)
+
+    def visit_KeywordCall(self, node):  # noqa
+        if node.keyword and "Dummy" not in node.keyword:
```
mnojek, line 20: Maybe let's not use "dummy" here as well?

tests/test_data/ext_rules/ext_rule_module_simple_import/RobocopRules/submodule/some_rules.py
```diff
@@ -0,0 +1,14 @@
...
+rules = {"9904": Rule(rule_id="9904", name="external-rule2", msg="This is external rule", severity=RuleSeverity.INFO)}
+
+
+class CustomRuleChecker2(VisitorChecker):
+    """Checker for missing keyword name."""
+
+    reports = ("external-rule2",)
+
+    def visit_KeywordCall(self, node):  # noqa
+        if node.keyword and "Dummy" not in node.keyword:
```
mnojek, line 13: Maybe let's not use "dummy" here as well?

mnojek (commented):
Thanks for this big refactor!

I am still hesitating if we should make a major release, since nothing is backward-incompatible here. For the users it's even less critical. I mean, it's a great step forward, but mostly for the readability of code and for a part of the users that uses the external rules.

docs/external_rules.rst
mnojek, line 85:
```suggestion
Parameter value is passed as string. Use ``converter`` argument to define a method that will be used to convert the value:
```

bhirsz:
> Thanks for this big refactor!
> 
> I am still hesitating if we should make a major release, since nothing is backward-incompatible here. For the users it's even less critical. I mean, it's a great step forward, but mostly for the readability of code and for a part of the users that uses the external rules.

I was also hesitating on it. But my other changes for disablers that will come in other PR are definietely backward incompatible (I will change how disablers disable lines, limit them in scope) so its fine to bumo version in this PR.

docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,22 @@ rule (E0417).
 Other features
 ==============
 
+Overhaul of rules & checkers importing (#790)
+---------------------------------------------
+
+Complete refactor of how Robocop imports rules and checkers internally. Our import mechanism is now more uniform.
```
bhirsz, line 81: @mnojek I have rewritten this section

docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,22 @@ rule (E0417).
...
+Additionally, it is now possible to load external rules from the modules imported inside Python file.
+
+Given following file with external rules::
+
+    import custom_rules.naming
+    import custom_rules.errors
+    from custom_rules.lengths import rules, CustomChecker
+
+Before the change, only ``CustomChecker`` would be loaded by Robocop and content of the ``custom_rules.naming`` and
```
mnojek, line 90:
```suggestion
before the change, only ``CustomChecker`` would be loaded by Robocop, while the content of the ``custom_rules.naming`` and
```

docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,22 @@ rule (E0417).
...
+``custom_rules.errors`` would be ignored. Now such imports will be parsed and external Robocop rules defined
+inside them will be also loaded.
```
mnojek, line 92:
```suggestion
inside them will be loaded as well.
```

docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,22 @@ rule (E0417).
 Other features
 ==============
 
+Overhaul of rules & checkers importing (#790)
+---------------------------------------------
+
+Complete refactor of how Robocop imports rules and checkers internally. Our import mechanism is now more uniform.
+Additionally, it is now possible to load external rules from the modules imported inside Python file.
+
+Given following file with external rules::
```
mnojek, line 84:
```suggestion
Let's assume we have the following code with loading external rules::
```

mnojek (approved):
Thanks, it's much better. I just added few suggestions and now I think it will be perfect :)

PR #887, opened 2023-06-26
Allow to disable rules withing the scope of the current block

### LINKED ISSUE #777: Diable(/Enable) rules from code in keyword/test case block
It should be possible to disable/enable rules on Keyword and Test Suite level so that they are only applied inside that scope.
(like in PyLint)

F.g. If I would add following line:
`# robocop: disable=too-many-calls-in-keyword`
Inside keyword (after the settings) it would be valid only for that keyword.

These lines inside Keywords/Test Cases should be applied to rules for:

- Keyword/Test Case name
- Keyword/Test Case settings
- Keyword/Test Case code

bhirsz:
Closes #777


robocop/checkers/documentation.py
```diff
@@ -126,12 +126,19 @@ def visit_File(self, node):  # noqa
...
         else:
+            ext_disablers = (node.lineno, node.end_lineno) if extend_disablers else None
             if hasattr(node, "name"):
-                self.report(msg, name=node.name, node=node, end_col=node.col_offset + len(node.name) + 1)
+                self.report(
+                    msg,
+                    name=node.name,
+                    node=node,
+                    end_col=node.col_offset + len(node.name) + 1,
+                    ext_disablers=ext_disablers,
```
mnojek, line 141: `ext_disablers` name can be confusing... `ext` can be understood as `extended`, `external` or `extensions`. Better to give a full name here, I think.

mnojek (commented):
Thanks for this big change!
Definitely a step forward.

docs/including_rules.rst
```diff
@@ -60,28 +60,35 @@ or::
 
 Ignored blocks can partly overlap. Rule name and rule id can be used interchangeably.
 
-It is possible to ignore whole file if you start file with ``# robocop: disable`` and won't provide
-``# robocop: enable`` before end of file.
+It is possible to ignore whole file if you start file with ``# robocop: disable``.
```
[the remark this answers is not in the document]
bhirsz: Added sentence along with example (I have tried to make it more real world like to adhere to #888)

docs/including_rules.rst
```diff
@@ -60,6 +60,23 @@ or::
 
 Ignored blocks can partly overlap. Rule name and rule id can be used interchangeably.
 
+The disabler is aware of the context where it was called, and it will be enabled again at the end of the current code
+block (such as keyword, test case, "for" and "while" loop and "if" statement). In following example,
+``wrong-case-in-keyword-name`` disabler will be enabled only inside "for" loop.
```
mnojek, line 65:
```suggestion
block (such as keyword, test case, "for" and "while" loops and "if" statement). In the following example,
``wrong-case-in-keyword-name`` disabler will disable the rule only inside the "for" loop.
```

mnojek (approved):

PR #889, opened 2023-06-28
Add JSON report

### LINKED ISSUE #663: Add JSONReport
We have internal ``JsonReport`` but it is only used to generate issues in json format for API purposes. It is confusing for users and we could solve some issues (like #554 or #617) if we do refactor this report and how it's handled internally.

bhirsz:
Closes #663

We already had JSON report but it was only used internally to return issues in json format when running Robocop programmatically. I renamed previous report to ``internal_json_report`` and created new ``json_report``.

### DIFF
--- a/docs/releasenotes/4.0.0.rst
+++ b/docs/releasenotes/4.0.0.rst
@@ -75,6 +75,19 @@ rule (E0417).
 Other features
 ==============
 
+New report ``json_report`` (#663)
+---------------------------------
+
+Added new ``json_report`` report that produces file with issues in JSON format. It's non default report which needs
+to be enabled using report name (it will not be loaded with ``--reports all``)::
+
+    robocop --reports json_report .
+
+It will save file to current working directory under ``robocop_output.json`` filename. Output directory and filename
+can be configured::
+
+    robocop --configure json_report:output_dir:output --configure json_report:report_filename:issues.json --reports json_report .
+
 Overhaul of rules & checkers importing (#790)
 ---------------------------------------------
 
--- a/robocop/config.py
+++ b/robocop/config.py
@@ -289,8 +289,9 @@ class Config:
             help="Generate reports after scan.\n"
             "You can enable reports by listing them in comma-separated list:\n"
             "--reports rules_by_id,rules_by_error_type,scan_timer\n"
-            "To enable all reports use all:\n"
-            "--reports all",
+            "To enable all default reports use all:\n"
+            "--reports all\n"
+            "List available reports with --list-reports option.",
         )
         optional.add_argument(
             "-f",
--- a/robocop/reports/json_report.py
+++ b/robocop/reports/json_report.py
@@ -1,3 +1,6 @@
+import json
+from pathlib import Path
+
 from robocop.reports import Report
 from robocop.rules import Message
 
@@ -6,6 +9,55 @@ class JsonReport(Report):
     """
     Report name: ``json_report``
 
+    Report that returns list of found issues in JSON format. The output file will be generated
+    in the current working directory with the ``robocop_output.json`` name.
+
+    This report is not included in the default reports. The ``--reports all`` option will not enable this report.
+    You can still enable it using report name directly: ``--reports json_report`` or ``--reports all,json_report``.
+
+    You can configure output directory and report filename::
+
+        robocop --configure json_report:output_dir:C:/json_reports
+        robocop --configure json_report:report_filename:robocop.json
+
+    """
+
+    DEFAULT = False
+
+    def __init__(self):
+        self.name = "json_report"
+        self.description = "Accumulates found issues in JSON format"
+        self.output_dir = None
+        self.report_filename = "robocop_output.json"
+        self.issues = []
+
+    def add_message(self, message: Message):
+        self.issues.append(message.to_json())
+
+    def get_report(self):
+        if self.output_dir is not None:
+            output_path = self.output_dir / self.report_filename
+        else:
+            output_path = Path(self.report_filename)
+        with open(output_path, "w") as fp:
+            json_string = json.dumps(self.issues, indent=4)
+            fp.write(json_string)
+        return f"Generated JSON report in {output_path}"
+
+    def configure(self, name, value):
+        if name == "output_dir":
+            self.output_dir = Path(value)
+            self.output_dir.mkdir(parents=True, exist_ok=True)
+        elif name == "report_filename":
+            self.report_filename = value
+        else:
+            super().configure(name, value)
+
+
+class InternalJsonReport(Report):
+    """
+    Report name: ``internal_json_report``
+
     Report that returns list of found issues in JSON format.
     """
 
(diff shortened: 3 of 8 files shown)


docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,19 @@ rule (E0417).
 Other features
 ==============
 
+New report ``json_report`` (#663)
+---------------------------------
+
+Added new ``json_report`` report that produces file with issues in JSON format. It's non default report which needs
+to be enabled using report name (it will not be loaded with ``--reports all``)::
```
mnojek, line 82:
```suggestion
Added new ``json_report`` report that produces file with issues in JSON format. It's a non-default report which needs
to be called explicitly by name (it will not be loaded with ``--reports all``)::
```

docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,19 @@ rule (E0417).
...
+
+    robocop --reports json_report .
+
+It will save file to current working directory under ``robocop_output.json`` filename. Output directory and filename
+can be configured::
```
mnojek, line 87:
```suggestion
It will save the results to a JSON file in the current working directory under ``robocop_output.json`` filename.
Output directory and filename can be configured with::
```

robocop/config.py
```diff
@@ -289,8 +289,9 @@ def _create_parser(self):
             help="Generate reports after scan.\n"
             "You can enable reports by listing them in comma-separated list:\n"
             "--reports rules_by_id,rules_by_error_type,scan_timer\n"
-            "To enable all reports use all:\n"
-            "--reports all",
+            "To enable all default reports use all:\n"
```
mnojek, line 292:
```suggestion
            "To enable all default reports use 'all':\n"
```

robocop/reports/json_report.py
```diff
@@ -6,14 +9,63 @@ class JsonReport(Report):
     """
     Report name: ``json_report``
 
+    Report that returns list of found issues in JSON format. The output file will be generated
+    in the current working directory with the ``robocop_output.json`` name.
```
mnojek, line 13:
```suggestion
    Report that returns a list of found issues in a JSON format. The output file will be generated
    in the current working directory with the ``robocop_output.json`` name.
```

robocop/reports/json_report.py
```diff
@@ -6,14 +9,63 @@ class JsonReport(Report):
...
+        robocop --configure json_report:output_dir:C:/json_reports
+        robocop --configure json_report:report_filename:robocop.json
+
+    """
+
+    DEFAULT = False
+
+    def __init__(self):
+        self.name = "json_report"
+        self.description = "Accumulates found issues in JSON format"
```
mnojek, line 29: Maybe this is a slightly better description? No strong opinion, though.
```suggestion
        self.description = "Produces JSON file with found issues"
```

robocop/reports/json_report.py
```diff
@@ -6,14 +9,63 @@ class JsonReport(Report):
...
+        self.output_dir = None
+        self.report_filename = "robocop_output.json"
```
mnojek, line 31: Maybe just `robocop.json`? Do we have any other use of JSON files in relation to Robocop which may conflict here?

robocop/reports/json_report.py
```diff
@@ -6,14 +9,63 @@ class JsonReport(Report):
...
+
+    def get_report(self):
+        if self.output_dir is not None:
+            output_path = self.output_dir / self.report_filename
+        else:
+            output_path = Path(self.report_filename)
+        with open(output_path, "w") as fp:
+            json_string = json.dumps(self.issues, indent=4)
+            fp.write(json_string)
+        return f"Generated JSON report in {output_path}"
```
mnojek, line 45:
```suggestion
        return f"Generated JSON report at {output_path}"
```

robocop/reports/json_report.py
```diff
@@ -6,14 +9,63 @@ class JsonReport(Report):
...
+            self.report_filename = value
+        else:
+            super().configure(name, value)
+
+
+class InternalJsonReport(Report):
+    """
+    Report name: ``internal_json_report``
+
     Report that returns list of found issues in JSON format.
```
mnojek, line 61: Maybe we should explain where it "returns" the issues?

docs/releasenotes/4.0.0.rst
```diff
@@ -75,6 +75,19 @@ rule (E0417).
...
+
+    robocop --configure json_report:output_dir:output --configure json_report:report_filename:issues.json --reports json_report .
+
```
mnojek, line 90: Shouldn't we mention here that the previous `json_report` has now a new name? So that the users know they need to update it.

tests/e2e/test_e2e.py
```diff
@@ -70,6 +70,7 @@ def test_list_all_reports(self, robocop_instance, capfd):
         exp_msg = (
             "Available reports:\n"
             "file_stats           - Prints overall statistics about number of processed files (disabled)\n"
+            "json_report          - Accumulates found issues in JSON format (disabled - non-default)\n"
```
mnojek, line 73: Update the tests accordingly.

mnojek (commented):
Huge enhancement!

robocop/reports/json_report.py
```diff
@@ -6,14 +9,63 @@ class JsonReport(Report):
...
+
+    This report is not included in the default reports. The ``--reports all`` option will not enable this report.
+    You can still enable it using report name directly: ``--reports json_report`` or ``--reports all,json_report``.
+
+    You can configure output directory and report filename::
+
+        robocop --configure json_report:output_dir:C:/json_reports
+        robocop --configure json_report:report_filename:robocop.json
+
+    """
```
mnojek, line 23: Maybe we should add an example of the short output that will be produced by this report?

robocop/reports/json_report.py
bhirsz: No,  there shouldn't be conflict. I also used robocop.json initially, not sure why I renamed it but I will revert it, sounds better.

robocop/reports/json_report.py
bhirsz: It is internal report, so it's only for us. But I will add description -> it actually doesn't return anything, it just stores the issues which can be retrieved later.

docs/releasenotes/4.0.0.rst
bhirsz: Previous one was internal report and shouldn't be used by users anyway (at least I hope..). But I will mention this in our release docs

robocop/reports/json_report.py
mnojek: I am thinking if it is not intended to be used by the users, maybe we should mark it in some way? Or just document it here that "Used internally by Robocop, not for external use." or something like that?

bhirsz: It is already marked as such, thanks to attributes:

```
    DEFAULT = False
    INTERNAL = True
```

Thanks to INTERNAL attribute it will not be listed in the ``--list-reports``, not present in the docs etc. We have ``json_reports`` for years already and since we added INTERNAL some time ago it was hidden to users. So nothing will change here, we just renamed this internal report and added new one with the same name as old internal one

mnojek (approved):

robocop/reports/json_report.py
mnojek: Sounds good to me.

PR #890, opened 2023-06-28
Add unnecessary-string-conversion rule

### LINKED ISSUE #442: [Rule] Unnecessary string conversion in conditions
A lot of people are not aware how evaluating condition works in Robot Framework (it uses Python) and create following conditions:
```
IF    r'''${value}''' == 'True'
    Keywords
END
```
instead of using value directly:
```
IF    ${value}
    Keywords
END
```
It could be tricky in some cases that's why we can work on this rule over the releases (extend for new cases).

bhirsz:
Closes #442

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -452,6 +452,51 @@ rules = {
         """,
         added_in_version="3.2.0",
     ),
+    "0923": Rule(
+        rule_id="0923",
+        name="unnecessary-string-conversion",
+        msg="Variable '{{ name }}' in {{ block_name }} condition has unnecessary string conversion",
+        severity=RuleSeverity.INFO,
+        version=">=4.0",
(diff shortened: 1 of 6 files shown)


robocop/checkers/misc.py
```diff
@@ -1153,3 +1198,39 @@ def search_by_removing_attr_access(variable_name, variable_scope):
...
+
+    def check_condition(self, node, condition):
+        if not condition:
+            return
+        try:
+            variables = list(VariableIterator(condition))
+        except VariableError:  # for example ${variable which wasn't closed properly
+            return
+        condition_token = node.header.get_token(Token.ARGUMENT)
+        pos = condition_token.col_offset + 1
```
mnojek, line 1223: Let's use `position` instead.

mnojek (approved):

PR #901, opened 2023-06-30
Add compare_runs report for comparing results of different Robocop runs

### LINKED ISSUE #489: Compare results between two Robocop runs
It would be good to have some way of comparing two Robocop runs ie number of new/fixed issues. It could be special report (--report compare).

Other cli tools usually create special directory (like .pylint or .coverage) and store run results there - which is accessed in 2nd and evaluated in 2nd run.

bhirsz:
Closes #489


bhirsz:
I see that I need to add few e2e tests as well to increase coverage.

tests/utest/test_reports.py
bhirsz, line 1: I split the file into separate files with tests (and also added missing tests, half of the reports didnt have dedicated unit tests).

docs/reports.rst
```diff
@@ -3,6 +3,70 @@
...
+Saving the results is disabled by default and can be enabled with ``--persistent`` flag::
+
+    robocop --persistent
+
+or in the toml configuration file::
+
+    [tool.robocop]
+    persistent = true
+
+Only the previous run for the current working directory is saved.
```
mnojek, line 64: Does it mean that only the last one is saved for the current working directory and all other results are removed?

robocop/reports/file_stats_report.py
```diff
@@ -13,24 +16,61 @@ class FileStatsReport(Report):
...
 
     def add_message(self, message: Message):
         self.files_with_issues.add(message.source)
 
-    def get_report(self) -> str:
+    def persist_result(self):
+        return {"files_count": self.files_count, "files_with_issues": len(self.files_with_issues)}
+
+    @staticmethod
+    def get_plurar(count):
```
mnojek, line 33:
```suggestion
    def get_plural(count):
```

Please also fix in other places.

robocop/reports/robocop_version_report.py
```diff
@@ -13,9 +13,16 @@ class RobocopVersionReport(Report):
...
         self.name = "version"
         self.description = "Returns Robocop version"
+        super().__init__(compare_runs)
 
-    def get_report(self) -> str:
-        return f"\nReport generated by Robocop version: {__version__}"
+    def persist_result(self):
+        return {"generated_with": __version__}
+
+    def get_report(self, prev_results) -> str:  # TODO Add note about previous run version
```
mnojek, line 24: Is this comment needed here?

docs/reports.rst
bhirsz: Yes. We could store more, but we would need mechanism to select them for comparison. For now it's only previous run for the given directory where Robocop was run.

robocop/reports/robocop_version_report.py
bhirsz: It should be removed, it was my work in progress comment.

docs/reports.rst
mnojek: I see it as an additional feature. For example, we can have a command that reads all saved runs and displays them as a list with date and some associated hash, and then we can use this hash to compare any runs. For now it's fine to have a simple solution and see what the users say about it.

bhirsz: Ok, so now we can go with current solution but think about extending it later?

No needs for hashes, just the date and run number (5th run in 03/07/23 etc). Though it would require some limits (ie last 20 runs)

mnojek (approved):
:bow:

PR #902, opened 2023-07-03
Add misplaced-not-in-condition rule

### LINKED ISSUE #893: [Rule] misplaced not in condition
Temporary naming for the issue. It should detect:

```
IF not ${var} is None
WHILE not ${var} in @{list}
```

bhirsz:
Closes #893


docs/releasenotes/4.0.0.rst
```diff
@@ -102,6 +102,34 @@ can be rewritten to::
     IF    not ${session_created}    Create Session
     Skip If    not ${rows}
 
+New rule for misplaced not in conditions (#893)
+-----------------------------------------------
+
+Added new rule I0925 ``misplaced-not-in-condition`` that checks if the evaluated condition contains ``not`` operator
```
mnojek, line 108: What about `misplaced-not-condition`?

mnojek (approved):

docs/releasenotes/4.0.0.rst
bhirsz: I don't have strong preference though ``misplaced-not-condition`` sounds like whole 'not-type condition' was misplaced, and not just position of not in condition

mnojek: What about something like `suboptimal-negative-condition`? Can't find a better word than `suboptimal`, though... or maybe `misplaced-negative-condition`?

bhirsz: ``misplaced-negative-condition`` sounds better, I will go with it

mnojek: Sure, please also update release notes.

mnojek (approved):

PR #909, opened 2023-07-05
Add rule dobule-negation-in-condition

### LINKED ISSUE #896: [Rule] unneeded not in condition
```
IF    not not ${value}
```

bhirsz:
Closes #896

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -577,6 +577,36 @@ rules = {
         """,
         added_in_version="4.0.0",
     ),
+    "0926": Rule(
+        rule_id="0926",
+        name="double-negation-in-condition",
+        msg="'{{ block_name }}' condition has unnecessary double negation",
+        severity=RuleSeverity.INFO,
+        version=">=4.0",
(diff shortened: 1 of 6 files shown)


robocop/checkers/misc.py
```diff
@@ -577,6 +577,36 @@
         """,
         added_in_version="4.0.0",
     ),
+    "0926": Rule(
```
mnojek, line 580: BTW Have you seen many cases like this? I am feeling that this rule is a bit redundant, because it checks for a very specific typo. I know that it may result in a wrong condition, but double negation may be put nearly everywhere where Python evaluation is performed, so why we look for it only here?

bhirsz: I saw, but only once in my life :) I have also wondered after raising PR if it's good candidate for rule. Quite specific like you said. I could park this branch for time being and revisit it in the future to decide if it's still needed

mnojek: Agreed. It may also become a community rule, though it's universal (but a very rare case).

PR #910, opened 2023-07-06
Release 4.0.0

bhirsz:
Happy to see proposal for better release 'summary', I'm not the best writer ;) 

Note: It's better to not merge this before I confirm that https://github.com/d-biehl/robotcode will work with our release. I vaguely remember they didn't use our API directly but instantiated disablers class on their own. Since our disablers were rewritten it is not backward compatible.


docs/releasenotes/4.0.0.rst
```diff
@@ -4,7 +4,13 @@
...
 =============
 
-Description.
+Added several new rules, new reports (including comparison with the previous run!). Our rule disablers are smarter
+(context aware) and new rules can be defined to be disabled by default. Checkers & rules importing was also rewritten to
+allow to load external rules from the modules imported inside Python file.
+
+Our documentation was also updated with better formatting of the rules.
+
+There are backward-incompatible changes to rule severity thresholds and rule disablers (described below in details).
```
mnojek: Maybe let's write it like this:

```
This release brings a handful of long-awaited features requested by our users and a few improvements we were keeping in our heads for some time.

Besides the 9 new rules, Robocop is now able to compare results between runs and see the difference in the report.

Moreover, the disablers are now smarter and they become context-aware, meaning that they disable the rules in the whole related code block.

Another improvement is the severity thresholds for the rules, which are now dynamically referenced in the rule messages.

We also added a possibility to create rules that are disabled by default, and this is a first step we made towards Community Rules, which is a huge feature we are working on, and we plan to add it in the next major release.

The results from the Robocop execution can also be exported to a JSON file with our new ``json_report``.

Another updates are:
- a new way of importing custom rules thanks to refactored mechanism of importing rules and checkers
- better readability and formatting of the documentation
- internal refactor of the reports to make it easier to maintain and extend they functionality


Backward-incompatible changes relate to rule severity thresholds and rule disablers (described below in details).

Sit down, grab a popcorn 🍿 , relax and enjoy our release notes!
```

mnojek (approved):
I reviewed and added my proposals as a commit.

PR #911, opened 2023-07-09
Update 4.0.0.rst

bhirsz:
The list doesn't render correctly without empty line before.


bhirsz:
On that note, I believe we need precommit script for validating rst and markdown files. I will open the issue

mnojek (approved):

PR #917, opened 2023-07-12
Rule docs examples

### LINKED ISSUE #888: Rewrite the code snippets to use real-world examples
Currently, our code snippets below the rules very often contain fake examples with a lot of "some", "something", "example" words.

To make them easier to understand, we should try to rewrite them with the examples that imitate real-world code. It would take some time, but it's worth doing.

In addition to that, we can:
- add more emojis to the documentation ([here's the full list](https://sphinxemojicodes.readthedocs.io/en/stable/#supported-codes) of what we can use)
- think of a consistent way of displaying the examples (maybe in a form of a "Good/Bad code" or something else)
- review other documentation (e.g. for disablers and CLI options) and see if those examples can be improved as well

mnojek:
I revisited and improved the documentation and examples for the rules from `naming` and `documentation` categories.

Also, added another improvements to the documentation in overall:

- Renamed `rules.rst` to `rule_basics.rst`
- Moved content from the "Including and excluding rules` to `Rule basics`
- Added new `configuration.rst` file and wrote the content of it
- Removed unnecessary content from "Getting started" and added some links to more extensive docs
- Fixed formatting in 3.2.1 release notes
- Added introduction to the rules list, 
- Added one more case to test for `wrong_case_in_keyword_name` rule
- Updated `nox -s docs` command to always rebuild Sphinx documentation

I think this one closes #888 and partly covers #688 (2/9 rule groups are updated).


docs/configuration.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Robocop provides a wide set of rules with good defaults and is designed to work well out-of-the-box.
+Nevertheless, it is natural that some projects may have different needs and requirements, which influence the way the code is written.
+Thus, Robocop allows to configure many rules to fit these needs.
+It's done using CLI options and such configuration can be stored and loaded from file, as described below.
+
+Configuring rules
+=================
+
+Rules are configurable. Severity of every rule message can be changed (see more :ref:`severity-threshold`) and also some of the rules have
```
bhirsz, line 15: Severity threshold is something different than configuring severity. You can configure severity as const, so it will always use configures value. Additionaly, some selected rules allow to configure 'dynamic' severity and thats severity threshold

docs/configuration.rst
```diff
@@ -0,0 +1,129 @@
...
+    robocop -c "todo-in-comment:markers:Remove me,Fix this!" .
+
+.. note::
+
+    Configuration parameter (``--configure`` / ``-c``) can also be used to configure reports. More about it here :ref:`configuring-reports`.
+
+Loading configuration from file
+===============================
+
+Robocop supports two formats of the configuration file: argument files and toml files. If argument file is not
```
bhirsz, line 53:
```suggestion
Robocop supports two formats of the configuration file: argument files and toml files. If configuration file is not
```

We are now support passing both argument and toml files though cli argument (--argumentfile or - - config), thats why we need to reword it

docs/configuration.rst
```diff
@@ -0,0 +1,129 @@
...
+    ext-rules = ["path_to_external\\dir"]
+    filetypes = [".txt", ".tsv"]
+    threshold = "E"
+    format = "{source}:{line}:{col} [{severity}] {rule_id} {desc} (name)"
+    output = "robocop.log"
+    configure = [
+        "line-too-long:line_length:150",
+        "0201:severity:E"
+    ]
+    no_recursive = true
```
bhirsz, line 129: Maybe we should replace no_recursive with persistent? I wanted to have flag like value in example configuration file. And persistent doesnt affect output that match as no recursive (which alter  whole file searching behaviour) so may be better as example

docs/rules_list.rst
```diff
@@ -7,7 +7,26 @@ Rules list
...
+* 03: :ref:`Naming`
+* 04: :ref:`Errors`
+* 05: :ref:`Lengths`
+* 06: :ref:`Tags`
+* 07: :ref:`Comments`
+* 08: :ref:`Duplications`
+* 09: :ref:`Misc`
+* 10: :ref:`Spacing`
+
+Each category has a corresponding 2-digit ID that is a prefix for the 4-digit rule ID.
```
bhirsz, line 23: It sounds like ID is 6 digits while it is 4. Maybe we should reword this sentence, ie "Each rule has 4-digit ID which starts with corresponding 2-digit category ID"

docs/configuration.rst
mnojek: You're right. I mixed these two.

docs/rules_list.rst
mnojek: Updated.
BTW I was thinking how we should name these parts and maybe something like this?
- rule ID - a 4-digit ID that contains:
  - 2-digit category ID
  - 2-digit rule number

"rule ID" - because we `rule_id` in code
"category ID" - because many checkers may relate to the same category, so checker ID is not a precise name, that's why we should use "category" here
"rule number" - because "rule ID" is already taken, but this is actually just a consecutive number of the rule

WDYT?

docs/configuration.rst
mnojek: Good point, updated.

docs/rules_list.rst
bhirsz: Good idea, using bullet list make it more clear and category is better than checker id (like you wrote, we have multiple checkers in single category already). And we use ordered number rarher than any unique id so its also better name :)

mnojek: Updated.

bhirsz (approved):

Issue #899, opened 2023-06-29
Prepare the basic structure for the community rules


mnojek:
Add an example of the community rule with all necessary structure of the repository:
- place new rule under `robocop/checkers/community_rules` directory
- use `1XXYY` ID as a format, where `1`is a flag for a community rule, `XX` is the checker category (do we need categories here?) and the `YY` is the rule ID (maybe we can just go with `1XXXX` where `XXXX` is just the rule ID?).
- adjust existing test framework to support running tests for community rules
- add test that checks if all community rules are disabled by default
- generate rule documentation and add a new "Community Rules" entry in the docs in the side menu
- ...?

This task is part of the #809 issue.

bhirsz:
For 2) "use 1XXYY ID as a format", we should have some internal tests that check if rule id is in given format.  For internal and community rules.

bhirsz:
Should community rules be filtered out (not visible) by default when listing them with ``--list``?

Currently, given rule ``10000``:

```
> robocop --list
...
Rule - 1017 [E]: bad-block-indent: Indent expected. Provide 2 or more spaces of indentation for statements inside block (enabled)
Rule - 10000 [W]: use-continue-on-failure-tag: Use robot:continue-on-failure tag instead of several Run Keyword And Continue On Failure keywords (disabled)

```

We could filter them out and list them if listed with special option, like 

```
robocop --list COMMUNITY
```

mnojek:
> Should community rules be filtered out (not visible) by default when listing them with `--list`?

I think not. By default, the `--list` option should only show built-in rules. Having a special name for the group of rules is a good idea. We can have a `COMMUNITY` to show all community rules, and we can also have `ALL` to show both built-in and community rules. Without the special name, it will just display the built-in ones.

bhirsz:
From the initial list:

> Add an example of the community rule with all necessary structure of the repository:
> 
> * place new rule under `robocop/checkers/community_rules` directory
> * use `1XXYY` ID as a format, where `1`is a flag for a community rule, `XX` is the checker category (do we need categories here?) and the `YY` is the rule ID (maybe we can just go with `1XXXX` where `XXXX` is just the rule ID?).
> * adjust existing test framework to support running tests for community rules
> * add test that checks if all community rules are disabled by default
> * generate rule documentation and add a new "Community Rules" entry in the docs in the side menu
> * ...?
> 
> This task is part of the #809 issue.

We have example of community rule and:

[...]

PR #918, opened 2023-07-15
Add first community rule

bhirsz:
Implements most of #899 except docs part. Community rules can be accessed in sphinx generation part by calling ``get_community_rules`` (internal rules are available through previous ``get_rules``).


robocop/checkers/__init__.py
```diff
@@ -290,4 +309,12 @@ def init(linter):
 def get_rules():
```
mnojek, line 309: Maybe we should call it `get_builtin_rules`?

robocop/checkers/community_rules/keywords.py
```diff
@@ -0,0 +1,89 @@
...
+from robocop.utils.misc import normalize_robot_name
+
+rules = {
+    "10001": Rule(
+        RuleParam(
+            name="max_time", default=0, converter=timestr_to_secs, desc="Maximum amount of time allowed in Sleep"
+        ),
+        rule_id="10001",
+        name="sleep-keyword-used",
+        msg="Sleep keyword with '{{ sleep_time }}' sleep time found",
```
mnojek, line 15: What do you think about using `duration time` instead of `sleep time`? (no strong opinion)

robocop/checkers/community_rules/keywords.py
```diff
@@ -0,0 +1,89 @@
...
+        severity=RuleSeverity.WARNING,
+        added_in_version="4.1.0",
```
mnojek, line 17:
```suggestion
        added_in_version="5.0.0",
```

robocop/checkers/community_rules/keywords.py
```diff
@@ -0,0 +1,89 @@
...
+        
+        Can be rewritten to::
+        
+            *** Keywords ***
+            Add To Cart
+                [Arguments]    ${item_name}
+                Wait Until Element Is Visible    ${MAIN_HEADER}
+                Click Element    //div[@name='${item_name}']/div[@id='add_to_cart']
+        
+        It is also possible to report only ``Sleep`` exceeding given time limit using ``max_time`` parameter::
```
mnojek, line 39:
```suggestion
        It is also possible to report only if ``Sleep`` exceeds given time limit using ``max_time`` parameter::
```

robocop/checkers/community_rules/keywords.py
```diff
@@ -0,0 +1,89 @@
...
+        
+            robocop -c sleep-keyword-used:max_time:1min .
+
+        """,
+    )
+}
+
+
+class ContinueOnFailureChecker(VisitorChecker):
```
mnojek, line 48: Update name of the checker.

tests/e2e/test_e2e.py
```diff
@@ -368,3 +368,10 @@ def test_all_rules_have_added_in_version_info(self, robocop_instance):
         robocop_instance.reload_config()
         without_version = {rule.name for rule in robocop_instance.rules.values() if not rule.added_in_version}
         assert without_version == set()
+
+    def test_all_community_rules_should_be_disabled_by_default(self, robocop_instance):
```
mnojek, line 372: Maybe the same for built-in rules but with enabled?

mnojek (commented):
Great job!

I am thinking that maybe we should also add a check for the rule ID?
For example, we should check:
- if community rule ID has `1` at the beginning and is 5-digit long
- if built-in rule ID is 4-digit long
- if the rule ID starts with a correct category ID (we can add a static category ID at the top of the file

bhirsz:
> Great job!
> 
> I am thinking that maybe we should also add a check for the rule ID? For example, we should check:
> 
> * if community rule ID has `1` at the beginning and is 5-digit long
> * if built-in rule ID is 4-digit long
> * if the rule ID starts with a correct category ID (we can add a static category ID at the top of the file

Good idea! Tests for category ID can also apply for built-in rules. Though we have two possibilities:

- define static category ID
or
- ensure that all IDs in the file use the same category ID (if first rule starts with 02, then second and third also should etc)

WDYT? The second option does not require defining any extra category ID but we can miss if someone uses incorrect category ID in all rules.

robocop/checkers/community_rules/keywords.py
bhirsz: I like it better, ``sleep time`` sounds quirky for me too

tests/e2e/test_e2e.py
bhirsz: I will try to implement such test. After I created the same type of the tests I have noticed that it also matches rules that are disabled due to version (ie ``nested-for-loop`` in RF 6). And I found the bug thanks for that - ``if-can-be-used`` for some reason is only enabled for 4.* and not for 5, 6 etc versions.

mnojek: Tests always reveal the true identity :D

mnojek:
> > Great job!
> > I am thinking that maybe we should also add a check for the rule ID? For example, we should check:
> > 
> > * if community rule ID has `1` at the beginning and is 5-digit long
> > * if built-in rule ID is 4-digit long
> > * if the rule ID starts with a correct category ID (we can add a static category ID at the top of the file
> 
> Good idea! Tests for category ID can also apply for built-in rules. Though we have two possibilities:
> 
> * define static category ID
>   or
> * ensure that all IDs in the file use the same category ID (if first rule starts with 02, then second and third also should etc)
> 
> WDYT? The second option does not require defining any extra category ID but we can miss if someone uses incorrect category ID in all rules.

I like the first option more because it protects from the case when someone creates a file with a new category but with already existing category ID (in another file). So I think having a static category ID is better. Also, it kinds of makes it more formal. Right now, the only reference to the category ID is in the rule ID itself.

PR #922, opened 2023-07-29
Fix loading non-python files in external rules

bhirsz:
I have discovered that Robocop loading mechanism tries to also to load non-Python files and it can throw fatal exception.


docs/releasenotes/5.0.0.rst
```diff
@@ -43,6 +43,11 @@ Enable ``if-can-be-used`` rule for Robot Framework 5+
 
 I0908 ``if-can-be-used`` was incorrectly enabled only for Robot Framework 4.*. It should now work for all >=4 versions.
 
+External rules with non-Python files in the directory
+-----------------------------------------------------
+
+Robocop should now be able to load external rules directory that contains non-Python files.
```
mnojek, line 49: I would rephrase to:

```suggestion
Robocop no longer reports a problem when loading the external rules from a directory that also contains non-Python files.
```

mnojek (approved):

PR #928, opened 2023-08-08
Fix variable-overwritten-before-usage using single scope for all TRY/EXCEPT branches

### LINKED ISSUE #925: [Bug] 0922 Local variable '${value}' is overwritten before usage (variable-overwritten-before-usage) warning when using TRY/EXCEPT
### What happened?

There seems to be an unnecessary "0922 Local variable '${value}' is overwritten before usage (variable-overwritten-before-usage)" warning when using TRY/EXCEPT.

`robocop Example.robot`

Example.robot:
```
*** Settings ***
Documentation     Incorrect 0922 warning example


*** Test Cases ***
Example Test
    [Documentation]    Incorrect 0922 warning example test
    TRY
        ${value}=    Randomly Failing Keyword    # In real use case something inside the TRY block may sometimes fail
    EXCEPT
        ${value}=    Set Variable    Keyword failed
    END
    Log To Console    ${value}


*** Keywords ***
Randomly Failing Keyword
    [Documentation]     This keyword randomly fails
    IF    random.randint(0, 1) == 1    Fail    ELSE    RETURN    Keyword succeeded
```

Example.robot:9:9 [W] 0922 Local variable '${value}' is overwritten before usage (variable-overwritten-before-usage)

I expected no 0922 warning because the variable ${value} is set either in the TRY block or in the EXCEPT block before use.

### Operating System

Windows

### Robocop version

4.0.1

bhirsz:
Fixes #925

### DIFF
--- a/docs/releasenotes/unreleased/fixes.3.rst
+++ b/docs/releasenotes/unreleased/fixes.3.rst
@@ -1,7 +1,7 @@
 TRY/EXCEPT with shared variables scope for all branches (#925)
 --------------------------------------------------------------
 
-``variable-overwritten-before-usage`` will no longer be raised  if the same variable names are used in different
+``variable-overwritten-before-usage`` will no longer be raised if the same variable names are used in different
 branches of the ``TRY/EXCEPT`` block::
 
     *** Test Cases ***
--- a/docs/releasenotes/unreleased/fixes.3.rst
+++ b/docs/releasenotes/unreleased/fixes.3.rst
@@ -1,0 +1,14 @@
+TRY/EXCEPT with shared variables scope for all branches (#925)
+--------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


docs/releasenotes/unreleased/fixes.3.rst
```diff
@@ -0,0 +1,14 @@
+TRY/EXCEPT with shared variables scope for all branches (#925)
+--------------------------------------------------------------
+
+``variable-overwritten-before-usage`` will no longer be raised  if the same variable names are used in different
```
mnojek, line 4:
```suggestion
``variable-overwritten-before-usage`` will no longer be raised if the same variable names are used in different
```

mnojek (approved):

PR #938, opened 2023-08-28
Fix invalid urls to rules in API and SARIF reports

### LINKED ISSUE #937: [Bug] issues_to_lsp_diagnostic returns a wrong link for codeDescription property
### What happened?

Just implementing the code description for `robocop` in `RobotCode`. For new versions of robocop, I think startet with 4.0 the  function [issues_to_lsp_diagnostic](https://github.com/MarketSquare/robotframework-robocop/blob/5187992f3ac675db3f20f75d0f5e5ab209d5979a/robocop/utils/misc.py#L74) returns a wrong link to the description of a rule.

- in robocop 4.0.0 you have included the severity of the rule in the anchor/link in `rules.html`. Because `issues_to_lsp_diagnostic` does not work correctly, I startet to build the link by myself, but if you reconfigure the severity of the rule, then it is not easy possible to get the original severity.

- in robocop 4.1.0 you changed the name of the rules.html to rules_list.html.

I think in SARIF protocol you get also a wrong link, but I did not check this.

 [issues_to_lsp_diagnostic](https://github.com/MarketSquare/robotframework-robocop/blob/5187992f3ac675db3f20f75d0f5e5ab209d5979a/robocop/utils/misc.py#L74)

the correct link to the rule description.

### Operating System

_No response_

### Robocop version

4.1, 4.0

bhirsz:
Fixes #937

### DIFF
--- a/docs/releasenotes/unreleased/fixes.1.rst
+++ b/docs/releasenotes/unreleased/fixes.1.rst
@@ -1,0 +1,11 @@
+Invalid links to rules after documentation refactor (#937)
+----------------------------------------------------------
+
+Fixed links to our rules after recent documentation refactor:
+
+- issues_to_lsp_diagnostic API method should now point to rules_list.html instead of rules.html
+- SARIF reports should now point to rules_list.html instead of rules.html
+- rule description now points to valid url
+
+Also, the SARIF reports will now contain explicit Robocop version in the rules url to decrease number of backward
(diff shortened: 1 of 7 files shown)


mnojek (approved):
I'm fine with this idea of having a version explicitly if it reduces the chance of errors.
Thanks for the fix :)

PR #961, opened 2023-10-07
Fix variable-overwritten-before-usage reported on overriding default values in IF

### LINKED ISSUE #950: [Bug] Rule variable-overwritten-before-usage should not be raised if overridden in an IF block
### What happened?

I am initiating a variable that I conditionally override in an IF block, but the warning variable-overwritten-before-usage is raised.

```robotframework
My Keyword
    [Documentation]  To reproduce the bug
    ${output}  Set Variable  ${TRUE}
    IF    os.path.isdir("/mydir")
        ${condition}  Set Variable  ${FALSE}  # here I have an actual keyword that performs an action an a check
        IF    $condition
            ${output}  Set Variable  ${FALSE}
        END
    END
    [Return]    ${output}
```

Warning variable-overwritten-before-usage is raised

Warning should not be raised if the variable is defined before the IF/ELSE block, and overridden in the IF or the ELSE
But if the variable is overridden in both IF and ELSE, then warning should be raised

### Operating System

_No response_

### Robocop version

4.1.0

bhirsz:
Fixes #950

### DIFF
--- a/docs/releasenotes/unreleased/fixes.4.rst
+++ b/docs/releasenotes/unreleased/fixes.4.rst
@@ -1,0 +1,17 @@
+variable-overwritten-before-usage should not be raised if overridden in an IF block (#950)
+------------------------------------------------------------------------------------------
+
+W0922 ``variable-overwritten-before-usage`` warns if the variable was overwritten before first use. It should now
+ignore cases where the variable was initiated and then conditionally overriden in a IF block:
+
+```
+Conditionally Overriden Variable
+    ${output}  Set Variable  default_value
+    IF    os.path.isdir("/special_dir")
+        ${output}  Set Variable  special_value  # it will not be reported from now on
+    END
+    RETURN    ${output}
+```
+
+Such cases could be also handled better by using `IF` with `ELSE` branch or `Set Variable If`. However, in more
+complicated case this rule produced lot of additional unnecessary reports and this fix should help with it.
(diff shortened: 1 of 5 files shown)


docs/releasenotes/unreleased/fixes.4.rst
```diff
@@ -0,0 +1,17 @@
+variable-overwritten-before-usage should not be raised if overridden in an IF block (#950)
```
mnojek, line 1:
```suggestion
variable-overwritten-before-usage should not be raised if overwritten in an IF block (#950)
```

docs/releasenotes/unreleased/fixes.4.rst
```diff
@@ -0,0 +1,17 @@
...
+------------------------------------------------------------------------------------------
+
+W0922 ``variable-overwritten-before-usage`` warns if the variable was overwritten before first use. It should now
+ignore cases where the variable was initiated and then conditionally overriden in a IF block:
```
mnojek, line 5:
```suggestion
ignore cases where the variable was initiated and then conditionally overwritten in an IF block:
```

docs/releasenotes/unreleased/fixes.4.rst
```diff
@@ -0,0 +1,17 @@
...
+
+```
+Conditionally Overriden Variable
```
mnojek, line 8:
```suggestion
Conditionally Overwritten Variable
```

mnojek (changes requested):

mnojek (approved):

PR #972, opened 2023-10-13
Fix variable with attributes detected as unused

### LINKED ISSUE #944: [Bug] argument-overwritten-before-usage does not recognize variables with attribute access
It should not report on following code:
```
Keyword
    [Arguments]    ${supplier}
    ${supplier}    Set Variable    ${supplier.lower()}
```

### LINKED ISSUE #971: [Bug] False positive "variable is assigned but not used" in case of extended variable assignment
### What happened?

image

MRE:
```robotframework
*** Settings ***
Library     DemoLib.py


*** Test Cases ***
DemoCase
    ${item}    Prepare Item
    Log    ${item.x}


*** Keywords ***
Prepare Item
    ${item}    Get Item
    ${item.x}    Set Variable    abc
    RETURN    ${item}
```

```python
from dataclasses import dataclass

@dataclass
class Item:
    x: str = ""

def get_item():
    return Item()
```

`robocop -i 0920 Demo.robot`

 `[I] 0920 Variable '${item.x}' is assigned but not used (unused-variable)`

Assigning to object attributes should not be marked as unused variable

### Operating System

_No response_

### Robocop version

4.1.1

bhirsz:
Fixes #971 
Fixes #944

I would call this fix as workaround since I'm not 100% happy with the current implementation - I will need to do some refactors soon. But it works and our tests are passing.

### DIFF
--- a/robocop/checkers/misc.py
+++ b/robocop/checkers/misc.py
@@ -3,6 +3,7 @@ Miscellaneous checkers
 """
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Optional
 
 from robot.api import Token
 from robot.errors import VariableError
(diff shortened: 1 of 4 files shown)


robocop/checkers/misc.py
```diff
@@ -1247,8 +1258,6 @@ def find_not_nested_variable(self, value, is_var):
                 self.find_not_nested_variable(match.base, is_var=True)
             for item in match.items:
                 self.find_not_nested_variable(item, is_var=False)
-        if not replaced:
```
bhirsz, line 1250: ;)

mnojek (approved):

PR #978, opened 2023-10-20
Fix: unused-variable if first declared as argument and used in the loop

### LINKED ISSUE #868: [Bug] Variables in loop are not detected as used

bhirsz:
Fixes #868 

Last fix for the release and we should be good to go - it's 12th already!


bhirsz:
I will merge it as it is so I can prepare next release PR - if you leave comments, I will fix it in release PR.

PR #984, opened 2023-10-23
Fix variables rules with global variables used in the loop before assigning variable

### LINKED ISSUE #868: [Bug] Variables in loop are not detected as used

bhirsz:
Fixes #868

It feels like a tip of iceberg with yet another bugs, but we constantly add new tests so my confidence in the solution slowly grows - and it can be good base for future refactor.

### DIFF
--- a/docs/releasenotes/unreleased/fixes.2.rst
+++ b/docs/releasenotes/unreleased/fixes.2.rst
@@ -1,0 +1,12 @@
+Global variables used in the loop before assigning variable (#868)
+------------------------------------------------------------------
+
+The variable rules (such as W0919 ``unused-argument`` or I0920 ``unused-variable``) were not aware of the recursive
+nature of the loops. For example if variable was used before it was declared in the loop (original value could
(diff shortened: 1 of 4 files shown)


docs/releasenotes/unreleased/fixes.2.rst
```diff
@@ -0,0 +1,12 @@
+Global variables used in the loop before assigning variable (#868)
+------------------------------------------------------------------
+
+The variable rules (such as W0919 ``unused-argument`` or I0920 ``unused-variable``) were not aware of the recursive
```
mnojek, line 4: I think that in the fix we should always include a description what's the result of the fix. You mention the problem that existed, but it should also state how it works now. Or at least that "it's now fixed".

mnojek (approved):

PR #989, opened 2023-10-29
Fix Inline If with assign variables handling variables in invalid order

### LINKED ISSUE #950: [Bug] Rule variable-overwritten-before-usage should not be raised if overridden in an IF block

### LINKED ISSUE #987: [Bug] Fix for #950 missing inline IF
### What happened?

Seems that the fix (https://github.com/MarketSquare/robotframework-robocop/issues/950) for variable-overwritten-before-usage does not cover inline IF.

```
KW
    [Arguments]  ${arg}
    ${arg}    IF  ${CONDITION}  Replace String  ${arg}  TAG  ${CONDITION_TAG}
```

0921 Keyword argument '${arg}' is overwritten before usage (argument-overwritten-before-usage)

No 0921 rule violation

### Operating System

Linux

### Robocop version

4.2.1

### LINKED ISSUE #988: [Bug]  Unused variable not raised if variable was overwritten in Inline If
Following code:

```

Inline If - Overwritten Variable
    ${var}    Set Variable    default
    ${var}    IF    condition    Use    ${var}
```

should raise ``unused-variable``

bhirsz:
Fixes #987
Fixes #988

### DIFF
--- a/docs/releasenotes/unreleased/fixes.1.rst
+++ b/docs/releasenotes/unreleased/fixes.1.rst
@@ -1,7 +1,7 @@
 Inline If with assign variables handling variables in invalid order (#987)
 ---------------------------------------------------------------------------
 
-Inline If did not properly recognized that variables were used inside if loop before assigning return value.
+"Inline If" did not properly recognize that variables were used inside "if statement" before assigning return value.
 Following code::
 
     *** Keywords ***
--- a/docs/releasenotes/unreleased/fixes.1.rst
+++ b/docs/releasenotes/unreleased/fixes.1.rst
@@ -1,0 +1,28 @@
+Inline If with assign variables handling variables in invalid order (#987)
+---------------------------------------------------------------------------
+
+Inline If did not properly recognized that variables were used inside if loop before assigning return value.
+Following code::
+
+    *** Keywords ***
+    Keyword With Argument
+        [Arguments]    ${arg}
+        ${arg}    IF    ${VALUE}    Use  ${arg}
+
+    Keyword With Local Variable
+        ${var}    Set Variable    default value
+        ${var}    IF    ${VALUE}    Use  ${var}
+
+Will no longer raise W0921 ``argument-overwritten-before-usage`` and W0922 ``variable-overwritten-before-usage``.
+
+And following code::
+
+    *** Keywords ***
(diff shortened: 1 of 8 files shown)


tests/atest/utils/__init__.py
```diff
@@ -118,14 +118,14 @@ def check_rule(
...
             missing_actual = sorted(set(expected) - set(actual))
-            present_in_actual = "\n    ".join(missing_expected)
-            present_in_expected = "\n    ".join(missing_actual)
-            raise AssertionError(
-                "Actual issues are different than expected.\n"
-                f"Actual issues not found in expected:\n    {present_in_actual}"
-                f"\n\nExpected issues not found in actual:\n    {present_in_expected}"
-            )
-        assert actual == expected, f"{actual} != {expected}"
+            error = "Actual issues are different than expected.\n"
```
bhirsz, line 121: This change improves readability of errors in atests - it will not display full trace now, just the difference between actual and expected issues.

docs/releasenotes/unreleased/fixes.1.rst
```diff
@@ -0,0 +1,28 @@
+Inline If with assign variables handling variables in invalid order (#987)
+---------------------------------------------------------------------------
+
+Inline If did not properly recognized that variables were used inside if loop before assigning return value.
```
mnojek, line 4:
```suggestion
"Inline If" did not properly recognize that variables were used inside "if statement" before assigning return value.
```

mnojek (approved):

PR #1021, opened 2023-12-26
Refactor Github actions to run on all OS, Python and RF versions

bhirsz:
This is test only so far - I came up with an idea how to run our tests on combination of all OS/Python/RF version without spending too much time (though it will take longer) and using too much of Github resources.

The major downside right now is logging - if the test fail for specific Python/RF version it may be harder to find it. I need to test if it's still okay.

Edit. It seems the builds are taking really long time..it's essentially running Py3.8-Py3.12 * RF3 - RF7 = 5 * 5 = 25 builds inside one docker * 3 OS. We could introduce it as separate action to run after the release/or on release branch before merging - long but covering all possible versions etc. And run current action for 'common' PRs. With that approach we could 'thin' the original run a bit (don't run on Windows, run on every RF version but do not cover all Py versions etc) - since those combinations rarely fail and take longer time. And we will run it anyway but just once before release.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -15,31 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - os: 'ubuntu-latest'
-            python-version: '3.8'
-            rf-version: 'rf3'
-          - os: 'ubuntu-latest'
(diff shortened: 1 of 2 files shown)


mnojek:
Makes sense to run Robocop for all possible combinations at least once before the release. Nice idea.

PR #1024, opened 2023-12-27
Handle RF development version as final version

bhirsz:
With previous implementation there was problem where some of our tests didn't run at all. For example if we used ">=7" version specifier, RF "7.0a1" didn't match it. I have decided to just drop support for development versions in version specifier and handle those as 'final' versions. Parsed version (the one from RF) can have development suffix (dev1, rc1..) but it will be ignored.

### DIFF
--- a/tests/atest/rules/errors/parsing_error/test_rule.py
+++ b/tests/atest/rules/errors/parsing_error/test_rule.py
@@ -3,7 +3,7 @@ from tests.atest.utils import RuleAcceptance
 
 class TestRule(RuleAcceptance):
     def test_rule(self):
-        self.check_rule(expected_file="expected_output.txt", target_version=">=7")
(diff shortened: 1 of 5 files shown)


tests/atest/rules/errors/parsing_error/test_rule.py
```diff
@@ -3,7 +3,7 @@
 
 class TestRule(RuleAcceptance):
     def test_rule(self):
-        self.check_rule(expected_file="expected_output.txt", target_version=">=7")
+        self.check_rule(expected_file="expected_output_rf7.txt", target_version=">=7")
```
bhirsz, line 6: Example of test that didn't run for 7.0a1 -> 7.0rc1 and this typo wasn't discovered

PR #1036, opened 2023-12-30
Allow for multiple RF versions in rule rf version specifier

### LINKED ISSUE #1026: Rule version specifier should allow for multiple versions
If the rule is only valid for set of versions, ie rule was introduced in RF 5 but will not make sense after RF 8:

```
version=">=5;<8"
```

We can use ";" in version specifier to separate different versions. Rule will be only enabled if it matches all conditions from the list.

bhirsz:
Closes #1026

### DIFF
--- a/robocop/utils/version_matching.py
+++ b/robocop/utils/version_matching.py
@@ -282,4 +282,4 @@ class VersionSpecifier:
         return prospective < Version(spec_str)
 
     def _compare_greater_than(self, prospective, spec_str: str) -> bool:
-        return prospective <= Version(spec_str)
+        return prospective > Version(spec_str)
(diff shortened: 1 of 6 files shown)


robocop/utils/version_matching.py
```diff
@@ -282,4 +282,4 @@ def _compare_less_than(self, prospective, spec_str: str) -> bool:
         return prospective < Version(spec_str)
 
     def _compare_greater_than(self, prospective, spec_str: str) -> bool:
-        return prospective <= Version(spec_str)
+        return prospective > Version(spec_str)
```
bhirsz, line 285: Even found a bug thanks to new tests..

Issue #973, opened 2023-10-13
Handle new VAR syntax


bhirsz:
Next Robot Framework release may bring new variable syntax (in place of Set Variable). We will need to update several rules (especially those checking the variable scope). 

Refer to https://github.com/robotframework/robotframework/issues/3761

bhirsz:
This issues splits into two:

1) Update existing rules to handle VAR
2) Add new rule(s) to warn on use of ``Set Local/Test/Task/Suite/Global Variable``

bhirsz:
Since VAR is additional syntax, it's not counted as keyword call for test/keyword call counts rules:

Test case 'Overwrite reserved with VAR' has too few keywords inside (0/1) (too-few-calls-in-test-case)

We should support count it as keyword call in such case.

bhirsz:
Also need to updated rule that checks variable definition for missing value. Following code is accepted:

```
VAR    ${scalar}
Log    ${scalar}  # empty string
VAR    @{list}
Log    ${list}  # empty list
```

We should suggest using explicit values (${EMPTY}, @{EMPTY}). Though I'm not sure since ``VAR  ${scalar}`` looks compact

bhirsz:
And still not sure about suggesting to replace ``Create List`` / ``Create Dictionary`` with ``VAR``. But maybe it makes sense and makes code cleaner?

```
Create List    ${list}    value1    value2
VAR   @{list}    value1    value2
```

bhirsz:
> And still not sure about suggesting to replace `Create List` / `Create Dictionary` with `VAR`. But maybe it makes sense and makes code cleaner?
> 
> ```
> Create List    ${list}    value1    value2
> VAR   @{list}    value1    value2
> ```

I will introduce two new rules:

``replace-set-variable-with-var`` (or similiar name) - it can be replaced with ``deprecated-statement`` in the future
``replace-create-with-var`` (or similiar name) - for Create List/Dictionary

bhirsz:
Rules that still need ``VAR`` support: ``ossible-variable-overwriting``, ``inconsistent-variable-name``

bhirsz:
Also check how Catenate could be replaced with VAR and if it does make sense to add rule for it (replace-catenate-with-var)

bhirsz:
All tasks are resolved. The only thing left is handling new parsing errors (such as invalid variable name in VAR) but I'm leaving it for now as we have other task open for revisiting all possible parsing-errors .

PR #1038, opened 2023-12-30
Add support for VAR in empty-variable rule

bhirsz:
Relates #973 

One syntax but requires so many updates to rules :)


borutzki:
Here, option to disable this rule for `VAR` and leave it for `*** Variables ***` section would be useful.

bhirsz:
> Here, option to disable this rule for `VAR` and leave it for `*** Variables ***` section would be useful.

Sounds good, can you create feature request? Just an issue with what you wrote here will suffice

borutzki:
Yup, I'll do it. I've just noticed the notification

PR #1047, opened 2024-01-15
Fix incorrect trigger of W0310

### LINKED ISSUE #1045: [Bug] W0310 incorrectly triggered with VAR syntax
### What happened?

The rule `non-local-variables-should-be-uppercase` seems to have issues with the new VAR syntax of Robot Framework 7. It's triggered although the variable is uppercase.

```VAR    ${CORRELATIONID}    ${response.json()}[correlationId]    scope=SUITE```

[W] 0310 Test, suite and global variables should be uppercase (non-local-variables-should-be-uppercase)

The warning is not shown.

### Operating System

Windows 10

### Robocop version

5.0.0

### PR DESCRIPTION
**Pull request by Shutgun**
Fixes #1045 

However, I'm not sure if this is the best possible solution.

### DIFF
--- a/robocop/checkers/naming.py
+++ b/robocop/checkers/naming.py
@@ -1094,7 +1094,7 @@ class VariableNamingChecker(VisitorChecker):
         self.check_for_reserved_naming_or_hyphen(variable, "Variable", is_assign=True)
         # TODO Check supported syntax for variable, ie ${{var}}?
         if not self._is_var_scope_local(node):
-            self.check_non_local_variable(variable.value, node, variable)
+            self.check_non_local_variable(search_variable(variable.value).base, node, variable)
 
     def visit_If(self, node):  # noqa
         for token in node.header.get_tokens(Token.ASSIGN):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: robocop/checkers/naming.py [hunk 1]
```diff
@@ -1094,7 +1094,7 @@ def visit_Var(self, node):  # noqa
         self.check_for_reserved_naming_or_hyphen(variable, "Variable", is_assign=True)
         # TODO Check supported syntax for variable, ie ${{var}}?
         if not self._is_var_scope_local(node):
-            self.check_non_local_variable(variable.value, node, variable)
+            self.check_non_local_variable(search_variable(variable.value).base, node, variable)
```
bhirsz, line 1097: search_variable may fail if the variable is invalid (for example not closed properly {}). We can just wrap it with try except as we did it in visit_Variable in the same file to safeguard it

bhirsz: Ok nvm, we have return if node contains any error just few lines before - it should cover it

bhirsz:
I have checked and it works with your code - it looks like this logic was invalid and missing search_variables resulted in getting empty string on isupper() comparison later on. Your code fixes that - thanks!

bhirsz (approved):

robocop/checkers/naming.py
Shutgun: It's that covered by the returns in lines 1090 and 1093?

Issue #107, opened 2020-08-25
Rule: check if dependencies are sorted


mnojek:
Check if dependencies in Settings section are sorted: Libraries first, then Resource files. Can also check if they are sorted alphabetically or whatever way we decide to implement it.

bhirsz:
It could be also possible to test if the import is local or external and propose sorting based on that (similiar as python):

```
Library  Collections  # internal
Library  common.py  # in the same directory
Library  MyTestLib  # external library
```

mnojek:
> It could be also possible to test if the import is local or external and propose sorting based on that (similiar as python):
> 
> ```
> Library  Collections  # internal
> Library  common.py  # in the same directory
> Library  MyTestLib  # external library
> ```

We already have this inside 'wrong-import-order' rule but without checking libraries from the same directory.

mnojek:
Also, I am thinking if this is a good idea to implement this rule. I remember a discussion on slack (probably on #lsp or #devel channel) about the order of the imports and how they relate to each other. The conclusion was that sometimes different order is required due to specific variables relying on each other. So forcing user to keep one and only order is probably not a good idea...
@bhirsz What do you think?

bhirsz:
Yeah, resources and libraries are often depending in the order of import which would be contradicting with this rule. And like you wrote there is w ring-import-order already. We can close it

mnojek:
Another thing that comes to my mind is to create a rule for keeping settings of the same type grouped together, but again... I am not 100% sure if it will not trigger some problems with import order. @bhirsz What do you think? Can we implement such rule or we should give up with all sort of these rules related to order of imports?

bhirsz:
If we don't have anything for it yet (I don't think so) it could be good idea. Robotidy already group settings, for example result [here](https://github.com/MarketSquare/robotframework-tidy/blob/main/tests/atest/transformers/OrderSettingsSection/expected/test.robot). Setting order other than imports (variables, resources, libraries) doesn't matter in execution. It looks better if let's say we're keeping tag settings together with other tag settings so I'm vouting for such rule.

bhirsz:
Users recently added several rules related to import orders - in 6.0 and before:
https://github.com/MarketSquare/robotframework-robocop/blob/main/src/robocop/linter/rules/imports.py

I believe they satisfy rules mentioned in this issue so we can close it (after 5 years! :))

bhirsz:
[...]

PR #1088, opened 2024-06-04
Add rules checking that builtin and custom libraries are imported in alphabetical order.

szymonslodkowski:
Add rules if builtin and custom libraries are sorted alphabetically


bhirsz:
Thanks for the good PR. In a way it implements https://github.com/MarketSquare/robotframework-robocop/issues/107 . 
Though there is caveat here: order of the custom imports may be important (and it often is. For example you need to import B before you import A because B keeps variables you need for A). However it should be fine for built in imports. So the solution could be moving those rules (or at least the custom imports one) to community rules (which are disabled by default and can be enabled). What do you think about it?

robocop/checkers/misc.py
```diff
@@ -586,6 +586,35 @@
         """,
         added_in_version="4.0.0",
     ),
+    "0926": Rule(
+        rule_id="0926",
+        name="builtin-imports-not-sorted",
+        msg="BuiltIn library import '{{ builtin_import }}' should be placed before '{{ previous_builtin_import }}'",
+        severity=RuleSeverity.WARNING,
```
bhirsz, line 593:
```suggestion
        severity=RuleSeverity.WARNING,
        added_in_version="5.2.0",
```

robocop/checkers/misc.py
```diff
@@ -586,6 +586,35 @@
...
+            Library    OperatingSystem
+            Library    Collections  # BuiltIn libraries imported not in alphabetical order
+
+        """,
+    ),
+    "0927": Rule(
+        rule_id="0927",
+        name="custom-imports-not-sorted",
+        msg="Custom library import '{{ custom_import }}' should be placed before '{{ previous_custom_import }}'",
+        severity=RuleSeverity.WARNING,
```
bhirsz, line 607:
```suggestion
        severity=RuleSeverity.WARNING,
        added_in_version="5.2.0",
```

szymonslodkowski:
@bhirsz thanks for quick review.
So I will leave the rule for builtins in the file I already implemented.
I was not aware about use cases in which order of imports is important - good to know that there are some. So  I will move rule checking rest of imports to custom rules.
Maybe there is also sense to implement rule verifying order of resources? I could try also to add such rule in this PR. Is it ok?

bhirsz:
Yeah, feel free to do so - since it will be also custom rule then we can have such rule for resources as well (people can opt in for it).

szymonslodkowski:
Ok, but I have a question about community rules numbering.
In robocop/checkers/community_rules there is only keywords.py with `RULE_CATEGORY_ID = "00"` I guess these new rules doesn't fit to this file. I would like to create new one, but what should be RULE_CATEGORY_ID and what should be rule_id?

bhirsz:
You can decide it. But since we started with 00, then next one will be 01. Community rules are prefixed with 1 so final rule will be for example 10101, 10102...

robocop/checkers/misc.py
szymonslodkowski: Added

robocop/checkers/misc.py
szymonslodkowski: Added

bhirsz (approved):

Issue #997, opened 2023-11-23
Check order of settings in test cases and keywords


bhirsz:
According to https://docs.robotframework.org/docs/style_guide/guide#keyword

bhirsz:
It will be released in 5.3.0

PR #1090, opened 2024-06-11
Implement rules checking order of elements in keyword and test case

szymonslodkowski:
Implement rules verifying order of elements in keyword and test case.


szymonslodkowski:
I believe in this  PR I implemented what in scope of #997

robocop/checkers/misc.py
```diff
@@ -601,6 +607,78 @@
...
+        RuleParam(
+            name="sections_order",
+            default="documentation,tags,timeout,setup,template,keyword,teardown",
+            converter=parse_test_case_order_param,
+            show_type="str",
+            desc="order of sections in comma-separated list",
+        ),
+        rule_id="0927",
+        name="test-case-section-out-of-order",
+        msg="'{{ section_name }} is in wrong place of Test Case. "
```
bhirsz, line 620: Missing closing '
```suggestion
        msg="'{{ section_name }}' is in wrong place of Test Case. "
```

robocop/checkers/misc.py
```diff
@@ -601,6 +607,78 @@
 
         """,
     ),
+    "0927": Rule(
+        RuleParam(
+            name="sections_order",
+            default="documentation,tags,timeout,setup,template,keyword,teardown",
+            converter=parse_test_case_order_param,
+            show_type="str",
```
bhirsz, line 615: You have used most of our misc stuff for creating rules - good to see that others can do it :) (meaning it's not dead code)

robocop/checkers/misc.py
```diff
@@ -601,6 +607,78 @@
...
+        ),
+        rule_id="0927",
+        name="test-case-section-out-of-order",
+        msg="'{{ section_name }}' is in wrong place of Test Case. "
+        "Recommended order of elements in Test Cases: {{ recommended_order }}",
+        severity=RuleSeverity.WARNING,
+        added_in_version="5.3.0",
+        docs="""
+        Sections should be defined in order set by ``sections_order``
+        parameter (default: ``documentation,tags,timeout,setup,template,keyword,teardown``).
```
bhirsz, line 626: keyword stands also for any other code like IF, FOR, TRY, comments etc? maybe it should be ``keyword_body``? Or it would be too descriptive

robocop/checkers/misc.py
```diff
@@ -601,6 +607,78 @@
...
+                [Documentation]    This is test Documentation
+                [Tags]    tag1    tag2
+                [Teardown]    Log    abc
+                Keyword1
+        """,
+    ),
+    "0928": Rule(
+        RuleParam(
+            name="sections_order",
+            default="documentation,tags,arguments,timeout,setup,keyword,teardown",
```
bhirsz, line 649: It may be now obvious when you did it, but I like that you're using single order for before / body / after. In Robotidy for corresponding transformer that does order the settings I have before / after parameters. It would be better to just use single parameter like you did - I may do it in the future

robocop/checkers/misc.py
```diff
@@ -1566,3 +1644,47 @@ def check_for_complex_condition(self, condition_token, node_name, left_side, var
...
+    def visit_File(self, node):  # noqa
+        self.rules_by_node_type = {Keyword: "keyword-section-out-of-order", TestCase: "test-case-section-out-of-order"}
+        self.expected_order = {
+            Keyword: self.param("keyword-section-out-of-order", "sections_order"),
+            TestCase: self.param("test-case-section-out-of-order", "sections_order"),
+        }
+        self.generic_visit(node)
+
+    def check_order(self, node):
+        if type(node) not in self.rules_by_node_type:
```
bhirsz, line 1666: This code may not be needed. If you're only calling this method in visit_Keyword and visit_TestCase, the node type will be always either Keyword and TestCase.

robocop/checkers/misc.py
```diff
@@ -1566,3 +1644,47 @@ def check_for_complex_condition(self, condition_token, node_name, left_side, var
...
+                    self.report(
+                        self.rules_by_node_type[type(node)],
+                        section_name=subnode_type,
+                        recommended_order=", ".join(self.expected_order[type(node)]),
+                        node=subnode,
+                        col=subnode.col_offset + 1,
+                        end_col=subnode.end_col_offset + 1,
+                    )
+                    break
+            already_used_sections.append(subnode_type)
```
bhirsz, line 1688: Since we're appending to already_used_sections for each element in node.body, all iterations (for example for keyword that has 5 items) would look like:

1 item:
  no iterations since already_used_sections = 0
 2nd item:
   1 iterations
  3nd item
    2 iterations 
    
    And so on - and we are checking the order of items that were already checked. It could be good idea to extract used_sections in one for loop, and then check the order:
    
 for subnode in node.body:
    # get type if no exception and in expected_order
    # save it to other list
 for item in other list:
     check order here

robocop/utils/misc.py
```diff
@@ -378,3 +378,39 @@ def _is_var_scope_local(node) -> bool:
...
             is_local = option.value.lower() == "scope=local"
     return is_local
+
+
+def parse_order_comma_sep_list(value: str, mapping: dict) -> list:
+    ordered = []
+    for item in value.split(","):
+        item_lower = item.lower()
+        if item_lower not in mapping:
+            raise ValueError(f"Invalid value: {item}. Supported values: {', '.join(mapping.keys())}")
```
bhirsz, line 388: We are missing negative test scenario for this code. 

But I couldn't find such test easily in Robocop (I have usually done them in Robotidy). I will consider it as task for me in the future, leaving story here: https://github.com/MarketSquare/robotframework-robocop/issues/1091 (feel free to resolve this comment without doing anything)

robocop/checkers/misc.py
szymonslodkowski: Frankly I had problem with theses statements like FOR, IF: they are not type of Token.KEYWORD and they don't have `type` property - this is why I added:
```
            try:
                subnode_type = subnode.type
            except AttributeError:
                continue
```
I was thinking how to handle these statements and I did assumption that in body of TRY / IF / FOR typically there is some keyword, so for example in case of such robot code:
```
[Teardown]    Cleanup
IF    ${something} == ${TRUE}
    Some Keyword
ELSE
    Some Other Keyword
END
```
Lines with with `IF`, `ELSE` will be not reported by the checker but lines inside: `    Some Keyword` and `    Some Other Keyword` will be reported.
Is this enough?

robocop/checkers/misc.py
szymonslodkowski: You are right, this was far away from optimal.
I decided to rework this in a bit different way: without list of already used sub-nodes, but only to remember maximal order indicator of already used sub-nodes. Thanks to that even second loop is not needed.

robocop/checkers/misc.py
bhirsz: Yes, it's good enough. Not always IF/FOR etc blocks contains keywords (it could be statement like RETURN or different block) but this simplification is okay for our purpose

robocop/checkers/misc.py
bhirsz: There is one downside of using one flag for it - if for example one setting is out of order, it may report invalid order for whole body (ie 20 lines) instead of reporting that single setting which is out of order. It could be workarounded by reporting this rule only once per test/keyword (but if several settings are out of order, it will hide it)

szymonslodkowski: Results returned by current and previous implementation are the same. Every test's / keyword's line is reported if is after one of previous lines when when should be before. `break` in second loop in previous implementation was only to not report that given line is after multiple lines.
To reduce number of reported issues we could:
- report this rule only once per test/keyword as you suggested.
- find issues when iterating through node.body and remember all issues, then do similar but by iterating reversed node.body), then compare number of issues found in each direction and call self.report() for shorter set of issues.
or we can leave it.
Please decide

bhirsz (approved):

Issue #1001, opened 2023-11-23
Keywords with split arguments: first argument in the same line


bhirsz:

bhirsz:
Implemented in linked PR

PR #1092, opened 2024-06-14
Add rule checking that first Argument is in the same line as [Arguments]

szymonslodkowski:
Implements rule from #1001

### DIFF
--- a/robocop/checkers/spacing.py
+++ b/robocop/checkers/spacing.py
@@ -286,6 +286,23 @@ rules = {
         """,
         added_in_version="3.0.0",
     ),
+    "1018": Rule(
+        rule_id="1018",
+        name="first-argument-in-new-line",
+        msg="First argument: '{{ argument_name }}' should be placed on the same line as [Arguments] setting",
+        severity=RuleSeverity.WARNING,
+        added_in_version="5.3.0",
+        docs="""
+        Example of rule violation::
+
+            *** Keywords ***
+            Custom Keyword With Five Required Arguments
+            [Arguments]
+            ...    ${name}
+            ...    ${surname}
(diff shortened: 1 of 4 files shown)


robocop/checkers/spacing.py
```diff
@@ -948,3 +965,31 @@ def parse_error(self, node, error):
                         col=indent + 1,
                     )
                     break
+
+
+class ArgumentsChecker(VisitorChecker):
+    reports = ("first-argument-in-new-line",)
+
+    def __init__(self):
```
bhirsz, line 973: If we don't change anything init init, we don't need to call it

robocop/checkers/spacing.py
```diff
@@ -948,3 +965,31 @@ def parse_error(self, node, error):
...
+        super().__init__()
+
+    def visit_File(self, node):
```
bhirsz, line 976: Parent class has visit_File that already does that - no point in overriding it if we don't add any extra code

robocop/checkers/spacing.py
```diff
@@ -948,3 +965,31 @@ def parse_error(self, node, error):
...
+        self.generic_visit(node)
+
+    def visit_Arguments(self, node):
+        eol_already = None
+        for t in node.tokens:
+            if t.type == Token.EOL:
+                eol_already = t
+                continue
+            if t.type == Token.ARGUMENT:
```
bhirsz, line 985:
```suggestion
            elif t.type == Token.ARGUMENT:
```

bhirsz:
Thanks, good work! Commented few nitpicks

bhirsz (approved):

PR #1102, opened 2024-07-25
Documentation fix: Changed `unused-keyword` introduction version to 5.3.0

Lakitna:
The community rule `unused-keyword` claims to be added in 5.1.0, but it was actually added in 5.3.0.

This is a docs-only change


bhirsz (approved):
Thanks!

PR #1137, opened 2024-10-30
Rules `no-global-variable`, `no-suite-variable`, and `no-test-variable`

### LINKED ISSUE #1117: Rule: No test variable, no suite variable, and no global variable
I'm working with a pretty big robot project where they used to constantly use `Set Test Variable` (and similar) instead of returning values. This causes significant confusion and issues when refactoring. I've since shown them why using returned values is a better idea, but there is still legacy code lurking in the depths.

What do you think of adding the following rules?

`no-test-variable`
Disallows setting Test/Task-level variables. Using `Set Test Variable`, `Set Task Variable`, `VAR  ...  scope=TEST`, and `VAR  ...  scope=TASK` is not allowed outside the variables section.

`no-suite-variable`
Disallows setting Suite-level variables. Using `Set Suite Variable`, `VAR  ...  scope=SUITE`, and `VAR  ...  scope=SUITES` is not allowed outside the variables section.

`no-global-variable`
Disallows setting Test/Task-level variables. Using `Set Global Variable` and `VAR  ...  scope=GLOBAL` is not allowed outside the variables section.

These rules would ignore variables in the variables section.

I'm willing to write the rules. Let me know if you're interested in adding them

Lakitna:
Closes #1117 

I'm not 100% sure about the tone of voice for the docs 🤔


robocop/checkers/misc.py
```diff
@@ -1725,3 +1870,83 @@ def check_order(self, node):
...
+
+        if scope == "GLOBAL":
+            self.report(
+                "no-global-variable",
+                node=option_token,
+                lineno=option_token.lineno,
+                col=option_token.col_offset + 1,
+                end_col=option_token.col_offset + len(option_token.value) + 1,
+            )
+
```
bhirsz, line 1935: typically if ``scope`` matches any of those, we can safely return without checking of other conditions as it will never match:

```
if scope == "GLOBAL":
   ...
   return
```

robocop/checkers/misc.py
```diff
@@ -1725,3 +1870,83 @@ def check_order(self, node):
...
+        if ROBOT_VERSION.major < 7:
+            return
+
+        if not node.scope:
+            return
+
+        option_token = node.get_token(Token.OPTION)
+        scope = node.scope.upper()
+
+        if scope == "GLOBAL":
```
bhirsz, line 1927: Can we assume that most popular variable scope if local scope (hopefully :)). Given that, we can test for it (scope=local or lack thereof) and return from this method early.

robocop/checkers/misc.py
```diff
@@ -1725,3 +1870,83 @@ def check_order(self, node):
...
+                lineno=keyword_token.lineno,
+                col=keyword_token.col_offset + 1,
+                end_col=keyword_token.col_offset + len(keyword_token.value) + 1,
+            )
+
+    def visit_Var(self, node):  # noqa: N802
+        if ROBOT_VERSION.major < 7:
+            return
+
+        if not node.scope:
```
bhirsz, line 1921: what about also leaving early on scope = 'local'?

robocop/checkers/misc.py
```diff
@@ -699,6 +699,151 @@ def comma_separated_list(value: str) -> list[str]:
...
+            *** Test Cases ***
+            My Amazing Test
+                Do A Thing
+                Should Be Equal    ${hello}    Hello, world!
+
+            *** Keywords ***
+            Do A Thing
+                VAR    ${hello}    Goodnight, moon!    scope=GLOBAL
+
+        There are exceptions. In some specific situations, global variables are a great tool. But
```
bhirsz, line 749: I would get rid of first sentence here, as it have the same meaning as next sentence (in this particular case)

```suggestion
        In some specific situations, global variables are a great tool. But
```

robocop/checkers/misc.py
```diff
@@ -1725,3 +1870,83 @@ def check_order(self, node):
...
+    )
+
+    def visit_KeywordCall(self, node: KeywordCall):  # noqa: N802
+        keyword_token = node.get_token(Token.KEYWORD)
+
+        if not keyword_token:
+            return
+
+        keyword_name = normalize_robot_name(keyword_token.value, remove_prefix="builtin.")
+
```
bhirsz, line 1889: We could speed up the look up for 'invalid' names here by grouping them under single set:

```
GLOBAL_VARIANTS = {'setglobalvariable', ..}

(...)

    if keyword_name not in self.GLOBAL_VARIANTS:  # O(1) look up time, and most keywords with return here
        return
    # rest of the existing code here
```

tests/atest/rules/misc/no_test_variable/test_var_syntax.robot
```diff
@@ -0,0 +1,15 @@
+*** Variables ***
+${amazing}    Hello world!
+
+*** Test Cases ***
+Test
+    VAR    ${foo}    bar
+    VAR    ${lorum}    ipsum    scope=LOCAL
+    VAR    ${amazing}    Hello universe    scope=TEST
+
```
bhirsz, line 9: Can you add some potential errors / missing stuff? For example:

```
    VAR
    VAR    $escaped_not_allowed
    VAR    ${name}    scope=typo
```

There is also dynamic scope (scope=${scope}) but we can't do anything about such abonominations

bhirsz:
Thanks for the PR, look good enough! The tone in the docs looks fine, though I'm not best doc writer .. :) If needed we can always improve it over time. I have left some comments but it's mostly nitpicks or improvements ideas (some stuff I've noticed I could also do with my rules, it's easier to see it if I look into someone else code).

robocop/checkers/misc.py
```diff
@@ -1725,3 +1870,83 @@ def check_order(self, node):
...
+            self.report(
+                "no-test-variable",
+                node=keyword_token,
+                lineno=keyword_token.lineno,
+                col=keyword_token.col_offset + 1,
+                end_col=keyword_token.col_offset + len(keyword_token.value) + 1,
+            )
+
+    def visit_Var(self, node):  # noqa: N802
+        if ROBOT_VERSION.major < 7:
```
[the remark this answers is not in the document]
Lakitna: No there is not. 

I'll change the `if` statement for a comment.

robocop/checkers/misc.py
Lakitna: I made the set in the checker class. I think moving it outside the class will cause readability issues.

bhirsz: Good, I didnt outline it but it was supposed to be in the checker class so good choice

tests/atest/rules/misc/no_test_variable/test_var_syntax.robot
Lakitna: Added those and a bunch more weirdness. I love abusing `${{ }}` evals in these situations.

robocop/checkers/misc.py
Lakitna: I assumed you meant a global variable due to the `SCREAMING_SNAKE_CASE` syntax you used :)

Lakitna:
Those where some good suggestions :)

I think I covered them all

PR #1138, opened 2024-11-02
Feat/explicitly define help url for each rule

### LINKED ISSUE #1123: [Bug] Help URL for custom rules links to Robocop docs
### What happened?

I use a bunch of custom lint rules. When generating a Sarif report, the value of `helpUri` always refers to the RoboCop docs. Even for custom lint rules.

Example:

Note the contents of the `helpUri` field.

```json
{
    "id": "CUSTOM-01",
    "name": "my-custom-lint-rule",
    "helpUri": "https://robocop.readthedocs.io/en/5.4.0/rules_list.html#my-custom-lint-rule",
    "shortDescription": {
        "text": "My amazing description"
    },
    "fullDescription": {
        "text": ""
    },
    "defaultConfiguration": {
        "level": "warning"
    },
    "help": {
        "text": "",
        "markdown": ""
    }
},
```

robocop . --reports sarif

None

I expect one of:

- Custom lint rules don't populate the `helpUri` field
- I can control the contents of the `helpUri` field manually

The first solution might be simpler to implement and would reduce confusion. The second solution is preferable since we have custom lint rule docs I'd like to reference.

### Operating System

Windows

### Robocop version

5.4.0

Lakitna:
Closes #1123

~~This is a proof of concept for explicitly defining the help URL for each rule. This approach will require all default and community rules to be extended with a URL.~~

Fixes rule help URLs for community rules and custom rules. URLs are now defined per rule instead of globally.

This is a backward compatible change.

### DIFF
--- a/robocop/checkers/comments.py
+++ b/robocop/checkers/comments.py
@@ -7,7 +7,7 @@ from robot.api import Token
 from robot.utils import FileReader
 
 from robocop.checkers import RawFileChecker, VisitorChecker
-from robocop.rules import Rule, RuleParam, RuleSeverity
+from robocop.rules import DefaultRule, RuleParam, RuleSeverity
 from robocop.utils import ROBOT_VERSION
 
 
@@ -22,7 +22,7 @@ def regex(value):
 RULE_CATEGORY_ID = "07"
 
 rules = {
-    "0701": Rule(
+    "0701": DefaultRule(
         RuleParam(
             name="markers",
             default="todo,fixme",
@@ -49,7 +49,7 @@ rules = {
         """,
         added_in_version="1.0.0",
     ),
-    "0702": Rule(
+    "0702": DefaultRule(
         RuleParam(
             name="block",
             default="^###",
@@ -87,7 +87,7 @@ rules = {
         """,
         added_in_version="1.0.0",
     ),
-    "0703": Rule(
+    "0703": DefaultRule(
         rule_id="0703",
         name="invalid-comment",
         msg="Invalid comment. '#' needs to be first character in the cell. "
@@ -107,7 +107,7 @@ rules = {
         """,
         added_in_version="1.0.0",
     ),
-    "0704": Rule(
+    "0704": DefaultRule(
         rule_id="0704",
         name="ignored-data",
         msg="Ignored data found in file",
@@ -136,7 +136,7 @@ rules = {
         """,
         added_in_version="1.3.0",
     ),
-    "0705": Rule(
+    "0705": DefaultRule(
         rule_id="0705",
         name="bom-encoding-in-file",
         msg="This file contains BOM (Byte Order Mark) encoding not supported by Robot Framework",
--- a/robocop/checkers/comments.py
+++ b/robocop/checkers/comments.py
@@ -9,7 +9,6 @@ from robot.utils import FileReader
 from robocop.checkers import RawFileChecker, VisitorChecker
 from robocop.rules import Rule, RuleParam, RuleSeverity
 from robocop.utils import ROBOT_VERSION
-from robocop.utils.misc import get_rule_help_url
 
 
 def regex(value):
@@ -49,7 +48,6 @@ rules = {
 
         """,
         added_in_version="1.0.0",
-        help_url=get_rule_help_url("default", "todo-in-comment"),
     ),
     "0702": Rule(
         RuleParam(
@@ -88,7 +86,6 @@ rules = {
 
         """,
         added_in_version="1.0.0",
-        help_url=get_rule_help_url("default", "missing-space-after-comment"),
     ),
     "0703": Rule(
         rule_id="0703",
@@ -109,7 +106,6 @@ rules = {
 
         """,
         added_in_version="1.0.0",
-        help_url=get_rule_help_url("default", "invalid-comment"),
     ),
     "0704": Rule(
         rule_id="0704",
@@ -139,7 +135,6 @@ rules = {
 
         """,
         added_in_version="1.3.0",
-        help_url=get_rule_help_url("default", "ignored-data"),
     ),
     "0705": Rule(
         rule_id="0705",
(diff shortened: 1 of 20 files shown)


robocop/checkers/comments.py
```diff
@@ -86,6 +88,7 @@ def regex(value):
 
         """,
         added_in_version="1.0.0",
+        help_url=get_rule_help_url("default", "missing-space-after-comment"),
```
bhirsz, line 91: Maybe there could be smarter way of doing it? Since name is already supplied, we only miss information whether rule is default or not. This could also be somehow automated.

I think what could be actually optionally configured would be whole url for the help. For example:

```
help_url='https://your_company.com/robocop/custom_rules/ruleABC
```

Or something similiar. We need to focus on actual problem at hand - all default rules resolve help url correctly but it's problematic for custom rules, and we need to focus on fixing it.

Also, this need to be optional as most users who create their own rules doesn't need their own custom docs url

Lakitna: Regarding the problem at hand, there are 2 issues at the moment:

- Default rules ✅
- Community rules ❌
- Custom rules ❌

This is why I didn't go for a more generic solution inside the RoboCop codebase. I also wanted to use the most explicit solution for this POC.

I briefly looked at auto-detecting the rule name, but that requires moving `get_rule_help_url` to the `Rule` class. In that case, what's happening will be more obscure. I also think it might cause weirdness when handling the custom rule syntax:

```python
Rule(
  help_url='https://your_company.com/robocop/custom_rules/ruleABC'
)
```

These weirdnesses might be caused by all rule types using the same class. In an ideal world, I think using `Rule`, `DefaultRule`, `CommunityRule`, and `CustomRule` would be great here. But that's a very backwards-incompatible change.

> Also, this need to be optional as most users who create their own rules doesn't need their own custom docs URL

It is in its current form. The help URL defaults to `""`

In any case, I agree that this solution is not perfect. Hence the POC instead of a full PR :)

bhirsz: Yeah I know, that's why I keep this as discussion and I'm open to ideas :) So the problem could be only:

1) either defining base url for rules
2) or defining 'format' that rules should use (ie. 'https://your_company.com/robocop/custom_rules/{rule_id}')

This could be solved by introducing some extra variable in the same files where rules are defined. Not the prettiest but the easiest:

```
RULE_DOCS_BASE_URL = "https://your_company.com/robocop/custom_rules/"
# or rule format string

rules = {

}
```

We may think outside the box here. Explicit is better than implicit, but I'm hesistant to forcing people to provide this url for every rule if the rule format stays the same. Another way would be creating class from Rule:

```
Rule:
    docs_base_url = "..."

YourCompanyRule(Rule):
    docs_base_url = "smth else"
```

The second option may be even better as it offers better flexibility and is more 'object friendly'

Lakitna: Also good to realize that the rule name part might have to differ for some situations. For example, in my situation my URL has to look something like:

```
https://company.com/linter-rules/custom.html#my-lint-rule-name-c01
```

Where `my-lint-rule-name` is the rule name and `c01` is the rule code.

At the moment, you're assuming that the only unique part of the URL is the rule name. A full URL in the `Rule` class removes those assumptions.

How about we flip your solution instead?

```
Rule:
    def __init__(self, ..., help_url):
        self.help_url = help_url

DefaultRule(Rule):
    docs_base_url = "https://robocop.readthedocs.io/en/stable/rules_list.html#{rule_name}"

    def __init__(self, name, ...):
        help_url = self.docs_base_url.format(rule_name=name)
        super().__init__(name, ..., help_url)

CommunityRule(Rule):
    docs_base_url = "https://robocop.readthedocs.io/en/stable/community_rules.html#{rule_name}"

    def __init__(self, name, ...):
        help_url = self.docs_base_url.format(rule_name=name)
        super().__init__(name, ..., help_url)
```

During `__init__` of `DefaultRule` and `CommunityRule` it can set the correct help URL based on the rule name. This might also simplify some of the default/community separation things elsewhere.

At the same time, this leaves the flexibility of giving a full URL for custom rules.

All we'd have to change for the rules in this repo is:

```diff
-    "0702": Rule(
+    "0702": DefaultRule(
```

bhirsz: Good idea! I'm in.

Thanks for the changes. Since I see you're involved, did you perhaps ha e access to our slack and saw recent dicsussion regarding merging tools? I'm currently refactoring a lot of core functionality while doing this (in a separate repo). So if there is need for any big change, or you see anything to change in tool fundamentals its good time for it

Lakitna: Awesome, let's go for this solution! I can probably start working on these changes somewhere next week

I hadn't seen the discussion. But I found the thread now. I'll keep an eye out for things that could be changed in the core :)

Lakitna:
Still needs a bit of testing, but otherwise done :)

Lakitna:
This PR is ready for review :)

bhirsz (commented):
I apologise for delay. I was conflicted with the topics because of the new ideas:

https://github.com/bhirsz/robotframework-cop-academy/issues/33

I'm also refactoring defining the rules in new robocop **heavily**: https://github.com/bhirsz/robotframework-cop-academy/tree/feature/refactor_rules 

That's why I was conflicted with introducing such breaking change, and then another. How do you feel about possibly postponing it, is it something you want sooner than later? Alternatively we can ignroe community rules for now (since they will be removed), and rename "DefaultRule" to "Rule" like in the beginning (sorry for the extra work). All the result may stay - in that case we will still have option to overwrite help url by creating own class but without any breaking change.

Lakitna:
> That's why I was conflicted with introducing such breaking change

This is a non-breaking change ;) It extends the existing `Rule` class with 1 additional optional attribute.

That being said, I'm fine with postponing this if you think you can find a better long-term solution :)

bhirsz:
> > That's why I was conflicted with introducing such breaking change
> 
> This is a non-breaking change ;) It extends the existing `Rule` class with 1 additional optional attribute.
> 
> That being said, I'm fine with postponing this if you think you can find a better long-term solution :)

You're right, I've taken a second look and I see it should be fine to merge. I will do it and make a new release. Thanks!

bhirsz (approved):

PR #1150, opened 2024-12-14
Feat/allow relative imports in custom rule folder

### LINKED ISSUE #1122: [Bug] Can't do relative imports in custom rules folder
### What happened?

I have a bunch of custom rules that only make sense for our organization/situation. But it can be a real pain to structure code written for custom rules.

As an example: I'm working on custom rules related to keyword names. For this, I wrote a keyword name parser to simplify the rule code. I want to use this like so:

```
robocop_rules/
- keyword_name_checker.py
- keyword_name_parser.py
```

```py
# keyword_name_checker.py

from .keyword_name_parser import KeywordNameParser
```

```py
# keyword_name_parser.py

class KeywordNameParser:
  pass
```

Running Robocop with this code will cause Python to raise the following error:

```
ImportError: attempted relative import with no known parent package
```

Changing the import line to the following also does not work:

```py
# keyword_name_checker.py

from robocop_rules.keyword_name_parser import KeywordNameParser
```
```
ModuleNotFoundError: No module named 'robocop_rules'
```

And this is also no dice

```py
# keyword_name_checker.py

from keyword_name_parser import KeywordNameParser
```
```
ModuleNotFoundError: No module named 'keyword_name_parser'
```

```
robocop .
```

See above

I expect the ability to import python files.

### Operating System

Windows

### Robocop version

5.4.0

Lakitna:
Closes #1122 

Still requires documentation.


bhirsz:
Looks good! Interesting enough, recently I got similar question from someone regarding relative importing in external rules. Workaround was to add 'sys.path.append' inside rule module, before import. This PR should fix it.

Lakitna:
Awesome. If you're convinced, I'll add some docs somewhere this week.

Lakitna:
I added a section on using an external rules folder. It includes the new import capabilities. Please let me know if it's clear like this.

bhirsz (approved):

PR #1176, opened 2025-02-18
Rules/too many required arguments

### LINKED ISSUE #1145: [Rule] Too many arguments without default
I am personally not a fan of the rule `too-many-arguments`. Instead, I want to propose `too-many-arguments-without-default`.

A lot of arguments can make a keyword cumbersome to use. It can also be an indication of too much keyword complexity. However, I don't think having a lot of arguments with defaults is always an issue. For example:

```robotframework
*** Keywords ***
Create user account
    [Arguments]
    ...  ${environment}=tst
    ...  ${type}=online
    ...  ${validateAge}=${True}
    ...  ${validateEmail}=${True}
    ...  ${acceptTerms}=${True}
    ...  ${retentionDays}=${14}
```

This is (a modified version of) a keyword we use all the time for user account creation. Note it has more than the default allowed 5 arguments, but all have a default value. The unmodified version has 10 arguments, all with a sensible default value.

In most tests, we use the keyword without arguments. In tests where we need a specific account state, we use the keyword with a few arguments overwritten. I think that with a pattern like this, you can create great keywords with many arguments. These keywords offer great flexibility.

At the same time, I think that the following is a bad pattern and should not be used:

```robotframework
*** Keywords ***
Create user account
    [Arguments]
    ...  ${environment}
    ...  ${type}
    ...  ${validateAge}
    ...  ${validateEmail}
    ...  ${acceptTerms}
    ...  ${retentionDays}=${14}
```

There are too many arguments here that you always have to explicitly fill. This makes the keyword hard to use and hard to understand.

There are a few ways we can achieve this:

1. Add the new rule `too-many-arguments-without-default` as a non-default rule.
2. Add a new parameter to the existing rule `too-many-arguments`. This parameter could be called `ignore_optional_arguments`.
3. Make this the new behavior of the existing rule `too-many-arguments`.

If you agree with me, please let me know and I'll open a PR.

Lakitna:
Closes #1145


Lakitna:
The failing test is caused by `too-many-arguments` being disabled. I'll hold off on looking for a fix until @bhirsz can chime in

robocop/checkers/lengths.py
```diff
@@ -502,6 +561,60 @@ def visit_Keyword(self, node):  # noqa: N802
...
+                col=arg_node.col_offset + 1,
+                end_col=arg_node.col_offset + len(arg_node.value) + 1,
+                extended_disablers=(keyword_node.lineno, keyword_node.end_lineno),
+                sev_threshold_value=arg_count,
+            )
+
+        required_arg_count = 0
+        optional_arg_count = 0
+        for arg in arguments.values:
+            if "}=" in arg:
```
bhirsz, line 585: We may want to ignore errors as well, for example:

```
[Arguments]    ${argument} = default
```

Not sure how it behaves but most likely it's parsing error, but may still come up here as 'optional' argument

robocop/checkers/lengths.py
```diff
@@ -502,6 +561,60 @@ def visit_Keyword(self, node):  # noqa: N802
                 sev_threshold_value=key_calls,
             )
 
+    def _visit_keyword_arguments(self, keyword_node: Keyword, arguments: Arguments):
+        arg_node = arguments.get_token(Token.ARGUMENTS)
+
+        arg_count = len(arguments.values)
+        if arg_count > self.param("too-many-arguments", "max_args"):
```
bhirsz, line 568:
```suggestion
        if (arg_count := len(arguments.values)) > self.param("too-many-arguments", "max_args"):
```

bhirsz:
LGTM! Only minor comments.

As for failing tests, I see it's test_all_builtin_rules_should_be_enabled_by_default - it can be ignroed (commented out even) as it will be deprecated in new robocop anyway.

bhirsz:
My apologies - it looks like Github closed the PR after I started clean ups in the repository. I have changed master branch to main.

Also, I have decided to keep merged Robocop inside this repository. I didn't want to lose all the stars, contributions and commit history after all! That means I will be copying the PRs from 'academy' repo into main branch here today. For your PR you have two options:

1) Raise it againsts 'deprecated_robocop' branch (which uses current, soon to be deprecated code) and I will migrate it before releasing merged Robocop
2) Raise it against main, but I will either update it on your branch or you can adjust the code to the new format of rules. It's not much work, so I don't mind either

Issue #1396, opened 2025-05-06
Suggest using type-slim instead of typer to avoid shellingham dependency


theodore86:
Hi,

Thanks for the great work on this project!

I noticed that the project uses [typer](https://github.com/tiangolo/typer), which brings in shellingham as a transitive dependency. In some environments (e.g., minimal containers or Alpine-based systems), shellingham can introduce platform-specific issues or unnecessary weight.

I’d like to suggest using [type-slim](https://github.com/tiangolo/typer/tree/master/typer#type-slim) as a drop-in replacement for typer. It offers the same API but excludes dependencies like shellingham, making it a more lightweight alternative for CLI tools where shell detection isn't critical.

If you're open to it, I’d be happy to submit a PR with this change.

Let me know what you think!.

bhirsz:
Yes, we have switched recently to typer (robotidy from click, and robocop from argparse). I'm not sure if we're using ``shellingham`` directly, but we can try to replace typer with typer-slim and see how tests will behave. I'm open to PR!

theodore86:
Great! i will try to update the pyproject.toml

PR #1398, opened 2025-05-07
Change typer to typer-slim in pyproject.toml

theodore86:


theodore86:
#1396

bhirsz:
It seems the tests are failing due to missing rich package. Should we include both typer-slim and rich or it would defeat the purpose for your? (slimming down robocop package). Optionally it may be possible to wrap rich so it's only optionally used, but it may require more work.

theodore86:
definitely we need to include rich, i will give one more try

theodore86:
I have update the PR with the rich package as direct dependency of the project

Issue #1429, opened 2025-08-06
[Rule] Add `ignore_templated` option to LEN04 (too-long-test-case)


GeoGegl:
**Description**
I would like to configure rule 
* LEN04: too-long-test-case 

to skip test cases using `[Template]` syntax
The same as following rules (These already support `ignore_templated`)

* LEN05: too-few-calls-in-test-case
* LEN06: too-many-calls-in-test-case

PR #1430, opened 2025-08-06
[Rule] Add ignore_templated option to LEN04 (too-long-test-case)

GeoGegl:
I also created an issue here: https://github.com/MarketSquare/robotframework-robocop/issues/1429

I would like the ignore_templated parameter for rule LEN04 in the same way rule LEN06 already has it


bhirsz:
Good work, thanks for contribution! I will try to release it as 6.6 today ~ tommorrow.

PR #1445, opened 2025-08-24
invalid-for-loop: Change test data to reflect a change in robot error message in RF7.3

p-zander:
I just noticed that this test is failing with latest RF, very slight change in the error message. I wasn't sure about the extended output, if it should continue to be checked for `>=5.0;<7.3`.

robotframework/robotframework@448fb07


bhirsz:
Ideally it should be checked for older versions, but since it was already validaded for those I don't think there will be much change in new versions which could be breaking for older ones. So it's fine with current versions, especially we are covering all versions with other type of output.

Issue #1443, opened 2025-08-20
[Rule] Propose too-long-variable-name


p-zander:
Hi!

I would like a rule for keeping the length of variable names in check. I would propose to take over parameters and default values from _too-long-keyword_:

Name | Default value | Type | Description
-- | -- | -- | --
severity | W | severity | Rule severity (E = Error, W = Warning, I = Info)
max_len | 40 | int | number of characters allowed in name

Would make sense to also enable use of Severity thresholds with this.

bhirsz:
Variable name resolution is not in the scope.  So variable in variable name will always be considered as it is, without checking dynamic name as it would be too complex. And it wouldnt make sense as its runtime dependant. 

My thoughts only related to variables itself, where they should be ignored totally of found in another variable name or simply counted as part of string.  It will be possibly the later. 

Given that, variable creation in suite settings, assignment and special syntax such as VAR will be considered. Also variable usages (arguments, for loop items etc). So basically everywhere we use variable. We may offer flag to only flag on variable creation, not the use.

bhirsz:
The code looks quite good! I like the use of walrus operators, it give me some ideas for improvements in existing code. :) Yeah, ``generic_visit`` is one of pitfalls. I may add it to contribution guidelines (which are also during development since 6.0 merge..). I usually don't put it unless I need to visit 'deeper' to save time, but then you need to remember to enable visiting when updating the behaviour of checker.

I agree that we shouldn't resolve dynamic variable names

I am thinking about one part:

```
    def visit_Arguments(self, node):  # noqa: N802
        for argument in node.get_tokens(Token.ARGUMENT):
            self.check_variable_name(argument)
```

Do you think it should be handled by this rule or another new rule ``too-long-argument-name``? One one hand I don't want to split rules too much, on other hand there may be people who want to track it separate. So I don't have decision on it yet.

I would also add some error testing - for example invalid variables:

```
*** Variables ***
JUST_THIS
${NOT_CLOSED    1
  ${with_spaces_before}
```

Also correct examples (variable names that shouldn't trigger error). I believe your test will also fail on older version of Robot Framework (because VAR will be considered as keyword call and its arguments not parsed the same) but we can handle it after opening PR by adding another expected output file for older versions.

p-zander:
Thanks for the hints! Yes, right, the test of course breaks with RF4.0, I'll fix that and add a test for malformed variable creations/assignments.

[...]

PR #1450, opened 2025-08-29
[Rule] Implement too-long-variable-name

p-zander:
Implements #1443

New rule checks all variable names in contexts such as

- `*** Variables ***` section (suite variables)
- `Set Variable` and `VAR` syntax (RF7+)
- `Set Local/Test/Task/Suite/Global Variable`
- Iteration variables used in `FOR`-loops
- Keyword `[Arguments]`-setting
- Keyword arguments [embedded into the keyword name](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#toc-entry-365) (BDD)
- Variables used for catching error messages in `TRY-EXCEPT AS` branches

A variable name that is longer than `too-long-variable-name.max_len (default: 40)` is reported **once per scope**, so an offending name is not overreported if it's assigned to several times.

Supports severity thresholds as well.

Additionally some reusable function were moved to utils/misc.py.

Need to complete test data for all edge cases; Might add some options to the rule if it makes sense.


src/robocop/linter/rules/lengths.py
```diff
@@ -803,6 +844,104 @@ def count_keyword_calls(self, node):
...
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.variables: dict[str, tuple] = {}
+
+    def visit_Node(self, node) -> None:  # noqa: N802
+        if not node.errors:
+            self.generic_visit(node)
+
+    def visit_Scope(self, node: Keyword | TestCase | VariableSection):  # noqa: N802
```
bhirsz, line 868: Is visit_Scope actual visitor from Robot we're overwriting, or simply our function name? If so, we could just simply name it after one of the visitors in line 876 (like visit_Keyword)

src/robocop/linter/rules/lengths.py
```diff
@@ -609,6 +622,27 @@ class EmptyKeywordTagsRule(Rule):
...
+
+    name = "too-long-variable-name"
+    rule_id = "LEN32"
+    message = "Variable name '{variable_name}' is too long ({variable_name_length}/{allowed_length})"
+    severity = RuleSeverity.WARNING
+    parameters = [
+        RuleParam(name="max_len", default=40, converter=int, desc="allowed length of a variable name"),
+    ]
+    severity_threshold = SeverityThreshold("max_len", compare_method="greater", substitute_value="allowed_length")
+    added_in_version = "6.6.x"
```
p-zander, line 640:
```suggestion
    added_in_version = "6.6.2"
```

p-zander:
Still open points:

- One instance of underreporting I don't quite understand, but might be acceptable: [here](https://github.com/MarketSquare/robotframework-robocop/pull/1450/files#diff-fd03ea4e5bec9bc9cd8c3063109bc75723516d12c14df2fd446c38a5c07b9f21R35)
- Adding test data for type hints/conversion
- Adding invalid test data

Other than that I'm pretty happy with the results, although the implementation might look a bit wonky.

src/robocop/linter/rules/lengths.py
```diff
@@ -609,6 +622,27 @@ class EmptyKeywordTagsRule(Rule):
...
+    sonar_qube_attrs = sonar_qube.SonarQubeAttributes(
+        clean_code=sonar_qube.CleanCodeAttribute.FOCUSED, issue_type=sonar_qube.SonarQubeIssueType.CODE_SMELL
```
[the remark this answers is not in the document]
bhirsz: I am also not sure! If you copied those from length rules, it should be fine. I tried to apply those attributes with accordance to what I have found on Sonar Qube documentation, but most of it reuses the same attributes in case of our rules anyway.

p-zander:
Mh, btw how sure are you that in your workflows the tests are actually executed against all these different RF versions? At least one package is updated [here](https://github.com/MarketSquare/robotframework-robocop/actions/runs/17516421821/job/49754471309?pr=1450#step:6:14), and if I had to guess that would be `robotframework` itself.

If I try to do it locally (downgrade RF and then run `uv run` with `--verbose` flag) I see this 
image

bhirsz:
I am quite sure if it comes for robocop pre merge. After merge..not so, since we also switch to uv and it could have its own quirks. Thanks for pointing it out, I will also take a look what's happening here.

bhirsz:
> Mh, btw how sure are you that in your workflows the tests are actually executed against all these different RF versions? At least one package is updated [here](https://github.com/MarketSquare/robotframework-robocop/actions/runs/17516421821/job/49754471309?pr=1450#step:6:14), and if I had to guess that would be `robotframework` itself.
> 
> If I try to do it locally (downgrade RF and then run `uv run` with `--verbose` flag) I see this 
> image
> 

I am quite sure if it comes for robocop pre merge. After merge..not so, since we also switch to uv and it could have its own quirks. Thanks for pointing it out, I will also take a look what's happening here.

p-zander:
Yeah, so I tested locally without `uv run` and tests were failing/should have been failing before for at least some earlier RF versions. I'm also not really familar with uv, but the `--with` option seems to be what you want: https://docs.astral.sh/uv/concepts/projects/run/#requesting-additional-dependencies

p-zander:
I think I might be done with this for now. Please let me know, if I should add some options to the rule or anything.

PR #1452, opened 2025-09-07
Actually run pytest tests with configured RF version

p-zander:
Using ~`uv run --with=...`~ `uv add` to explicitly install correct RF version for each test run. Also add sanity check before, that confirms the correct version (see [here](https://github.com/MarketSquare/robotframework-robocop/actions/runs/17534026091/job/49794716701?pr=1452#step:6:14)).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,7 +52,7 @@ jobs:
         run: uv sync --all-extras --dev
 
       - name: Run tests
-        run: uv run --with=robotframework==${{ matrix.rf-version }} coverage run --source src/robocop -m pytest
+        run: uv run --with=robotframework==${{ matrix.rf-version }} coverage run --source src/robocop -m pytest -v
 
       - name: Calculate coverage
         run: uv run coverage report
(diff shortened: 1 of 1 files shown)


p-zander:
Sorry for playing around, but I think something is still not right. Tests that should run on certain versions are still skipped, [see here](https://github.com/MarketSquare/robotframework-robocop/actions/runs/17533657374/job/49793881005?pr=1452#step:6:866) for example. I'm not sure, why this
`uv run --with=robotframework==4.1.3 robot --version`
or this
`uv run --with=robotframework==4.1.3 python -c 'from robot.version import VERSION; print(VERSION)'`
both work, but tests are still skipped based on wrong version.

.github/workflows/tests.yml
```diff
@@ -51,10 +51,12 @@ jobs:
       - name: Install the project
         run: |
           uv sync --all-extras --dev
-          uv pip install robotframework==${{ matrix.rf-version }}
+          uv add robotframework==${{ matrix.rf-version }}
 
       - name: Run tests
-        run: uv run coverage run --source src/robocop -m pytest
+        run: |
+          uv run robot --version || true
```
p-zander, line 58: See https://github.com/robotframework/robotframework/issues/5184

bhirsz:
Huge thanks with helping out with this! I already see some bugs that one of the user reported but I couldn't reproduce on newer versions.... It lookslike it now executes as it should

PR #1547, opened 2025-11-30
typo

dahagag:


bhirsz:
Thanks! I need to check how to verify automatically if all external urls are correct.

PR #1564, opened 2025-12-18
feat: add commented-out-code detection rule (COM06)

mikeleppane:
## Summary
- Add new rule `commented-out-code` (COM06) that detects Robot Framework code patterns in comments
- Uses RF's native tokenizer to accurately distinguish code from prose (e.g., "IF you need help" is not flagged)
- Disabled by default to avoid noise in existing projects

## Details

The rule detects:
- Variable assignments (`# ${var}=    Get Value`)
- Control structures (`# IF`, `# FOR`, `# WHILE`, `# TRY`, etc.)
- Setting brackets (`# [Tags]`, `# [Arguments]`, `# [Setup]`, etc.)
- Settings section statements (`# Library`, `# Resource`, `# Variables`, etc.)
- RF 7.0+ syntax (`# VAR`, `# RETURN`, `# GROUP`)

The rule ignores:
- Comments starting with TODO/FIXME markers (configurable via `markers` parameter)
- Comments inside `[Documentation]` sections (code examples are common there)
- `*** Comments ***` section (meant for free-form text)
- Plain prose that happens to contain RF keywords

## Test plan
- [x] Unit tests for all detection patterns
- [x] Test for custom markers configuration
- [x] Extended output format tests
- [x] Verify no false positives on prose comments
- [x] Verify `*** Comments ***` section is skipped
- [x] ruff format and ruff check pass

### DIFF
--- a/src/robocop/linter/rules/comments.py
+++ b/src/robocop/linter/rules/comments.py
@@ -568,7 +568,9 @@ class CommentChecker(VisitorChecker):
             if self.setting_pattern.match(value):
                 return True
 
-        return False
+        # Check KEYWORD tokens - in older RF versions (4.x, 5.x), standalone control keywords
+        # like ELSE, END, etc. are tokenized as KEYWORD instead of ERROR
+        return bool(first_token.type == "KEYWORD" and first_token.value in self.control_keywords)
 
     def is_block_comment(self, comment: str) -> bool:
         return comment == "#" or self.missing_space_after_comment.block.match(comment) is not None
--- a/tests/linter/rules/comments/commented_out_code/custom_markers/expected_output.txt
+++ b/tests/linter/rules/comments/commented_out_code/custom_markers/expected_output.txt
@@ -1,1 +1,1 @@
-custom_markers/test.robot:5:5 [W] COM06 Commented out code: '${result}=    this should be reported'
+custom_markers${/}test.robot:5:5 [W] COM06 Commented out code: '${result}=    this should be reported'


src/robocop/linter/rules/comments.py
```diff
@@ -268,6 +467,111 @@ def check_comment_content(self, token: Token, content: str) -> None:
...
+        # Skip if comment starts with ignore markers (TODO, FIXME, etc.)
+        low_body = comment_body.lower()
+        for marker in self.commented_out_code.markers:
+            if low_body.startswith(marker):
+                return
+
+        # Check for RF code patterns using the tokenizer
+        is_rf_code = self._is_rf_code(comment_body)
+
+        if is_rf_code:
```
bhirsz: Potentially unneeded variable ``is_rf_code`` as we can evaluate directly:

```
if self._is_rf_code(comment_body):
```

bhirsz:
Great work! As for failing tests, it looks like new syntax like GROUP, EXCEPT, VAR etc failed for older versions. I usually handle it by splitting expected files and having separate test cases. Sometimes I reuse the same input files, sometimes I create new without new syntax at all:

```
    def test_formatter(self):
        self.compare(source="test.robot", test_on_version=">=5")

    def test_formatter_rf4(self):
        self.compare(source="test.robot", expected="test_rf4.robot", test_on_version="<=4")
```

```
    def test_rf5_syntax(self):
        self.compare(source="test_5.robot", test_on_version=">=5")
```

PR #1565, opened 2025-12-18
feat: add file-level caching for linter and formatter to skip unchanged files

mikeleppane:
This PR introduces a high-performance file-level caching system for Robocop's linter and formatter, dramatically improving subsequent run times by skipping analysis of unchanged files.

Large Robot Framework projects with hundreds of test files experience slow linting times, even when most files haven't changed. Users running Robocop in CI/CD pipelines or pre-commit hooks repeatedly encounter this. Before: Every run processes all files from scratch. After: Only changed files are reprocessed, and cached results are restored instantly

###  Cache Architecture
```mermaid
flowchart TD
    A[Start: robocop check] --> B{Cache enabled?}
    B -->|No| C[Process all files normally]
    B -->|Yes| D[Load cache from .robocop_cache/cache.msgpack]
    D --> E{Cache exists?}
    E -->|No| F[Create empty cache]
    E -->|Yes| G[Deserialize cache data]
    G --> H{Version matches?}
    H -->|No| F
    H -->|Yes| I[Cache loaded successfully]
    F --> J[For each file...]
    I --> J
    J --> K{Check cache entry}
    K -->|Not in cache| L[Process file]
    K -->|In cache| M{Validate entry}
    M -->|File changed mtime/size| L
    M -->|Config changed hash| L
    M -->|Valid| N[Restore cached diagnostics]
    L --> O[Parse & analyze file]
    O --> P[Store results in cache]
    N --> Q[Continue to next file]
    P --> Q
    Q --> R{More files?}
    R -->|Yes| J
    R -->|No| S[Save cache to disk]
    S --> T[Report diagnostics]
    C --> T
```
### Cache invalidation
```mermaid
flowchart LR
    A[Cache Entry] --> B{Config hash matches?}
    B -->|No| Z[❌ Invalid - Rescan]
    B -->|Yes| C{File mtime matches?}
    C -->|No| Z
    C -->|Yes| D{File size matches?}
    D -->|No| Z
    D -->|Yes| E[✅ Valid - Use cache]
    
    style E fill:#90EE90
    style Z fill:#FFB6C6
```
The cache automatically invalidates when:

- File modification time (mtime) changes
- File size changes
- Robocop version changes
- Linter/formatter configuration changes
- Language settings change
- Selected rules change

### Benchmark Results

Benchmarks performed on realistic Robot Framework test files (each containing 3 test cases, multiple keywords, variables, and documentation). All measurements averaged over multiple runs.

| Files | Without Cache | With Cache (warm) | Speedup | Time Saved | Cache Size |
|------:|-------------:|-----------------:|--------:|-----------:|-----------:|
| **100** | 450.5ms | 31.2ms | **14.4x** | 93.1% | 106 KB |
| **1,000** | 4,514ms (4.5s) | 153.5ms | **29.4x** | 96.6% | 1.04 MB |
| **10,000** | 53,097ms (53s) | 2,057ms (2s) | **25.8x** | 96.1% | 10.37 MB |

### DIFF
--- a/src/robocop/config.py
+++ b/src/robocop/config.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import hashlib
 import os
 from dataclasses import dataclass, field, fields
 from enum import Enum
@@ -383,6 +384,9 @@ class LinterConfig:
         Used for cache invalidation - if configuration changes, cached results are invalidated.
         Note: This makes LinterConfig usable as dict key, but only hash config-affecting fields.
 
+        Uses stable hashing (SHA256) to ensure consistent hashes across Python process restarts,
+        avoiding issues with Python's hash randomization (PEP 456).
+
         Returns:
             Hash value of the configuration options.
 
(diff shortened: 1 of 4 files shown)


src/robocop/linter/runner.py
```diff
@@ -189,7 +232,9 @@ def make_reports(self) -> None:
         for report in self.reports.values():
             prev_result = prev_results.get(report.name) if prev_results is not None else None
             report.generate_report(
-                diagnostics=diagnostics, config_manager=self.config_manager, prev_results=prev_result
+                diagnostics=diagnostics,
```
bhirsz: I saw several lines split, did you perhaps run it with formatter with configured length 88 on it? We use 120 by default so it shouldn't split, unless I'm not seeing something.

src/robocop/config.py
```diff
@@ -680,12 +800,18 @@ def overwrite_from_config(self, overwrite_config: Config | None) -> None:
...
+                "cache",
                 "config_source",
             ):  # TODO Use field metadata
                 continue
             value = getattr(overwrite_config, config_field.name)
             if value:
                 setattr(self, config_field.name, value)
+        # Handle cache config - CLI cache settings override file config
+        if overwrite_config.cache is not None and (
+            overwrite_config.cache.enabled is False or overwrite_config.cache.cache_dir is not None
```
bhirsz: Is there a reason why we would like to overwrite config only when enabled is False or cache_dir is set? I know that the first though would be 'no point overwriting if it's the same as default' but what about this case:

Configuration file
```
[tool.robocop]
cache = false
```

CLI
```
robocop check --cache
```

The configuration file explictly says 'disable the cache' but we don't want to enable it in this run, so we call --no-cache.

bhirsz:
I saw interesting feature in many tools (pytest, ruff). If we're creating the cache dir, we can also create .gitignore file inside with the following content:

```
*
```

It will auto-ignore our cache from git. Some are also adding ``README.md`` with the explanation what's the purpsoe of the directory (though it's not needed for us). I would suggest to also add .gitignore generation

Issue #1439, opened 2025-08-10
[Rule] Keyword argument should have type


bhirsz:
It should be optional.

```
*** Keywords ***
Move
    [Arguments]    ${distance: float}    ${missing_type}    ${log: bool}=False
    IF    ${log}
        Log    Moving ${distance} meters.
    END
```

#1399

bhirsz:
@mikeleppane Calling you here for additional question. We have now rules for variable and argument typing, stored in variables rule group. Do you think it is good idea to create a new group for those rules? For example robocop.linter.rules.typing.py which would containg TYPE01, TYPE02.. rules. It's bit murky area, since those rules are variable / argument rules, but they are specific enough it may be worth creating new group for them (and make it easier to enable / disable them in one go).

mikeleppane:
> [@mikeleppane](https://github.com/mikeleppane) Calling you here for additional question. We have now rules for variable and argument typing, stored in variables rule group. Do you think it is good idea to create a new group for those rules? For example robocop.linter.rules.typing.py which would containg TYPE01, TYPE02.. rules. It's bit murky area, since those rules are variable / argument rules, but they are specific enough it may be worth creating new group for them (and make it easier to enable / disable them in one go).

Perhaps, it would be a good idea to separate them into their own group.

PR #1587, opened 2025-12-23
refactor: move variable and argument typing rules to ANN (typing) group

bhirsz:
As discussed in #1439

@mikeleppane Asking you as the author, do you think it's better to keep those rules in  variables or split it into new group for typing only? For now I have called the group "ANN" (for annotations) and it's stored in rules.typing.py .


mikeleppane:
> As discussed in #1439
> 
> @mikeleppane Asking you as the author, do you think it's better to keep those rules in variables or split it into new group for typing only? For now I have called the group "ANN" (for annotations) and it's stored in rules.typing.py .

This approach sounds very reasonable!

PR #1589, opened 2025-12-23
fix(mcp): fix limit handling bugs and add enhancements

mikeleppane:
Bug fixes:
- Fix lint_and_format issues_fixed calculation when limit is applied. Previously issues_after was truncated by limit while issues_before was not, giving incorrect counts.
- Fix lint_directory summary counts when limited. Now processes all files for accurate statistics and only truncates the issues list.
- Fix server.py docstring typo (stray apostrophe)

Enhancements:
- Add configure parameter to linting tools (lint_content, lint_file, lint_directory, lint_and_format) allowing rule parameter configuration e.g., ["line-too-long.line_length=140"]
- Add formatter parameters and skip_options to get_formatter_info response, matching the parameter info available in get_rule_info

### DIFF
--- a/tests/mcp/test_tools.py
+++ b/tests/mcp/test_tools.py
@@ -443,6 +443,7 @@ class TestLintAndFormatIntegration:
         # Formatting should have fixed some issues or at least not made it worse
         assert isinstance(issues_before, list)
         assert isinstance(issues_after, list)
+        assert len(issues_after) <= len(issues_before)
 
     def test_lint_and_format_with_limit_accurate_counts(self):
         """Test that lint_and_format returns accurate counts even with limit."""


tests/mcp/test_tools.py
```diff
@@ -234,3 +252,216 @@ def test_lint_file_with_file_in_result(self, tmp_path: Path):
...
+        # Format the content
+        format_result = _format_content_impl(content)
+        assert format_result["changed"] is True
+
+        # Get issues after formatting
+        issues_after = _lint_content_impl(format_result["formatted"])
+
+        # Formatting should have fixed some issues or at least not made it worse
+        assert isinstance(issues_before, list)
+        assert isinstance(issues_after, list)
```
bhirsz: The comment suggest we want to check if number of issues did no go up. But it looks we only check for type.

Maybe it's missing?:

```
assert len(issues_after) <= len(issues_before)
```

PR #1603, opened 2025-12-27
refactor(mcp): use Pydantic models for structured tool outputs and add AI-powered auto-fix generation 

mikeleppane:
### Summary

Replace dict-based return values with Pydantic models across all MCP tools to enable structured output with automatic JSON schema generation and type validation.

Structured output with Pydantic models enables the following benefits:

- **JSON Schema generation** - Pydantic automatically generates JSON schemas from the models. MCP uses these schemas to tell the LLM what fields each tool returns, their types, and descriptions. The LLM can then understand and use the responses better.
- **Validation** - If a tool accidentally returns wrong data (e.g., string instead of int), Pydantic catches it immediately rather than silently passing bad data.
- **Self-documenting** - Field descriptions like Field(description="Count of Error-level issues") serve as documentation and are included in the schema.
- **Developer experience** - IDE autocompletion, type hints, and catch bugs at development time.

Changes:
- Add models.py with Pydantic models for all tool return types (DiagnosticResult, FormatContentResult, LintAndFormatResult, etc.)
- Update all tool implementations to return Pydantic models instead of dicts
- Fix SeveritySummary field name from 'I' to 'INFO' with alias for backward compatibility (resolves ruff E741 ambiguous variable name)

Bug fixes discovered during integration testing:
- Fix dict-style access (result["key"]) to attribute access (result.key) in registration.py context messages - Pydantic models use attributes
- Remove 'from __future__ import annotations' from registration.py as it breaks Pydantic's runtime type evaluation for Annotated fields

Testing:
- Update existing unit tests for new Pydantic model structure

### DIFF
--- a/src/robocop/mcp/tools/documentation.py
+++ b/src/robocop/mcp/tools/documentation.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import inspect
+import operator
 
 from fastmcp.exceptions import ToolError
 
@@ -16,7 +17,11 @@ from robocop.mcp.tools.models import (
     RuleSearchResult,
     RuleSummary,
 )
-from robocop.mcp.tools.utils.helpers import _create_match_snippet, _iter_unique_rules, _rule_to_dict
+from robocop.mcp.tools.utils.helpers import (
+    _create_match_snippet,
+    _iter_unique_rules,
+    _rule_to_dict,
+)
 
 
 def _list_rules_impl(
@@ -46,7 +51,7 @@ def _list_rules_impl(
         )
         for rule in _iter_unique_rules(category, severity, enabled_only)
     ]
-    return sorted(rules, key=lambda r: r.rule_id)
+    return sorted(rules, key=operator.attrgetter("rule_id"))
 
 
 def _list_formatters_impl(enabled_only: bool = True) -> list[FormatterSummary]:
--- a/src/robocop/mcp/tools/formatting.py
+++ b/src/robocop/mcp/tools/formatting.py
@@ -74,8 +74,8 @@ def _format_content_impl(
 
             diff_text = None
             if changed:
-                old_lines = [line + "\n" for line in old_model.text.splitlines()]
-                new_lines = [line + "\n" for line in new_model.text.splitlines()]
+                old_lines = old_model.text.splitlines(keepends=True)
+                new_lines = new_model.text.splitlines(keepends=True)
                 diff_text = "".join(unified_diff(old_lines, new_lines, fromfile="before", tofile="after"))
 
             return FormatContentResult(formatted=new_model.text, changed=changed, diff=diff_text)


mikeleppane:
Yes, you are right to question this. In our case dataclasses vs pydantic give nearly identical benefits. The primary benefit you’re getting from Pydantic is construction-time validation, which is a reasonable safety net but not strictly necessary. But because fastmcp already has a dependency for pydantic we get it a "free". 

When you zoom out and think about where MCP is heading, Pydantic might become a better choice. 

For a standalone tool returning simple results to a human via an LLM, dataclasses are sufficient.
For tools that will participate in agentic pipelines, be composed with other tools, evolve over time, and operate in a growing ecosystem, Pydantic is the investment that pays off.

Given where MCP is headed — toward more sophisticated agents, more complex tool chains, and tighter integration between AI systems — Pydantic is the forward-looking choice. You’re building infrastructure, not just tools, and infrastructure benefits from stronger contracts.

src/robocop/mcp/tools/documentation.py
```diff
@@ -25,39 +33,39 @@ def _list_rules_impl(
...
+            rule_id=rule.rule_id,
+            name=rule.name,
+            severity=rule.severity.value,
+            enabled=rule.enabled,
+            message=rule.message,
+        )
         for rule in _iter_unique_rules(category, severity, enabled_only)
     ]
-    return sorted(rules, key=operator.itemgetter("rule_id"))
+    return sorted(rules, key=lambda r: r.rule_id)
```
bhirsz: Why not operator attrgetter since we used operator previously?

src/robocop/mcp/tools/formatting.py
bhirsz: In the current scope, while we still redesign it its fine though ultimately we will need to reuse duplicated code like this one. I am also in fault here since I have several duplicates or functions with similar purpose in the code ;)

mikeleppane:
@bhirsz, I have some additional ideas for enhancing the MCP functionality. What do you think about these:

1. **Semantic Code Analysis**

Because Robocop rules are syntactic, we could improve this by adding a level of semantic analysis which could detect duplicate keywords (find semantically similar keywords/tests (not just identical), dead code, and missing abstractions. Yes, we have some of these already in place, but now this could be done across files.

Example:

```bash
User: "Analyze my test suite for duplicate keywords and dead code"
```

2. **Natural Language Rule Configuration**

We could have a prompt that will generate a correct Robocop configuration based on the user's request.

Example:

```bash
User: "Configure Robocop for a new project. We want strict quality, but:
       - Line length up to 140 is fine
       - We don't care about keyword naming conventions  
       - Missing test documentation should be an error
```

3. Intelligent Code Review 

Combine Robocop's rules with team-specific guidelines.
We could have a prompt that does this, and team guidelines reside in a specific place where the MCP can load them.

Issue #1684, opened 2026-02-11
Replace depracated dependency typer-slim


Houbein:
Robocop has as a dependency typer-slim, this had just been declared depracated. 
From their project description (https://pypi.org/project/typer-slim): 
"This package, typer-slim, does nothing other than depend on typer. 

There used to be a slimmed-down version of Typer called typer-slim, which didn't include the dependencies rich and shellingham, nor the typer command. 

However, since version 0.22.0, we have stopped supporting this, and typer-slim now simply installs (all of) Typer. 

If you want to disable Rich globally, you can set an environmental variable TYPER_USE_RICH to False or 0. 

The only reason this package exists is as a migration path for old projects that used to depend on typer-slim, so that they can get the latest version of typer. 

You should not install this package."

PR #1685, opened 2026-02-12
feat: Replace `typer-slim` with `typer`

svlandeg:
Hi! Just a quick heads-up that we've removed support for `typer-slim`. From v0.22.0 onwards, `typer-slim` will be the same as `typer`, with external requirements `shellingham` and `rich` installed by default, see https://github.com/fastapi/typer/pull/1522. I saw that `rich` is already an external dependency of this library so this should have little influence.

Let me know if you have any questions/concerns!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
         args: ["--config-file=pyproject.toml", "src/robocop"]
         pass_filenames: false
         additional_dependencies:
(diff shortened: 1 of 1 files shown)


bhirsz:
Thanks for PR, we actually also got report about this: closes #1684

We used slim since we didnt use rich, but then we merged two tools out of which one used rich. So it should be fine to use typer directly.

PR #1786, opened 2026-07-30
fix: mark --ignore values as matched for rules below --threshold

### LINKED ISSUE #1775: Warning when utilizing severity thresholds: Option value 'xyz' from '--ignore' did not match with any rule name or id.
If one utilizes severity thresholds, ex. `robocop check -t E` and there is an overlap of excluded severities and the ignore parameters, a warning message is produced ex.:
```
> robocop check --threshold E --ignore sleep-keyword-used
Option value 'sleep-keyword-used' from '--ignore' did not match with any rule name or id.
```
I was just thrown off by this behavior for some time, thinking nothing changed about my rule "selection". The documentation reads
```
Use --threshold / -t option to filter out rules
```
so I was thinking of `--select`/`--ignore` along the lines of "before the fact" and `--threshold` as "after the fact", so thought it wouldn't affect the rules "selection". Maybe it would make sense to omit this new warning message, if rules are ignored by the threshold or clarify in the documentation.

Sanjays2402:
Closes #1775

`robocop check --threshold E --ignore sleep-keyword-used` printed `Option value 'sleep-keyword-used' from '--ignore' did not match with any rule name or id.` even though the rule exists. `RuleMatcher._is_rule_disabled` returned early on the severity-threshold check, so `ignore_filter.matches()` never ran for those rules and the ignore value was never recorded as matched. The ignore filter is now evaluated before the threshold short-circuit; rule enable/disable behaviour is unchanged.

Regression test added to `tests/linter/test_rule_matcher.py` (fails without the fix, passes with it); `tests/linter` is green, ruff check/format on the changed files is baseline-identical.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/src/robocop/runtime/resolver.py
+++ b/src/robocop/runtime/resolver.py
@@ -129,9 +129,12 @@ class RuleMatcher:
     def _is_rule_disabled(self, rule: Rule) -> bool:
         if rule.is_disabled(self.target_version):
             return True
+        # Evaluate --ignore first so the filter is recorded as matched even when the rule is
+        # also filtered out by --threshold, otherwise it is falsely reported as unmatched.
+        ignored = self.ignore_filter.matches(rule)
         if rule.severity < self.threshold and not rule.config.get("severity_threshold"):
             return True
-        return self.ignore_filter.matches(rule)
+        return ignored
 
     def is_rule_fixable(self, rule: Rule) -> bool:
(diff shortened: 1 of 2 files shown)


bhirsz (approved):
Looks good, thanks for contribution!

PR #1788, opened 2026-07-30
fix(formatter): keep inline comment cells together in ReplaceWithVAR

Sanjays2402:
Closes #1715

Robot Framework tokenizes a single inline comment into one `COMMENT` token per data cell, and only the first cell starts with `#`. When `ReplaceWithVAR` moved a multi-cell comment above the generated `VAR`, it emitted one comment line per token, so every cell after the first lost its `#` and became an executable line — silently breaking the suite at runtime.

Cells that do not start with `#` are now appended to the preceding comment, so the original comment stays on one line. Regression test added to `tests/formatter/formatters/ReplaceWithVAR/test_formatter.py` (parametrized over `Create Dictionary` and `Create List`); it fails on master and passes with the fix, and the full `tests/formatter` suite stays green.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/src/robocop/formatter/formatters/ReplaceWithVAR.py
+++ b/src/robocop/formatter/formatters/ReplaceWithVAR.py
@@ -191,8 +191,26 @@ class ReplaceWithVAR(Formatter):
             else:
                 node.tokens = [*node.tokens[:-1], Token(Token.SEPARATOR, "  "), comments[0], node.tokens[-1]]  # type: ignore[union-attr]
             return node
-        comment_nodes = [Comment.from_params(comment=comment.value, indent=indent) for comment in comments]
+        comment_nodes = [
+            Comment.from_params(comment=comment, indent=indent) for comment in self.merge_comment_values(comments)
+        ]
         return *comment_nodes, node
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


bhirsz:
I didn't comment on the PR yet as there were CI failures. Will you check it? I can do it as well, at this point I'm familiar with typical issues for older versions of Robot we support (for example test use VAR which wasn't available in RF 5 etc). 

And thanks for the PR!

Sanjays2402:
Checked the failures — good instinct. The two failing jobs are the RF 5.\* and 6.\* builds, and it's the new `test_inline_comment_is_not_split_into_keyword_calls` that fails there: `ReplaceWithVAR` itself requires RF 7 (it prints "requires Robot Framework 7.\* version" and leaves files unchanged on older versions), but my test called `format_files` directly without the version gate that `FormatterAcceptanceTest.run_tidy` applies.

Fixed in b82fb29: the test now skips via `self.enabled_in_version(">=7")` on older versions, matching the existing convention. Verified locally — the two new cases skip on RF 6.1.1 and the full test file (12 tests) passes on RF 7.3.2.

Issue #1677, opened 2026-02-10
New formatter suggestion: make keyword call use named parameters


Guekka:
## Problem

In our code, we have some keywords with 15+ parameters (not my choice...)
Positional ordering is bound to fail, and I've already fixed many related issues

If we could instead force named arguments, errors would be much more obvious

## Implementation

The main difficulty is that, as far as I know, robocop does not pass context to the formatters. For this formatter to work, a list of keyword definitions in the workspace would be needed. This could be done in a first pass, before running any formatter / check

Once we have the keyword definitions, the difficulty is the same as any standard formatter

## Proof of concept

With Copilot, I wrote a quick PoC. It seems to work correctly. The main issue I had was that I need it to run *before* `SplitTooLongLine`, and the only way I found is to redefine the full list of formatters

Here's [the code](https://gist.github.com/Guekka/0de25b57718580d72965518379103db7)

wsalmonct:
Im working on a moderately simple project that has some verbs with 3/4 arguments and its very easy to get them wrong when they are purely positional. Even when you try to embedded them into a sentence to give context. So we are using named args more and more so having linting to enforce them would be great.

We have started using robocop and we appreciate it as is. But also think this would be a valuable improvement.

bhirsz:
Thanks for keeping issue active - I am currently going over the issues and selecting for next wave of bigger changes and it may be in next cuts.. . There are several connected issues regarding keeping the context. We partly keep it already, but only 'single-file-wise', not on project level. Once we manage to 'find definition' of keyword we would be able to add a lot of new rules (not used keyword, invalid usage and so on). 

This was always too big to tackle for me as I am doing it in my free time, but recently I am using AI a lot (don't worry, we're benefitting from thousands existing robocop tests added over the years, they work as good guardlines) and this is starting to be a possiblity. Stay tuned.

wsalmonct:
Hi @bhirsz 

> This was always too big to tackle for me as I am doing it in my free time, but recently I am using AI a lot (don't worry, we're benefitting from thousands existing robocop tests added over the years, they work as good guardlines) and this is starting to be a possiblity. Stay tuned.

We are all trying to find our way!

I cant promise a lot of help but when you have the building blocks setup we could maybe help, or if your AI races ahead then having a human test might be more useful? Let me know, I will see what I can do.

bhirsz:
[...]

PR #1822, opened 2026-08-13
feat: project level context with rules using it

### LINKED ISSUE #1017: [Rule] Not used keyword
This rule was previously rejected because Robocop does not support dynamic code. Keywords can be defined under variables and executed using Run Keyword, resource can be imported dynamically, Python library keywords cannot be discovered without initializing library first etc. Those kind of limitations lead to creation of Sherlock which focus on using log from actually execution to avoid such issues. But maintaining yet another project is not feasible for me and there is little to no progress here recently despite it is one of most wanted feature from users.

But thanks to our planned feature - community rules it is possible to define optional rule. Such rule, with known limitations and pitfalls (which I will list in this thread and in the documentation) can be enabled by the user. Also we will need access to all files in the project which can be done thanks to #1015.

### LINKED ISSUE #1119: [Rule] Duplicated Variable Name
Robot Framework still have flat, common variables namespace. When you define the same variable in more than one resource or variablefile, and import them both, execution results may be unpredictable. That would be great if robocop will catch such problems.

### LINKED ISSUE #1780: Rule Request: Unused resource
Not sure if I am producing some duplicate issue or asking for already discussed rules, but it would be amazing if there was a rule for unused imported resource/libraries.

bhirsz:
## Summary

Adds project level context to the linter and the first rules built on top of it. This is the foundation discussed in #1677, where the direction was to investigate how `robotcode` solves this. Implemented self-contained, without depending on the `robotcode` LSP stack.

Still a draft - more rules are planned on this branch.

## Problem

`run_project_checks()` never parsed anything. It only handed the `ConfigManager` to each checker, which then had to `rglob` and parse files on its own. Writing a project rule meant reimplementing discovery, parsing and import resolution every time.

That is why `unused-keyword` was deprecated and disabled: it defines `visit_File`, but `ProjectChecker` has no visiting machinery, so it was effectively dead code.

## Project context

A `ProjectContext` is now built once and passed to `ProjectChecker.scan_project`:

- **Keyword definitions and usages** indexed across the project, with embedded arguments, `Resource.Keyword` prefixes and BDD prefixes.
- **Argument specifications** parsed with Robot Framework's own `UserKeywordArgumentParser`, so no library is imported and no user code is executed.
- **Keyword visibility** resolved through transitive resource imports, so same named keywords in unrelated files are not confused.
- **Variables** with Robot Framework compatible scoping (command line > file's own `*** Variables ***` > imported > builtin).
- **Imports resolved to real paths** via `robot.variables.Variables.replace_string`, so `%{ENV}`, nesting and escaping behave exactly like in Robot Framework.

Imports using variables that cannot be resolved are marked `UNRESOLVABLE` and are **never reported**. This is the main safety valve against false positives on dynamic paths.

The context is only built when an enabled checker actually uses it, so a run without project rules does not pay for parsing the project.

## Configuration

Variables can be provided the same way as in Robot Framework:

```
robocop check-project --variable RESOURCE_DIR:resources
robocop check-project -v RESOURCE_DIR:resources
```

or in the configuration file:

```toml
[tool.robocop.variables]
RESOURCE_DIR = "resources"
```

Variable names are no longer normalized, so `MY-VAR` stays `MY-VAR`. Previously `normalize_config_keys` rewrote `-` to `_` recursively, which would have silently corrupted user defined variable names.

## New rules

All of them are disabled by default and have to be selected explicitly.

### `invalid-argument-count` (ARG08)

Compares the arguments used in a call with the `[Arguments]` of the definition.

```
test.robot:19:5 ARG08 Keyword 'Login' expects 2 arguments but 1 provided, missing ${password}
test.robot:20:5 ARG08 Keyword 'Login' expects 2 arguments but 3 provided
test.robot:21:5 ARG08 Keyword 'With Default' expects from 1 to 2 arguments but 0 provided, missing ${a}
```

Named arguments are resolved the way Robot Framework does it: `name=value` is only named when `name` matches an argument of the keyword or the keyword accepts free named arguments, otherwise it is a positional value.

Skipped to avoid false positives: dynamic keyword names, keywords not found or ambiguous, embedded arguments, `@{list}` / `&{dict}` expansion, and template keywords.

### `unused-keyword` (KW04)

Un-deprecated and rewritten. Reports keywords never called anywhere in the project, taking into account setups, teardowns, templates and `Run Keyword` variants.

Private keywords are only matched against calls from their own file. Calls with a name built from a variable become a pattern, so `Login ${type}` marks both `Login Admin` and `Login User` as used.

### `unused-resource-import` (IMP05)

Reports resource imports whose keywords and variables are never used.

```
test.robot:3:13 IMP05 Imported resource file 'resources/unused.resource' is not used
```

Resource imports are transitive in Robot Framework, so an import is checked against the usages of every file that can see it. A resource imported only to re-export its own imports is therefore not reported, and neither is a resource that defines no keywords and no variables.

Skipped to avoid false positives: unresolved imports and files calling keywords by a name built from a variable.

### `duplicated-variable-in-project` (DUP11)

Reports the same variable defined in several files that are visible together through imports.

```
other.resource:2:1 DUP11 Variable 'browser' is also defined in 'common.resource' (line 2)
```

Robot Framework does not complain about this, but the value used at runtime depends on the import order, which makes it a common source of hard to debug problems. Names are normalized, so `${my var}`, `${MY_VAR}` and `${myvar}` are the same variable.

### `unresolved-resource-import` (IMP07)

Reports resource imports pointing to files that do not exist.

## Backwards compatibility

`ProjectChecker.scan_project` gained an optional third `context` argument. Custom checkers written against the old two argument signature are detected by inspecting the signature and are still called without the context.

Signature inspection is used rather than `try/except TypeError`, so a `TypeError` raised *inside* a checker is not silently misinterpreted as an old signature.

## Docs

The project checks documentation described a `scan_project` signature and a `source` argument that never worked. Rewritten and corrected, which closes #1722.

## Follow-up

Planned next, not in this PR yet:

- `unused-library-import` (#1780), limited to standard libraries until library introspection lands
- `ambiguous-keyword-name`, `keyword-not-found`
- per-file context caching
- exposing the context to formatters (#1677)

## Test plan

- **Cross-checked `invalid-argument-count` against `robot --dryrun`** on a fixture with valid, invalid, nested and dynamic calls - Robocop reports exactly the same set of errors, no more and no less.
- The previously skipped `unused-keyword` acceptance tests are enabled again. Expected output gained one keyword defined in a resource file and never called, which is the exact case the old rule could not detect.
- New unit tests for argument specs and call validation, nested run keyword argument association, keyword visibility through imports, `--variable` parsing and config key normalization.
- Manually verified that an unresolvable import variable produces **no** diagnostic, and that supplying it via `--variable` makes the import resolve.
- Full suite: 1969 passed, 77 skipped (was 1883 passed, 79 skipped on `main`).
- `ruff check` / `ruff format` clean; `mypy` unchanged from baseline (7 pre-existing errors in untouched files).

Closes #1017
Closes #1119
Closes #1780
Closes #1722


bhirsz:
I did thorough review and asked for changes in the separate process - the rest is 'LGTM'. I will be extending some of the tests manually after this PR is merged.

Issue #1631, opened 2026-01-16
Fix rules


bhirsz:
Meta ticket for rules that can have fixes.

The tables below contain a brief overview of every rule: whether an automatic fix is possible, how difficult and how
safe it would be, whether an existing formatter already covers the rewrite, and the current implementation status.

**Legend**

| Column | Values |
| --- | --- |
| Fix | ✅ possible / ❌ not possible (needs manual intervention, usually a bigger refactor) |
| Difficulty | easy / mid / hard |
| Safety | safe / unsafe (fix may change the behaviour, would need `UNSAFE` applicability) |
| Status | ✅ implemented / 🔄 in progress / ⬜ to do / ➖ not planned |

Rules with an existing formatter counterpart are marked in the *Notes* column - implementing the fix may allow us to
deprecate the formatter (as it happened with `RemoveEmptySettings`).

### ANN

| Rule | Name | Fix | Difficulty | Safety | Status | Notes |
| --- | --- | --- | --- | --- | --- | --- |
| ANN01 | missing-section-variable-type | ❌ | - | unsafe | ➖ | could be derived from the value, but unsafe |
| ANN02 | missing-argument-type | ❌ | - | - | ➖ | |
| ANN03 | missing-for-loop-variable-type | ❌ | - | - | ➖ | |

### ARG

| Rule | Name | Fix | Difficulty | Safety | Status | Notes |
| --- | --- | --- | --- | --- | --- | --- |
| ARG01 | unused-argument | ✅ | easy | unsafe | ➖ | extremely unsafe |
| ARG02 | argument-overwritten-before-usage | ❌ | - | unsafe | ➖ | unclear what the fix should be |
| ARG03 | undefined-argument-default | ✅ | easy | safe | ⬜ | add `${EMPTY}`, similar to `ReplaceEmptyValues` |
| ARG04 | undefined-argument-value | ✅ | easy | unsafe | ⬜ | |
| ARG05 | invalid-argument | ❌ | - | - | ➖ | |
| ARG06 | duplicated-argument-name | ❌ | - | - | ➖ | |
| ARG07 | arguments-per-line | ✅ | mid | safe | ⬜ | split into multiple lines |
| ARG09 | missing-argument-name | ✅ | - | safe | ✅ | #1837 |

### COM

| Rule | Name | Fix | Difficulty | Safety | Status | Notes |
| --- | --- | --- | --- | --- | --- | --- |
[...]

PR #1859, opened 2026-08-16
feat: add fixes for the assignment sign and negative condition rules

bhirsz:
Part of #1631.

## MISC05 `inconsistent-assignment-in-variables` (`ALWAYS`)
Replaces the assignment sign of the reported variable with the expected one (`\\ =` -> `\\`, etc.). The realignment of the values afterwards is left to `robocop format`.

## MISC14 `misplaced-negative-condition` (`SOMETIMES`)
Rewrites `not \\ is None` to `\\ is not None` in place, so the rest of the condition is preserved.

Not fixed (reported only):
- double negations such as `not not \\ is None` - the intention is not clear and the rule's own suggestion is not correct there;
- conditions that cannot be located unambiguously in the source (unexpected separators, the condition appearing more than once in the same argument).

## MISC11 `multiline-inline-if`
Not implemented on purpose - reformatting an inline IF is the responsibility of \
obocop format\. A documentation note pointing at the `InlineIf` formatter was added instead, following the scope policy in #1631.


bhirsz:
@copilot fix rebase conflicts

PR #1928, opened 2026-08-31
feat: register deprecated rules so they are recognized in configuration

bhirsz:
## What
Register deprecated rules so they are recognized during configuration.

## Why
Three deprecated rule classes exist (`deprecated = True`) but were never loaded, so they were completely invisible to users:

- `DEPR01 if-can-be-used` (legacy id `0908`)
- `DEPR02 deprecated-statement` (legacy id `0319`)
- `MISC12 unnecessary-string-conversion` (legacy id `0923`)

Before this change, `--select`/`--configure` targeting these (by name, `rule_id`, or legacy id) reported *"did not match with any rule"* or was silently ignored, and `robocop list --filter DEPRECATED` showed nothing.

## What changed
- `LinterImporter._collect_deprecated_rules()` discovers `Rule` subclasses with `deprecated=True` and registers them by `name`, `rule_id`, and every `deprecated_names` alias. Wired into both internal rule modules and external checker modules.
- Removed the dead `register_deprecated_rules` method and the stale `# FIXME` / commented-out registration lines.
- `RuleMatcher.check_unmatched_filters()` now splits unmatched filters into **deprecation warnings** vs genuine **"did not match"** errors (both to stderr).
- `RulesLoader.register_deprecated_rules()` registers each deprecated rule with `enabled = False` and applies any user configuration (emitting a deprecation warning).

## Result
- `--select`/`--configure` on a deprecated rule now emits a clear deprecation warning instead of an error.
- `robocop list --filter DEPRECATED` now lists all 3.
- Default `robocop list` still reports 194 rules (no regression).

## Notes
- Deprecated rules default to `enabled=True`; the MCP rules catalog reads `rule.enabled` directly, so they are registered with `enabled=False` to avoid leaking into the catalog (regression caught + covered by tests).

## Tests
5 new tests (3 in `test_rule_matcher.py`, 2 in `test_custom_rules.py`). Full suite: 2258 passed, 87 skipped. ruff + mypy clean.

### DIFF
--- a/src/robocop/runtime/resolver.py
+++ b/src/robocop/runtime/resolver.py
@@ -123,6 +123,8 @@ class RuleMatcher:
         self.ignore_filter = RuleFilter.from_rules(ignore)
         self.fixable_filter = RuleFilter(exact_matches=set(fixable))
         self.unfixable_filter = RuleFilter(exact_matches=set(unfixable))
+        # Populated after the rules are loaded. Maps a deprecated rule name / id / old name to its rule.
+        self.deprecated_rules: dict[str, Rule] = {}
 
     def is_rule_enabled(self, rule: Rule) -> bool:
         if self._is_rule_disabled(rule):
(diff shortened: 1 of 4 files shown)


src/robocop/runtime/resolver.py
```diff
@@ -295,10 +302,28 @@ def _get_checkers_from_modules(
...
+        Discover deprecated rule classes defined in a module.
+
+        Deprecated rules are not attached to any checker, so they are not loaded through the regular flow.
+        Collecting them here keeps them recognizable by ``--select`` / ``--configure`` and by ``robocop list``,
+        so that referencing a removed rule reports a deprecation warning instead of an "unknown rule" error.
+        """
+        for name, rule_class in inspect.getmembers(module, inspect.isclass):
+            if not is_rule((name, rule_class)):
+                continue
+            rule = rule_class()
```
bhirsz: Wouldn't it create instance of rule class for each imported rule? (Now at 180+ rules). Do we have any other option other than this?

bhirsz: Good catch. Fixed in 513bdb57: `deprecated` is a class attribute (`deprecated: bool = False` on the `Rule` base), so we now gate on `rule_class.deprecated` **before** instantiating. Only the deprecated rule classes (3 today) are constructed instead of building an instance of every imported rule (180+) just to read the flag. `name` / `rule_id` / `deprecated_names` are likewise class attributes, so the only instantiations left are the deprecated ones we actually register.

PR #1929, opened 2026-08-31
chore: remove stale TODO on unnecessary-string-conversion rule

bhirsz:
## What
Remove a stale `# TODO: Not used atm` comment on `UnnecessaryStringConversionRule`.

## Why
The rule is a deprecated rule that is now properly surfaced (see the deprecated-rule registration work), so the "not used" note is no longer accurate.

## What changed
- `src/robocop/linter/rules/misc.py`: delete the stale comment.

### DIFF
--- a/src/robocop/linter/rules/misc.py
+++ b/src/robocop/linter/rules/misc.py
@@ -851,7 +851,7 @@ class MultilineInlineIfRule(Rule):
         )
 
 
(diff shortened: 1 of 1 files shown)


bhirsz:
Folded into #1928 (feat/register-deprecated-rules) as commit 975603fe — the comment removal only makes sense alongside the deprecated-rules registration.

PR #1933, opened 2026-09-02
feat: add too-many-required-arguments and too-many-optional-arguments rules

### LINKED ISSUE #1145: [Rule] Too many arguments without default

bhirsz:
Adapts #1176 by @Lakitna to the current Robocop architecture (post robocop + robotidy merge) and re-opens it. Original authorship is preserved in the commit (authored by Sander van Beek).

Closes #1145

## What changed

Adds two new `lengths` rules:

- `too-many-required-arguments` (`LEN34`) - default limit `max_args=5`
- `too-many-optional-arguments` (`LEN35`) - default limit `max_args=10`

These allow configuring separate limits for required vs optional keyword arguments. As discussed in #1145, the existing `too-many-arguments` (`LEN07`) rule is now **disabled by default** in favor of the two new rules.

## Notes on adaptation

The original PR targeted the old rule/checker structure. In the new style:
- Rules are declarative classes in `src/robocop/linter/rules/lengths.py` with a `check()` method; wiring lives in `src/robocop/linter/checkers/test_case_keyword.py`.
- Rule IDs use the category prefix form (`LEN34`/`LEN35`) instead of the old numeric `0533`/`0534`.
- Acceptance tests moved to `tests/linter/rules/lengths/too_many_{required,optional}_arguments/` and now also cover `extended` output.
- `docs/rules_list.md` regenerated.

The main logic and test behavior mirror the original PR: required = arguments without a default, optional = arguments with a `}=` default; the `[Arguments]` setting is highlighted (already the behavior on `too-many-arguments`).

### DIFF
--- a/docs/rules_list.md
+++ b/docs/rules_list.md
@@ -5456,6 +5456,8 @@ Redesign the test and move complex logic to separate keywords to increase readab
 
 ### LEN07: too-many-arguments
 
+> Rule is disabled by default. Enable it by using ``--select too-many-arguments`` option.
+
 Added: `v1.0.0`
 
 Supported RF version `All`
@@ -5471,6 +5473,18 @@ Fix availability: There is no automatic fix.
 **Documentation**:
 
 Keyword has too many arguments.
+
+Keywords with many arguments can be hard to understand. A large number of arguments can
+indicate that the keyword does multiple different things and/or that it's complex.
+
+Try to reduce the number of arguments. For example:
+
+- Split the keyword into multiple keywords
+- Collect the required data in a different way
+
+This rule is disabled by default in favor of the ``too-many-required-arguments`` and
+``too-many-optional-arguments`` rules, which allow to configure separate limits for required and
+optional arguments.
 
 **Style guide**:
 
@@ -5511,7 +5525,7 @@ Keyword has too many arguments.
 
 ??? example "max_args"
 
-    Number of lines allowed in a file
+    Number of allowed keyword arguments
 
     **Default value:** 5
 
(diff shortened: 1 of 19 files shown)


Lakitna:
Looks good to me :) Thanks for pushing this across the finish line!

